PR #35, opened 2026-07-26
feat(capabilities): add Lean declaration discovery

N0zoM1z0:
## Problem

Agents can replay completed Lean source, but the installed portfolio cannot retrieve or inspect premises from the same pinned Lean and Mathlib environments. That forces models to guess declaration names or hide retrieval outside the capability trust and provenance model.

## Change

- add atomic EXPLORE capabilities `lean.declaration.search` and `lean.declaration.inspect`
- use a bounded read-only Lean helper with deterministic name order and structured elaborated-type constant patterns
- return typed declaration, source, coverage, stop-reason, and exact environment-manifest identity records
- keep CORE aligned with `Init.*`, exclude helper-local/private declarations, and fail closed on runtime mutation, timeout, oversized output, malformed protocol data, or missing names
- add the public Mathlib retrieval, inspection, and `lean.check` composition plus reference documentation and roadmap measurements

This adds two experimental capability IDs only when the optional pinned Lean runtime is healthy. Existing capability inputs, outputs, and verification authority are unchanged. Declaration retrieval is COMPUTED evidence; only the existing authorized `lean.check` path can return VERIFIED.

## Validation

- `uv sync --dev`
- `uv run pytest` (431 passed)
- `uv run ruff check .`
- `uv run mypy`
- `uv build`
- `git diff --check`
- real pinned Mathlib search, inspect, structured type-pattern scan, and independent sqrt-two replay
- built wheel inspected to confirm both the Python adapter and Lean helper are packaged

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,9 +202,12 @@ lake build
 ```
 
 The operation and trust boundary are documented under
-[`lean.check`](docs/reference/tools.md). This is a trusted local integration,
-not a broker sandbox; the pinned Lake environment still has host-local runtime
-access.
+[`lean.check`](docs/reference/tools.md). Read-only
+[Lean declaration discovery](docs/reference/lean-declaration-discovery.md)
+can retrieve and inspect premises before completed source crosses that checker
+boundary; the [guided reproduction](docs/tutorials/lean-declaration-discovery.md)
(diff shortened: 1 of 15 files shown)


N0zoM1z0:
LGTM!

PR #37, opened 2026-07-26
feat(eval): benchmark Lean declaration discovery

N0zoM1z0:
## Problem

The experimental `lean.declaration.search` and `lean.declaration.inspect` capabilities had a public reproduction but no held-out measurement of their marginal value against the same portfolio with `lean.check` only.

## Change

- add two held-out Mathlib declaration families and a strict structured report contract;
- add a local construction-time portfolio ablation so both conditions use the same MCP surface while control omits only declaration search/inspect;
- require exact statement/proof trace, durable claim/candidate/certificate bindings, and an authorized Lean verification record;
- separate attempted/successful discovery, proof rejection, structured MCP error, false certification, and operational checker failure telemetry;
- record the three-valid-pair pilot and the decision to revise rather than recommend, consolidate, or retire the capabilities.

The pilot found no completion lift. On the `List.revzip` case, treatment added 226–288 seconds, 208k–365k input tokens, and 3–6 additional tool errors. The next step is a reusable pinned index or persistent query service plus compact catalog discovery, not a larger timeout or more goal/tactic tools.

No capability contract or assurance rule changes. Construction-time exclusions are local evaluation composition, not runtime authorization. Search and exact inspection remain separate experimental `EXPLORE` outcomes; only `lean.check` may produce the accepted `VERIFIED` result.

## Validation

- `uv sync --dev`
- `uv run pytest` (`438 passed`)
- `uv run ruff check .`
- `uv run mypy`
- `uv build`
- `git diff --check`
- held-out paired Codex pilot with `gpt-5.6-sol`, high reasoning effort, randomized condition order, exact Lean replay oracle

### DIFF
--- a/benchmarks/ab_cases/lean-report.schema.json
+++ b/benchmarks/ab_cases/lean-report.schema.json
@@ -1,0 +1,87 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "case_id",
+    "statement",
+    "proof",
+    "declarations",
+    "conclusion",
+    "assurance",
+    "verification_record_uri",
+    "limitations",
+    "feedback"
+  ],
+  "properties": {
+    "case_id": {
+      "type": "string"
+    },
+    "statement": {
+      "type": "string"
+    },
+    "proof": {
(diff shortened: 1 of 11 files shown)


N0zoM1z0:
LGTM!

PR #42, opened 2026-07-26
feat(sat): define canonical SAT artifact contracts

N0zoM1z0:
## Problem

The SAT vertical slice had no domain-owned identity for CNF instances, assignments, or raw UNSAT proofs. A JSON Schema alone also cannot enforce canonical ordering or derived digest invariants before generic artifact writes.

## Change

- add canonical named-variable CNF normalization with deterministic DIMACS projection and exact map/projection digests;
- add total assignment and raw DRAT proof artifacts bound to the exact CNF URI, object and payload digests, variable map, projection version, full scope, producer runtime, and resource budget;
- preserve raw proof bytes as bounded canonical base64 without parsing or claiming UNSAT;
- add model-backed schema registration so cross-field invariants run through the normal artifact validation path and after kernel restart;
- install the schemas in the base kernel without adding a SAT solver, checker, capability, or assurance promotion;
- document the contracts, roadmap checkpoint, and threat boundary.

The artifact formats are new and pre-stable. `SchemaRegistry.register_model` is additive. No existing capability ID or MCP tool changes, and no evidence in this slice can become `VERIFIED`.

## Validation

- `uv sync --dev`
- `uv run pytest -n 1` — 463 passed
- `uv run ruff check .`
- `uv run mypy`
- `uv build`
- `git diff --check`
- wheel and source-distribution content inspection

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,9 @@ Release contracts and engineering evidence are:
 - [Tool surface](docs/reference/tools.md)
 - [Provider runtime contract](docs/reference/provider-runtime.md) for
   availability, exact backend identity, install tiers, and local measurement.
+- [SAT artifact contracts](docs/reference/sat-artifacts.md) for canonical CNF,
+  total assignment, and raw proof identity before solver or checker
+  installation.
 - [v0.2 specification](docs/reference/specifications/v0.2.md) and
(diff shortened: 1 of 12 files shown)


N0zoM1z0:
LGTM!

PR #49, opened 2026-07-26
feat: verify exact SAT assignments

N0zoM1z0:
## Problem

SAT assignment artifacts are exact and source-bound, but storage alone cannot establish satisfiability. The roadmap needs a small independent checker that can promote only a satisfying total assignment while keeping malformed evidence, rejection, and operational failure mathematically inconclusive.

## Solution

- Add a standard-library-only clean-process SAT assignment checker that independently validates canonical CNF shape/order, payload and projection digests, exact assignment/evidence bindings, lineage, total strict-Boolean values, and every clause.
- Add the operator-authorized `sat.model.verify` capability. It re-derives source bindings before dispatch and uses the existing `VerificationService`/`VerificationRecord` promotion path.
- Keep the capability absent unless bundled reference checkers are explicitly installed.
- Preserve legacy checker request shapes; the additional payload-digest and parent metadata is opt-in only for the new SAT checker.
- Add attack, clean-process, rejection, timeout, crash, and pre-dispatch misbinding tests, plus reference/roadmap/threat-model documentation.

## Testing

```sh
uv sync --dev
uv run pytest                              # 504 passed
uv run ruff check .
uv run ruff format --check .
uv run mypy
uv run deptry .
uv build
git diff --check
git diff -- AGENTS.md README.md docs/
```

The pinned Lean REPL required by the latest main was materialized locally with `lake build repl` before the final full-suite run.

This is trust-critical checker and evidence-promotion code. `VERIFIED` is produced only after the measured, authorized checker accepts in a clean process and the kernel commits a binding `VerificationRecord`. An unsatisfying assignment remains `UNKNOWN` and does not establish UNSAT; malformed/misbound inputs fail before dispatch; timeout, cancellation, and checker failure remain non-verified. The threat model and SAT reference contract are updated.

The SAT contracts and capability are experimental pre-stable additions. Existing checker request shapes remain unchanged unless a caller explicitly requests the new storage metadata path.

## Checklist

- [x] Python validation passes locally
- [x] Relevant documentation updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,8 +116,7 @@ Release contracts and engineering evidence are:
 - [Provider runtime contract](docs/reference/provider-runtime.md) for
   availability, exact backend identity, install tiers, and local measurement.
 - [SAT artifact contracts](docs/reference/sat-artifacts.md) for canonical CNF,
-  total assignment, and raw proof identity before solver or checker
-  installation.
+  raw model and proof identity, and independently checked total assignments.
 - [v0.2 specification](docs/reference/specifications/v0.2.md) and
   [conformance gate](docs/reference/conformance-v0.2.md)
 - [Testing strategy](docs/reference/testing-strategy.md),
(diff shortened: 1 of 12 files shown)


N0zoM1z0:
LGTM!

PR #53, opened 2026-07-26
test: restore fast feedback and balanced shards

N0zoM1z0:
## Problem

The default test workflow had become a development bottleneck. The fast lane spent most of its time rebuilding the full product kernel for each direct DRAT-trim checker case, and CI requested duration-based sharding without a checked-in duration file, so fresh runners silently fell back to test-count splitting. Unfiltered local pytest also mixed memory-heavy Lean tests into the general xdist pool.

## Result

- Reuse a minimal, real, module-scoped SAT artifact fixture in the direct DRAT-trim checker tests while preserving every binding, lineage, malformed-proof, output-limit, runtime-digest, and authorization attack case.
- Reduce the measured fast lane from 43.92 seconds to 6.55 seconds (about 85%).
- Commit the measured non-Lean `.test_durations` data and add `make test-durations`; both CI groups currently estimate 558 seconds and select 263 tests.
- Document the measured lane costs, why real filesystem/subprocess/backend coverage remains necessary, and which commands belong in edit, non-Lean, Lean, and final-validation loops.

No runtime API, mathematical contract, assurance rule, checker behavior, or release specification changes. This changes test setup, CI scheduling data, and maintainer guidance only.

## Validation

- `uv run pytest -m "not lean_runtime" --store-durations --clean-durations --durations=20` — 526 passed
- `make test-lean PYTEST_ARGS="--durations=30"` — 20 passed
- `make test-fast` — 246 passed in 5.05 s on the rebased final tree
- Both exact CI shard collection commands — group 1 and group 2 each estimated at 558.00 s
- `make lint`
- `make typecheck`
- `uv build`
- `jq empty .test_durations`
- `git diff --check`

The untracked local `droid.resume.txt` was not modified or included.


N0zoM1z0:
LGTM!

PR #68, opened 2026-07-26
feat(sat): validate the certificate portfolio

N0zoM1z0:
## Problem

The four SAT outcomes had unit and attack coverage, but no frozen public backend reproductions or model-in-the-loop portfolio evaluation. Real CaDiCaL output also exposed two interoperability gaps: cleanup deletions after an empty clause were rejected, and positional assignment artifacts were easy for agents to join against the wrong variable order.

## Change

- add unscored public CaDiCaL/DRAT reproductions for `BOOL-MUS-001`, a satisfiable CNF, and pigeonhole UNSAT;
- admit only cleanup deletions after the empty clause and replay exact raw proofs with DRAT-trim forward checking (`-f -W`), preserving fail-closed syntax and runtime checks;
- add `sat_decision` A/B cases, a hidden brute-force oracle, durable binding/trace checks, and clean-kernel replay scoring;
- make `sat.model.find` return an inline named assignment map beside its durable unverified URI;
- record the frozen Terra-high pilot: control and treatment both passed 4/4, while treatment produced 4/4 independently replayed records with no false certification or tool error, at materially higher latency/token cost;
- mark roadmap item 8 complete and document the next discovery-cost concern.

## Compatibility and trust impact

`SatModelFindOutput` now requires the named `assignment` map whenever `ASSIGNMENT_PRODUCED` carries an `assignment_uri`. This is an experimental pre-stable contract expansion. The inline map remains unverified producer output; only `sat.model.verify` can promote the exact bound assignment. Checker authorization, evidence binding, and fail-closed status rules are unchanged.

## Validation

- `make test-fast` — 253 passed
- `make test` — 537 passed
- focused SAT integration/public reproduction run — 10 passed
- `make lint`
- `make typecheck`
- `make build`
- `make test-lean` before rebase — 19 passed, one transient Lean runtime result lacked a record; the identical kernel probe and isolated pytest case then passed
- frozen local Codex evaluation — 8 runs, `gpt-5.6-terra`, high reasoning, 4 valid pairs

### DIFF
--- a/benchmarks/ab_cases/sat-report.schema.json
+++ b/benchmarks/ab_cases/sat-report.schema.json
@@ -1,0 +1,66 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "object",
+  "properties": {
+    "case_id": {"type": "string"},
+    "status": {"enum": ["SATISFIABLE", "UNSATISFIABLE", "UNKNOWN"]},
+    "conclusion": {"enum": ["TRUE", "UNKNOWN"]},
+    "assurance": {"enum": ["SELF_CHECKED", "COMPUTED", "VERIFIED"]},
+    "final_verification": {"enum": ["UNVERIFIED", "VERIFIED"]},
+    "evidence_kind": {"enum": ["ASSIGNMENT", "UNSAT_PROOF", "NONE"]},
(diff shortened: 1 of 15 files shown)


N0zoM1z0:
LGTM!

PR #70, opened 2026-07-26
docs: record follow-up test cost audit

N0zoM1z0:
## Problem

After the SAT portfolio work, the complete Python and Lean lanes were visibly slow enough to question whether they contained redundant tests or checks. The newly merged #69 already changed the routine developer and CI lanes, so this follow-up records which costs are necessary and avoids layering a conflicting runner change on top.

## Change

- record the current 537-test duration distribution and the distinct boundaries covered by the slow tail;
- retain full Python-version, clean-process, external-backend, end-to-end, and pinned-Lean coverage;
- record an exploratory 4-versus-8-worker measurement, while keeping the stable four-worker default because the single-host result is not sufficient for a subprocess-heavy suite;
- identify procedural repetition—not the mathematical cases—as the actionable redundancy;
- document how #69 addresses it through an 8-second local handoff lane, named focused targets, change-aware docs/npm CI, and remote Python/Lean debug workflows.

## Impact

Documentation only. No test, marker, CI gate, runtime setting, product contract, or mathematical-assurance boundary changes.

## Validation

- `make check` — Ruff plus 256 selected tests passed in 8.36 seconds
- exploratory exhaustive runs — all 537 non-Lean tests passed with both worker configurations used during measurement
- `uv run --frozen ruff format --check docs/contributing/test-suite-cost-audit.md`
- `git diff --check`

### DIFF
--- a/docs/contributing/test-suite-cost-audit.md
+++ b/docs/contributing/test-suite-cost-audit.md
@@ -103,3 +103,41 @@ caused it. Ordinary test edits do not need a refresh.
   blanket `slow` marker or weakening required coverage.
 - Move backend combinations to a slower lane only when the pull-request lane
   still exercises every affected trust boundary.
+
+## Follow-up audit
+
+The expanded 537-test non-Lean suite has a median recorded case duration of
+0.08 seconds, but its slowest five percent take at least 6.62 seconds. Those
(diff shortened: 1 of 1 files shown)


N0zoM1z0:
LGTM!

PR #72, opened 2026-07-26
feat: verify compatible SMT proofs with Carcara

N0zoM1z0:
## Problem

The cvc5 producer preserves Alethe proofs, but Jacobian had no independent
checker that could promote compatible UNSAT evidence. Zero reported holes is
not itself verification, and Carcara can exit successfully for holey proofs or
be configured to tolerate unknown rules.

## Change

- Pin Carcara 1.1.0 at cvc5 1.3.4's compatible source commit, including
  executable digest and adjacent provenance.
- Add the operator-authorized experimental `smt.unsat_proof.verify`
  capability.
- Accept only the strict, currently reproduced QF_UF/Alethe profile with zero
  lexical holes, exact checker output, and clean-process replay.
- Preserve unsupported, holey, malformed, timed-out, or changed-runtime cases
  as unverified.
- Add mutation, runtime, clean-process, compatibility-matrix, and held-out
  harness coverage; document the installation and trust boundary.
- Refresh test durations so the slower process-boundary cases remain balanced
  in integration shards.

QF_LIA and QF_LRA proof artifacts remain `COMPUTED`/`UNKNOWN`; this change does
not overstate Carcara coverage.

## Compatibility and trust

This is an additive experimental contract. The verifier is exposed only when
reference capabilities are enabled and the exact pinned runtime is available.
Carcara is independent of cvc5 but is not a formally verified checker. No
solver, search implementation, or model output authorizes its own promotion.

## Validation

- `make check` — 271 passed, 4 deselected
- focused checker/kernel/benchmark integration — 11 passed, 25 deselected
- actual pinned Carcara compatibility tests — 4 passed
- full non-Lean suite — 578 passed, 4 expected optional-runtime skips
- `make check-static`
- `uv run pip-audit`
- `git diff --check`
- frozen 2-case `gpt-5.6-terra` high prescribed-capability pilot — control 2/2,
  treatment 2/2, zero false certifications, expected clean replay behavior 2/2

The Terra run evaluates contract usability and fail-closed behavior; it is not
used as code-review authority.


N0zoM1z0:
LGTM!

PR #79, opened 2026-07-26
feat(polynomials): add verified typed polynomial normalization

N0zoM1z0:
## Problem

Jacobian has exact polynomial capabilities, but it did not expose a safe atomic operation for turning a symbolic expression into canonical coefficients. Accepting formula strings would make parsing part of the trust boundary, and a successful CAS computation must not certify its own equivalence claim.

## Change

- Add `polynomial.expression.normalize`, backed by pinned SymPy 1.14.0 in an isolated bounded worker.
- Accept only a closed, versioned AST over an explicitly ordered `QQ` polynomial ring; reject strings, division, functions, assumptions, and over-budget expressions.
- Materialize separately bound source-expression and normalization artifacts with exact runtime provenance.
- Add `polynomial.expression_normalization.verify`, whose standard-library-only checker independently expands the full AST with `Fraction`, checks canonical sparse coefficients and every artifact binding, and promotes only clean-process acceptance to `VERIFIED`.
- Install the capabilities through the runtime catalog, document the contract and trust boundary, and add a private-case A/B scorer with clean replay.

This adds two experimental version-1 capability contracts and new artifact schemas. It does not add a top-level MCP tool. The base SymPy dependency is narrowed from `>=1.14,<2` to exactly `1.14.0` so producer provenance and replay behavior remain reproducible. Provider completion remains `COMPUTED/UNVERIFIED`; only the authorized independent checker can return `VERIFIED`.

## Validation

- `make check` — 347 passed, 4 deselected
- `make check-static` — Ruff, format, deptry, vulture, mypy, and wheel/sdist build passed
- focused contract/checker/integration/scorer suite — 36 passed, 29 deselected
- MCP/provider/reference-kernel integration set — 59 passed, 3 deselected
- deterministic differential probe — 30/30 random bounded expressions verified by clean replay
- `gpt-5.6-terra`, high, two private held-out expressions, four runs: control 2/2 passed; treatment 2/2 passed and 2/2 clean-replayed; zero false certifications, tool errors, parameter errors, capability rejections, or operational failures
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ availability can depend on local backends.
 | Domain | Agent-visible outcomes |
 | --- | --- |
 | Polynomial maps | Evaluate maps, compute Jacobians, search for collisions, independently verify collisions |
-| Polynomial algebra | Factor univariate polynomials, verify identities, verify exact system solutions |
+| Polynomial algebra | Normalize typed expressions, factor univariate polynomials, verify identities, verify exact system solutions |
 | Exact linear algebra | Compute determinants, rank, kernels, and integer row Hermite normal forms; find and verify rational solutions to `Ax = b` |
 | Graphs | Construct and inspect graphs, enumerate paths, realize degree sequences, test isomorphism, search colorings |
 | SAT and SMT | Find models or proof artifacts; independently replay assignments, DRAT proofs, and Alethe proofs |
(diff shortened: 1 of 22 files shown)


N0zoM1z0:
LGTM!

PR #80, opened 2026-07-26
feat(math): add verified rational inconsistency certificates

N0zoM1z0:
## Summary

- add `linear.rational_inconsistency.find`, backed by pinned Python-FLINT, to produce a normalized exact left witness `y` with proposed equations `y^T A = 0` and `y^T b = 1`
- add a standard-library clean-process checker and `linear.rational_inconsistency.verify`, with `VERIFIED` reserved for the operator-authorized independent replay
- add strictly bound inconsistency artifacts, fail-closed operational outcomes, attack/integration tests, a public worked reproduction, and portfolio documentation

## Why

The atomic capability portfolio explicitly deferred inconsistent-system outcomes until Jacobian could expose a separately checkable left-nullspace certificate. This is a small formal-facing slice that reuses the existing exact rational system artifacts and Python-FLINT runtime without adding a dependency or a workflow-shaped tool.

The linear semantics descriptor is extended to name both solution and inconsistency relations. These contracts are experimental and pre-stable.

## Validation

- `make check` (352 passed, 4 deselected)
- `make check-static` (Ruff, deptry, vulture, strict mypy, sdist/wheel build)
- `uv run pytest -q tests/integration/test_linear_rational_inconsistency.py tests/integration/test_linear_rational_solution.py` (16 passed)
- focused checker, contract, and integration suite (33 relevant tests passed before final handoff run)

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ availability can depend on local backends.
 | --- | --- |
 | Polynomial maps | Evaluate maps, compute Jacobians, search for collisions, independently verify collisions |
 | Polynomial algebra | Normalize typed expressions, factor univariate polynomials, verify identities, verify exact system solutions |
-| Exact linear algebra | Compute determinants, rank, kernels, and integer row Hermite normal forms; find and verify rational solutions to `Ax = b` |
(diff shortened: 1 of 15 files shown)


N0zoM1z0:
LGTM!

PR #123, opened 2026-07-27
fix: harden remote MCP solver workflows

N0zoM1z0:
## Summary

- propagate MCP cancellation into bounded solver process groups and cap synchronous SAT/SMT calls at a cross-client-safe 150 seconds
- repair deployed MATHLIB health reporting and legitimate single-expression Lean checking
- improve finite-coloring SAT discovery, canonical variable-map visibility, and workspace fail-closed notices
- normalize CaDiCaL DRAT deletion steps under separate raw and durable proof bounds, while requiring independent strict replay before VERIFIED
- add a frozen public f(4) agent regression and deployment/evaluation guidance

## Compatibility

The pre-stable synchronous SAT and SMT exploration budget maximum is reduced to 150 seconds. Durable artifact budget models retain their prior ranges. New result fields are additive.

## Validation

- make check: 492 passed, 4 deselected
- focused post-review suite: 22 passed
- mypy: 264 source files passed
- real MATHLIB smoke: True / by trivial VERIFIED after cold startup
- public remote smoke: SAT discovery, canonical n1/n2/n10 mapping, SAT replay, DRAT replay, and Lean all passed
- remote cancellation smoke: client cancellation returned CANCELLED and left no CaDiCaL or cvc5 worker
- local Codex gpt-5.6-sol xhigh MCP-only regression: f(4)=45 with a VERIFIED N=44 witness and six independently VERIFIED_UNSAT exhaustive N=45 branches

## Follow-up boundary

A generic durable asynchronous job and opt-in evaluation telemetry protocol remain separate design work because they require kernel ownership, retention, idempotency, and privacy contracts rather than a transport-only patch.


N0zoM1z0:
LGTM!

PR #138, opened 2026-07-28
fix(ci): use --locked instead of --frozen for fail-fast lockfile validation

morluto:
## Problem

`uv run --frozen` and `uv sync --frozen` skip lockfile freshness checks — they run with whatever `uv.lock` is on disk without verifying it matches `pyproject.toml`. If a PR modifies `pyproject.toml` without running `uv lock`, CI silently uses a stale lockfile instead of failing fast.

## Root cause

`--frozen` means "run without updating the lockfile" — it does **not** verify the lockfile is fresh. `--locked` asserts the lockfile is up-to-date and fails if it's not. The `setup` Makefile target already used `--locked`, but all `uv run` and `uv sync` commands in CI, the Makefile, and workflow files used `--frozen`.

## Changes

- **Makefile**: `UV_RUN := uv run --locked` (was `--frozen`)
- **CI workflow** (`ci.yml`): all `uv sync --frozen` → `--locked`, all `uv run --frozen` → `--locked`
- **Reusable actions** (`setup-python-tests`, `setup-lean`): same
- **Debug workflows** (`lean-debug`, `python-debug`): same
- **Scheduled validation** (`scheduled-validation.yml`): same
- **CONTRIBUTING.md**: Updated stale reference to committing `.test_durations` (it was untracked in a prior commit but the docs still said to commit it)
- **Makefile `test-durations` target**: Updated comment to reflect the file is untracked

## Validation

- `make lint` passes with `--locked`
- `make test-fast` passes: 544 passed, 4 deselected (82s)
- `test_ci_plan` + `test_ci_timings`: 38 passed

---

### DIFF
--- a/.github/actions/setup-lean/action.yml
+++ b/.github/actions/setup-lean/action.yml
@@ -46,7 +46,7 @@ runs:
           uv.lock
     - name: Install Python dependencies
       shell: bash
-      run: uv sync --frozen --dev
+      run: uv sync --locked --dev
     - name: Build Lean exploration runtime
       shell: bash
       working-directory: lean
--- a/.github/actions/setup-python-tests/action.yml
+++ b/.github/actions/setup-python-tests/action.yml
@@ -21,13 +21,13 @@ runs:
       shell: bash
       env:
         UV_PYTHON: ${{ inputs.python-version }}
-      run: uv sync --frozen --dev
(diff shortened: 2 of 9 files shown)


morluto:
Closing because these changes are already present on `main` as commit `391ab3a`.

PR #154, opened 2026-07-28
feat(mcp): add verified capabilities for frontier math workflows

N0zoM1z0:
## Summary

- add exact projective line-arrangement flat materialization plus an operator-authorized independent exhaustive checker
- add graded Jacobian syzygy minimum-degree computation from sparse polynomials or labelled rational linear factors, with an independent exact checker
- add bounded Hamiltonian-path decision and independent positive/negative verification
- expose transparent strong/weak/no lexical discovery fit instead of implying mathematical suitability
- connect producer/checker relations, provider runtimes, compact output controls, docs, reproductions, and adversarial tests

## Why

Repeated postdoc-challenge agent runs exposed recurring provenance work between factorized arrangements and sparse polynomial inputs, missing verification for projective incidence materialization, and opaque weak lexical discovery matches. Three pre-repair runs independently reproduced the factor-to-sparse bridge. The accepted changes keep cross-result theorem composition agent-owned while making the recurring atomic computations typed, inspectable, bounded, and independently checkable.

## Model-in-the-loop evidence

Two final isolated `gpt-5.6-sol` xhigh runs solved the public nine-line arrangement challenge using direct factor input. Each invoked two arrangement producers/checkers and two syzygy producers/checkers; all four domain results were `VERIFIED`, both answers were correct, and both runs completed with zero MCP execution errors. The sample is deliberately reported as a repeated range, not as statistically powered performance evidence.

## Verification

- `make check` — 596 passed in 163.06s
- `make check-static` — Ruff/format, deptry, vulture, strict mypy (293 source files), sdist and wheel build passed
- focused MCP/provider/kernel/round-trip integration — 55 passed in 675.73s
- frontier reproductions and attacks cover forged map digests, rank minors, kernel vectors, derivatives, projective normalization, duplicate projective lines, and forged negative Hamiltonian decisions

## Boundaries

Natural-language transcription, comparison of independently verified finite outputs, and final theorem composition remain agent-owned. Continuous Gaussian/Wick contraction, formal power-series operations, and lattice decomposition remain deferred until recurrence and checker-ready evidence justify stable contracts.


N0zoM1z0:
LGTM!

PR #192, opened 2026-07-29
feat(graph): verify exact distance matrices

N0zoM1z0:
## Problem

`graph.distance_matrix.compute` produces a complete exact all-pairs distance artifact, but the portfolio had no independent path to promote that exact result beyond `COMPUTED`. The small-delta plan also calls for checking whether this fundamental primitive is sufficient for a restricted-set distance composition before adding a more specialized contract.

## Result

- adds domain-owned `graph.distance_matrix.verify` without changing shared envelopes or the public MCP surface;
- replays every source with a standard-library BFS checker independent of NetworkX/the producer;
- fail-closes on malformed semantics, ordering, shape, types, diagonal, symmetry, edge distances, triangle constraints, disconnected-component closure, binding mismatches, timeout, and checker failure;
- emits an operator-authorized verification record bound to the exact graph input, matrix candidate, semantics, checker identity, witness, and provider runtime;
- routes exact-domain verifier ownership by installed result schema rather than an ambiguous capability-ID prefix;
- adds boundary/adversarial checker tests, public compute→verify integration coverage, and contract/reference documentation;
- adds an answer-visible matched A/B harness case with independent multi-source BFS scoring, ordered trace checks, durable binding checks, and clean-kernel replay.

The producer remains capped at `COMPUTED`. Only this dedicated verifier can return `VERIFIED`, and that assurance covers the exact matrix—not agent-derived restricted-set conclusions.

## Compatibility

No changes to `CapabilityResult`, common request/result envelopes, artifact URI semantics, checker authorization policy, generic installers, or the two public MCP tools. The change is an explicit graph-domain bundle/checker addition.

## Evaluation handoff

- frozen evaluation commit: `4300dea5feb67f684a207da5ea49caca859ceb74`
- frozen tree: `0d612340b761afbe574897ef08f937fef07d4291`
- run: `20260729T095955Z`, three matched pairs, configured-default Codex model, `xhigh`, 600 seconds/run, seed `20260801`
- case digest: `sha256:9a5105e1620cf039af7f7320bd7147c003c979116995073734b6854c2cbd2d1a`
- treatment/control catalog digests: `sha256:7857aa3b5399960de0cf250990291e39c165d75802421fea9d0008aa294f0c53` / `sha256:8fd971e79febc315cd6904c89598f59c59eef80dbfdb67a4c1489eb33efd8fe2`
- checker-policy digest: `sha256:3c8655dcbcecf1965b7727a9333e0f6905b5e1dc7b1e750bd9840223292acddf`
- checker provider: `AVAILABLE`, `composite-1`, `sha256:574c4727579ae1a234a2779ef2ea706a5c0313f13f82831583b91e9bd3036b74`
- result: control 3/3 and treatment 3/3 passed; 0 false certifications; treatment 3/3 clean replay and 3/3 autonomous discovery/use
- raw local evidence: `benchmarks/results/ab/20260729T095955Z/`

The public case supports retaining the complete matrix primitive; it does not establish broad portfolio value or justify a restricted-set profile capability. Two treatment runs recovered from passing the distance input wrapper to a graph-artifact consumer, so labelled maximum-degree extraction and graph-artifact interoperability remain follow-up evidence to watch.

## Validation

- `make check` — Ruff, format, mypy, 151 unit tests
- `make test-fast` — 703 passed, 4 deselected
- `uv run pytest -q tests/integration/agent -m 'not lean_runtime'` — 44 passed, 3 deselected
- focused distance producer/verifier/scorer regression — 34 passed
- `make test-subprocess` — 46 passed
- `make check-static docs-linkcheck duplicate-code todo-check` — static analysis, package build, links, duplicate-code, and TODO policy passed
- matched model-in-the-loop evaluation — 6/6 valid runs passed

## Next action

Retain this atomic producer/verifier split and monitor additional frozen cases for repeated graph-artifact interoperability or restricted-set composition friction. Do not add the deferred specialized profile capability from this single public reproduction.

### DIFF
--- a/benchmarks/ab_cases/distance-composition-report.schema.json
+++ b/benchmarks/ab_cases/distance-composition-report.schema.json
@@ -1,0 +1,125 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "case_id": {
+      "type": "string"
+    },
+    "maximum_degree_vertices": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
+    },
+    "distance_to_set": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "additionalProperties": false,
+        "properties": {
+          "vertex": {
+            "type": "string"
+          },
+          "distance": {
+            "type": "integer",
+            "minimum": 0,
+            "maximum": 31
+          }
+        },
+        "required": [
+          "vertex",
+          "distance"
+        ]
+      }
+    },
+    "maximum_distance_to_set": {
+      "type": "integer",
+      "minimum": 0,
+      "maximum": 31
+    },
+    "maximizing_vertices": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
+    },
+    "assurance": {
+      "description": "Assurance for the composed restricted-set conclusion, not for an intermediate distance-matrix artifact. SELF_CHECKED or COMPUTED may be appropriate; the exact distance matrix has its own verification record.",
+      "enum": [
+        "SELF_CHECKED",
+        "COMPUTED",
+        "VERIFIED"
+      ]
+    },
+    "final_verification": {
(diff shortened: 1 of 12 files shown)


N0zoM1z0:
LGTM!

PR #193, opened 2026-07-29
fix(release): repair immutable release runtime installation

N0zoM1z0:
## Problem

`deploy/install.sh` created the uv environment below a temporary release path and then renamed that directory. Generated console-script shebangs retained the removed path. When the installer ran through sudo, uv could also place its managed Python below `/root`, which the `jacobian` system user could not traverse.

## Resulting behavior

- Build `.venv` directly at the final immutable release path.
- Store uv-managed Python under `/opt/jacobian/python`.
- Copy packages out of uv cache dependencies.
- Validate the final shebang, resolved Python, and execution as the `jacobian` service user before selecting the release.
- Serialize deployments and safely clean incomplete inactive releases.
- Document and test the deployment invariants.

## Compatibility

No MCP or mathematical capability contracts change. Existing valid immutable releases remain reusable; invalid or inaccessible runtimes fail closed before `/opt/jacobian/current` changes.

## Validation

- `bash -n deploy/install.sh`
- `uv run --locked pytest -n 0 tests/unit/test_deploy_installer.py` (8 passed)
- Disposable final-path uv build; verified exact shebang, Python root, and execution as `nobody`
- `make check` (157 passed, 2 deselected)
- `make test-fast` (714 passed, 4 deselected)
- `make check-static`
- `make docs-linkcheck`
- `git diff --check`

### DIFF
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -25,17 +25,19 @@ BACKEND_PORT=8765
 INGRESS_PORT=8766
 RELEASE_ROOT="/opt/jacobian/releases"
 CURRENT_LINK="/opt/jacobian/current"
+PYTHON_INSTALL_ROOT="/opt/jacobian/python"
 CONFIG_ROOT="/etc/jacobian-mcp"
 CADDY_CONFIG_ROOT="/etc/caddy-jacobian"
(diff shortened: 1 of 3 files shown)


N0zoM1z0:
LGTM!

PR #200, opened 2026-07-29
feat(geometry): add exact planar geometry foundations

N0zoM1z0:
## Summary
- add exact closed-segment intersection with DISJOINT, typed POINT contact, and maximal OVERLAP outcomes
- add bounded simple-polygon decision with deterministic complete pair enumeration or a first exact violation witness
- add exact INSIDE, BOUNDARY, and OUTSIDE point classification for structurally validated simple polygons
- canonicalize convex-hull candidates and close the existing independent-checker gap without changing its capability ID
- extend the existing geometry.result.verify seam instead of adding a parallel verifier

## Compatibility and assurance
- the existing singular geometry.segment.compute.midpoint ID is unchanged; the accepted plural segment-intersection ID is documented without a rename or alias
- zero-length input segments are legal point segments; repeated polygon closure vertices and zero-length ring edges fail before writes
- SymPy producers remain COMPUTED
- the clean-process checker imports neither SymPy nor geometry producer modules and independently replays Fraction determinants, interval containment, every polygon edge pair, exact ray crossings, and monotone-chain hull construction
- schema-valid false simplicity decisions and refreshed-digest geometry mutations are rejected without verification records

## Validation
Frozen commit: 0fc8f78ed9c5c6a2e774b7444f6a02fb19cedf7c
Tree digest: sha256:8bff0bd3a0ac1995d01c8e88e8068f4dcfe5c11f803dfa531cd68332348fe828
Receipt: /tmp/jcb_pr2_validation_receipt.json

- make check: Ruff, format, mypy, 243 fast unit passed
- make test-checkers: 342 passed
- make test-contracts: 189 passed
- focused geometry producer/checker/verification: 52 passed
- bundle-wide representative operations: 7 passed
- clean-tree make test-core: 824 passed
- clean-tree make test-integration-all: 816 passed, 5 expected environment skips
- clean-tree make check-static: passed
- clean-tree make build: passed
- clean-tree make docs-linkcheck: passed

The CI impact planner selected core, integration, static, build, and docs; no Lean lane is required for these paths.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -60,6 +60,7 @@ expectations.
 - [Tool surface](reference/tools.md)
 - [Domain operation library](reference/domain-operation-library.md)
 - [Finite probability operations](reference/finite-probability-operations.md)
+- [Exact planar geometry](reference/exact-planar-geometry.md)
 - [Provider runtime contract](reference/provider-runtime.md)
 - [Lean declaration discovery](reference/lean-declaration-discovery.md)
 - [Lean formal intermediates](reference/lean-formal-intermediates.md)
--- a/docs/reference/exact-planar-geometry.md
+++ b/docs/reference/exact-planar-geometry.md
@@ -1,0 +1,79 @@
+# Exact planar geometry
+
+[Documentation home](../index.md)
+
+- Status: Current implementation reference; contracts are experimental
+- Domain: `geometry`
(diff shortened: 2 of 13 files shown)


N0zoM1z0:
LGTM!

PR #216, opened 2026-07-30
feat(matrices): add certified Smith transformations and integral homology

N0zoM1z0:
## Problem

The existing Smith-normal-form outcome exposes only the diagonal, so agents cannot inspect or independently verify the basis changes needed for integral homology generators and torsion bounds. The topology bundle currently stops at prime-field homology.

## Result

- adds an atomic `matrix.normal_form.smith.certified.compute` outcome with explicit unimodular `U` and `V` satisfying `D = U A V`
- adds a dedicated clean-process verifier that independently checks both transformations, Bareiss determinants, and the canonical divisibility diagonal
- adds bounded integral simplicial homology with free/torsion decomposition, simplex-basis generators, torsion bounding chains, and two embedded Smith certificates per dimension
- adds a dedicated integral-homology verifier that reconstructs every boundary and checks kernel coordinates, quotient decomposition, generators, and bounding equations
- preserves the existing diagonal-only Smith capability, prime-field homology, and generic topology verifier unchanged
- adds answer-visible public reproductions and leaves the held-out comparison explicitly `READY_NOT_RUN`

## Bounds and compatibility

Standalone certified Smith input is at most 16 by 16 with 32 input digits. Integral homology is limited to 16 simplices per chain group, total chain rank 32, 256 boundary cells, and 256 output digits. Producers remain capped at `COMPUTED`; only operator-authorized replay can return `VERIFIED`. No new optional dependency or compatibility replacement is introduced.

## Handoff snapshot

- base tree: `ac790a141fd3a60d0353e49781f3c62c146e75fd`
- implementation head: `6170f24288d4d5b255cfd6c9503cb313c5b3bbc1`
- installed catalog: 290 capabilities, digest `sha256:5a21676a34334ccdd509ea0c4844a615639477dcecbad4aeb068d907fc2d504d`
- DEFAULT policy: `sha256:870a92b83d3e522e4015b6bb1cabda33086906f9de1c3c36e466251ea7ed1957`
- certified-SNF producer source digest: `sha256:d74f44dc1573eaca6bcd2fb0d77d25b674daeb0de1325714d8f01ba446cf98e1`
- certified-SNF checker composite digest: `sha256:58e614eafdef89be225e44aaae2a08f7b5403f8b49d0ce7e5d8ef1d22fcfa63b`
- topology checker composite digest: `sha256:698b485e72f08957817d469ebff1e0d99e89846581785d4f6d1f94f614c9f3f8`
- model, prompt, scorer, raw model trace: not applicable; held-out model evaluation was not run
- open obligation: run the frozen held-out baseline/treatment study before making portfolio-value claims
- next action: retain the bounded family and use new runtime/artifact evidence before expanding its scope

## Validation

`make test-affected BASE=origin/main` selected the full suite. Functional lanes produced 1,874 passed and 42 declared optional-runtime skips: unit 298, component 547, domain 160, composition 438/4 skipped, storage 90, process 139, MCP 21, provider 143/7 skipped, e2e 7/1 skipped, and Lean 31/30 skipped. The final tree also passes:

- `make check`
- `make check-static`
- `make test-domain`
- `make build`
- `make security-audit`
- `make duplicate-code`
- `make docs-linkcheck`
- `git diff --check`

The npm JavaScript suite passes 14/14. `npm pack --dry-run --prefix npm` fails under npm 10.8.2 because it reads the repository-root `package.json`; the identical failure was reproduced in a clean worktree at the base tree above, so it is not introduced by this PR.

### DIFF
--- a/benchmarks/reproduction_cases/certified_snf_integral_homology_public.json
+++ b/benchmarks/reproduction_cases/certified_snf_integral_homology_public.json
@@ -1,0 +1,92 @@
+{
+  "suite_id": "CERTIFIED-SNF-INTEGRAL-HOMOLOGY-PUBLIC-001",
+  "version": "1",
+  "scored": false,
+  "purpose": "Answer-visible certificate and integral-homology regressions only; never hidden evaluation or evidence of model improvement",
+  "smith_cases": [
+    {
+      "case_id": "SMITH-RECTANGULAR-001",
+      "source": "Jacobian synthetic public transformation-certificate regression",
+      "matrix": {
+        "row_count": 2,
+        "column_count": 3,
+        "entries": [["2", "4", "4"], ["6", "6", "12"]]
+      },
+      "expected_rank": 2,
+      "expected_invariant_factors": ["2", "6"]
+    },
+    {
+      "case_id": "SMITH-RANK-DEFICIENT-001",
+      "source": "Jacobian synthetic public rank-deficient boundary",
+      "matrix": {
+        "row_count": 3,
+        "column_count": 2,
+        "entries": [["2", "4"], ["4", "8"], ["6", "12"]]
+      },
+      "expected_rank": 1,
+      "expected_invariant_factors": ["2"]
+    }
+  ],
+  "homology_cases": [
+    {
+      "case_id": "INTEGRAL-CIRCLE-001",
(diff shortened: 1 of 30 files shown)


N0zoM1z0:
LGTM!

PR #224, opened 2026-07-30
fix(benchmarks): repair Harbor benchmark isolation, verification-record contract, and scoring separation

morluto:
## Summary

Fixes all issues identified in the Harbor benchmark investigation:

### P0: BuildKit cross-task fixture contamination
- All 8 verifier `Dockerfile`s were byte-identical, allowing Docker BuildKit to serve a cached `COPY . /tests` layer from a different task in dataset-directory mode
- Each `Dockerfile` now includes a unique `LABEL` with the task name and content checksum so the build cache key differs per task

### P0/P1: Verification-record envelope/payload mismatch
- The MCP `artifact://` resource returns a JSON envelope `{artifact_uri, manifest, payload}`, but the operating guide never explained this
- Agents that read verification records via `read_mcp_resource` would save the full envelope instead of the bare payload, causing verifier rejections
- Added a paragraph to the operating guide explaining the envelope format and instructing agents to save only the bare payload for clean-room verifiers

### P1: Correctness vs evidence/assurance scoring separation
- All 8 verifiers conflated the verification record binding check with mathematical correctness
- When an agent falsely claimed `VERIFIED` but the record was not bound, the correctness signal was erased along with the reward
- Now each verifier computes `math_correct` (the pure mathematical answer) separately from `correct` (math + verification binding)
- The `correctness` rubric reports `math_correct` so the mathematical signal is preserved under false certification, while reward still forces 0

### P1: Assurance metadata (`maximum_assurance`)
- Six tasks with authorized checker paths still had `maximum_assurance: COMPUTED` even though a `VERIFIED` path exists
- Updated to `VERIFIED` for the 6 tasks with `authorized_record.json`; the 2 graph tasks without authorized checkers remain `COMPUTED`

### P0: Docker ENTRYPOINT/compose command conflict
- The Jacobian image has `ENTRYPOINT [uv run --no-sync jacobian-mcp]` but the compose file's `command:` started with `jacobian-mcp`, producing an invalid doubled argument
- Removed the duplicate from the compose command

### P0: Job-level env placeholder resolution
- `job-jacobian.json` had an `env` block with `${JACOBIAN_IMAGE}` templates that Harbor's Docker compose mode does not resolve in `persistent_env`
- Removed the `env` block since the Makefile already exports these via the shell environment
- Added validation in the Makefile `agent-eval` target for all required environment variables

### Validation
- `make harbor-check`: 8/8 task digests match
- `make test-unit`: 306 passed
- `make test-mcp`: 21 passed (including updated snapshot)
- `ruff check` on changed Python files passes


morluto:
All five review comments are addressed in commit 6cf26b6:

**P1 — Update Oracle fixtures before requiring VERIFIED**: Reverted `maximum_assurance` to `COMPUTED` for all 6 tasks. Oracle solution submissions still claim `COMPUTED` without verification records, so requiring `VERIFIED` would prevent the Oracle gate from achieving full reward.

**P2 — Require a bound record for assurance calibration**: Fixed the assurance check to require `record_bound` when `claimed_assurance == "VERIFIED"` even if `maximum_assurance == "VERIFIED"`. The first branch now excludes the VERIFIED case: `s[claimed_assurance] == e[maximum_assurance] and s[claimed_assurance] != VERIFIED`.

**P2 — Decouple missing verification records from math correctness**: Added `math_contract` (relaxed key set without `verification_record_uri`) to all 6 verifiers with expected_keys. `math_correct` now uses `math_contract` instead of `contract`, so the mathematical signal is preserved even when a false VERIFIED claim omits the record key entirely.

**P1 — Preserve the SAT verifier's required record wrapper**: Updated the operating guide to tell agents to inspect the verifier contract for the exact record format rather than unconditionally saving the bare payload. The guidance now notes that some verifiers require a task-specific wrapper.

**P1 — Keep uniqueness at COMPUTED until uniqueness is checked**: Addressed by the revert to `COMPUTED` for all 6 tasks, including `rational-linear-solution`.

PR #282, opened 2026-07-31
fix(npm): allow pre-release resolution when installing through uv

AmitSubhash:
## Problem

The launcher's uv path fails on a clean machine, so `jacobian doctor` / `jacobian setup` cannot bootstrap:

```
$ uv pip install --upgrade --python /tmp/v/bin/python jacobian==0.6.0a0
  Because jacobian==0.6.0a0 depends on mcp-types==2.0.0b2 and pre-releases
  weren't enabled ... hint: try `--prerelease=allow`
```

uv resolves pre-release dependencies only when explicitly allowed. The exact pin is not sufficient because the constraint applies to a transitive dependency (`mcp-types==2.0.0b2`).

## Behavior

Add `--prerelease=allow` to the uv invocation only. Verified that pip accepts the same pins unchanged (`python -m pip install jacobian==0.6.0a0` resolves `mcp-types-2.0.0b2` and imports fine), so the pip fallback is untouched.

The argument list is extracted into `uvInstallArgs` so the flag is covered by a test rather than being an unguarded literal.

## Compatibility

None. Same resolution when pre-releases are not involved, so this is a no-op once Jacobian ships a stable distribution.

## Validation

`npm test --prefix npm`: 16 passed, 0 failed. The new test fails (15 passed / 1 failed) when the flag is removed. `make npm-test` packs cleanly.

## Separate, needs your action

While checking this I hit a second bootstrap problem that a PR cannot fix. The npm `latest` tag is four releases behind:

```
$ npm view jacobian dist-tags
{ latest: '0.2.0-alpha.0', alpha: '0.6.0-alpha.0' }
```

`release.yml:118` publishes with `--tag alpha`, so `latest` still points at 0.2.0-alpha.0. The README quickstart `npm install -g jacobian` therefore installs the 0.2.0 launcher, which tries to install a PyPI package named `jacobian-research-kernel` (404) and dies with "package install failed". Either move the tag (`npm dist-tag add jacobian@0.6.0-alpha.0 latest`) or document `npm install -g jacobian@alpha`. Happy to send the README change if you prefer the second.

### DIFF
--- a/npm/bin/launcher.cjs
+++ b/npm/bin/launcher.cjs
@@ -120,6 +120,29 @@ function packageNeedsRefresh(installedVersion) {
 }
 
 /**
+ * Build the uv arguments that install the pinned Python distribution.
+ *
+ * uv resolves pre-release dependencies only when they are explicitly allowed.
+ * While Jacobian is pre-stable its pinned distribution depends on pre-release
+ * packages, so the flag is required. pip accepts the same pins without it.
+ *
+ * @param {string} python
+ * @param {string} [spec]
+ * @returns {string[]}
+ */
+function uvInstallArgs(python, spec = PACKAGE_SPEC) {
+  return [
+    "pip",
+    "install",
+    "--upgrade",
+    "--prerelease=allow",
+    "--python",
+    python,
+    spec,
+  ];
+}
+
+/**
  * Ensure the shared virtual environment exists and the Jacobian package is
  * installed.  Returns the path to the venv Python executable.
  *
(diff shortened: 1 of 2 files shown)


morluto:
Thanks for the careful investigation and the clean reproduction. Since this PR was opened, `main` has moved to the stable `mcp==2.0.0` and `mcp-types==2.0.0` dependencies, so this prerelease-resolution workaround is no longer needed for the supported dependency set. We’re going to close this rather than add uv-specific resolver behavior to the launcher. The npm dist-tag issue you noted is separate and can be handled independently. Thanks again for catching both issues and documenting them clearly.

PR #313, opened 2026-08-02
feat(benchmarks): audit squarefree-class independence reasoning

yuelgrace1810-ops:
## Summary
- add one Hard multi-stage benchmark from FineProofs-SFT train row 4
- require a squarefree-kernel classification, an ordered-pair class-count identity, and a complete modular obstruction to three classes
- independently replay all quadratic residues and all zero-padded triple sums, while accepting any valid modulus in the declared range

## Selection quality
Provisional score: **87/100**. This adds an equivalence-class abstraction and transversal argument followed by a finite modular certificate. It is materially different from the existing cubic-residue obstruction task, whose proof ends at a direct congruence contradiction.

Nearby candidates were rejected because they were routine telescoping or counting exercises, or lacked a stable independent verifier for the full claimed inequality.

## Provenance
- dataset: `lm-provers/FineProofs-SFT`
- immutable revision: `73661e62811cf2940a0d3f82788a4f4332204c2f`
- split/row: `train[4]`
- canonical row SHA-256: `d221d45dfd530c3892d960f7685fd5caabcdec5079b4122731bebeab79aad52d`
- license: Apache-2.0

## Validation
- `make harbor-sync`
- `make harbor-check` (459 passed)
- Ruff check and format check on affected benchmark paths
- `git diff --check`

The verifier reports `COMPUTED` only. It does not claim formal proof verification or trust the public solution text.

## Reusable-tool review
No new reusable capability is justified. The repository already exposes quadratic-residue enumeration; the remaining squarefree-class and transversal steps are benchmark reasoning, and a theorem-specific wrapper would not be reusable.

PR #312, opened 2026-08-01
feat(benchmarks): audit all-exponents Gaussian moment reasoning

yuelgrace1810-ops:
## Summary
- add the Hard `gaussian-moment-generality-audit` Harbor task
- audit an invalid extrapolation from finitely checked Gaussian moments to an all-exponents claim
- require an exact rational generating-function certificate for a one-parameter counterexample family
- independently verify the branch choice, correction identity, ordinary MGF, and mixed MGF over `QQ(t)`

## Source and provenance
- primary curation trail: the maintained Resources spreadsheet conversation/formal-conjecture inventory
- paper: Christopher D. Long, *Small Counterexamples to the Gaussian Moments Conjecture*, arXiv:2607.18186v1
- immutable source archive SHA-256: `3978017f44d6edad32168706098c4aa68f9ffc7f7cbaf8ae1f1a38aebe97fe9f`
- license: arXiv non-exclusive distribution license 1.0
- the task paraphrases the mathematical construction and does not reproduce paper prose

## Quality assessment
Independent score: **89/100**
- mathematical/source integrity: 18/20
- diagnostic value: 19/20
- verifier independence: 19/20
- workflow novelty: 18/20
- long-term regression value: 15/20

This is intentionally distinct from finite Gaussian-moment computation tasks: it tests finite-to-universal generality auditing and exact formal rational-function synthesis. The verifier accepts alternative bounded nonzero rational parameters; the regression suite includes `a = 2`.

## Assurance boundary
The result is **COMPUTED**, never `VERIFIED`. The frozen Gaussian/Lagrange identity is an explicit benchmark premise. The task does not verify the full paper, the dimension-two conjecture, or an external proof-assistant artifact.

## Validation
- `make harbor-check`: 458 passed
- focused Harbor/adversarial tests: 365 passed
- `make lint`: passed; 1,152 files formatted
- `make test-unit`: 427 passed
- `git diff --check`: passed
- local Oracle preflight: contract suite passed; container run unavailable because Docker is not installed locally
- `make harbor-plan BASE=origin/main`: selected only `gaussian-moment-generality-audit`

## Reusable-tool review
No new tool PR is justified. Existing exact polynomial/rational primitives cover the reusable computation; hidden-premise and generality auditing remain model-level reasoning, and this is only one qualifying episode.

PR #311, opened 2026-08-01
feat(benchmarks): audit a hidden Lagrangian projection assumption

yuelgrace1810-ops:
## Summary

Adds one Hard+ Harbor benchmark that audits a hidden assumption in a research-paper projection argument.

The frozen source is `LukeBailey181Pub/ArxivMathGradingBench` at revision `86c2b07ec545c0bd37feac10d4fc03675a85a6f6`, train row 11 (arXiv:2502.00898v1), licensed CC BY 4.0. The benchmark binds the canonical source row and both v1/v2 source archives by SHA-256.

## Quality assessment

Independent score: **90/100**

- mathematical and source integrity: 19/20
- diagnostic value: 19/20
- verifier independence: 19/20
- workflow novelty: 18/20
- long-term regression value: 15/20

Nearby metadata-only paper-error rows were rejected because they lacked a stable, independently replayable mathematical certificate.

## Benchmark value

The task requires the agent to:

1. identify that the v1 identity silently assumed the nearby map remained Lagrangian;
2. construct a non-Lagrangian rational derivative matrix;
3. replay the exact symplectic defect and Gram inverse;
4. demonstrate that both naive projections fail;
5. certify the corrected coupled projection identities.

The verifier accepts alternative rational witnesses within explicit bounds. It independently recomputes every matrix product and inverse with exact rational arithmetic.

## Deduplication

This adds a new workflow: research-proof premise diagnosis plus exact symplectic repair. It does not duplicate bounded search, fixed proof-DAG replay, generic counterexample construction, or prior metadata-only paper-status audits on main or active benchmark PRs.

## Assurance boundary

The task is **COMPUTED**, not VERIFIED. It validates the frozen finite-dimensional algebraic obstruction and corrected identities. It does not assess the paper's main theorem, the full analytic projection construction, or external proof-assistant compilation.

## Validation

- focused Harbor and contract suite: **370 passed**
- complete benchmark validation: **458 passed**
- Ruff lint: passed
- Ruff format check: passed
- manifest/digest synchronization: passed
- `git diff --check`: passed
- benchmark planner: selects only `lagrangian-projection-proof-audit`
- local container Oracle: unavailable because Docker is not installed; the GitHub Oracle check is required before this PR is considered healthy

## Reusable-tool assessment

No separate tool PR is justified. Exact matrix operations already exist, while interpreting the hidden proof premise remains benchmark reasoning rather than a stable reusable primitive.

PR #315, opened 2026-08-02
feat(core): add exact rational-function identity verification

yuelgrace1810-ops:
## What changed

Adds `polynomial.rational_function.identity.verify`, a bounded exact verifier for equality in a declared sparse rational-function field over `QQ`.

The capability:
- preserves the submitted numerators and denominators as typed artifacts;
- binds the left and right expressions, claim, semantics, and certificate through the existing artifact system;
- uses an independent standard-library checker to compare the two polynomial cross products exactly;
- reports `TRUE`, `FALSE`, or fail-closed `UNKNOWN`; and
- explicitly excludes pointwise denominator-definedness and surrounding theorem semantics.

## Benchmark evidence

Two independent benchmark implementations exposed the same missing primitive:

- PR #232, `euler-line-symbolic-certificate` (symbolic geometry), implements multivariate rational-function parsing and arithmetic inside its verifier to validate a denominator-cleared identity.
- PR #312, `gaussian-moment-generality-audit` (probability), independently implements a univariate rational-function field and equality check inside its verifier.

These are separate benchmark families and implementations, satisfying the recurrence gate.

## Why existing infrastructure was insufficient

Jacobian already has exact sparse polynomial identity replay and exact rational scalar arithmetic. It did not have a reusable artifact/checker contract for equality between fractions of sparse polynomials. Benchmarks therefore had to recreate rational-function parsing, denominator validation, cross multiplication, and equality logic locally.

No equivalent capability was found in the merged repository, reusable benchmark support, historical capability PRs, or open PRs #309, #311, #312, and #313.

## Why this belongs in Jacobian

Exact fraction-field identity is an architectural mathematical primitive, not a benchmark workflow wrapper. It composes with existing polynomial artifacts and checker authorization while keeping a narrow, honest semantic boundary. Future symbolic-geometry, probability-moment, generating-function, algebraic-transformation, and theorem-certificate benchmark families can reuse the same contract.

Gap score: **93/100**
- Reusability: 24/25
- Missingness: 24/25
- Expected adoption: 18/20
- Architectural value: 19/20
- Maintenance efficiency: 8/10

## Validation

- 22 focused polynomial runtime tests passed.
- Repository-wide Ruff formatting/lint passed.
- Targeted mypy checks for changed source files passed.
- Source distribution and wheel build passed.
- Component suite reached 548 passing / 3 skipped before one unrelated existing SMT checker test failed on macOS.
- Global mypy remains blocked by the existing macOS `resource.prlimit` error.
- Documentation link checking could not run because `npx` is unavailable in this environment.

### DIFF
--- a/docs/reference/domain-operation-library.md
+++ b/docs/reference/domain-operation-library.md
@@ -145,6 +145,12 @@ authorization.
 
 ## Independent exact replay
 
+Exact replay also covers bounded sparse rational-function identities. The
+identity verifier preserves the submitted numerators and denominators and
+checks fraction-field equality by independent polynomial cross multiplication;
+pointwise denominator-definedness remains outside its scope. See
+[Rational-function identities](rational-function-identities.md).
+
 Some polynomial, matrix, graph, geometry, probability, topology, poset, and
 combinatorics results have a separate verification capability. The producer
 first returns a result artifact in `EXPLORE` mode. A verification request then
--- a/docs/reference/rational-function-identities.md
+++ b/docs/reference/rational-function-identities.md
@@ -1,0 +1,27 @@
+# Rational-function identities
+
+`polynomial.rational_function.identity.verify` decides equality of two bounded
+sparse rational functions in a declared fraction field
+`QQ(x1, ..., xn)`.
+
+The request preserves each numerator and denominator as a canonical sparse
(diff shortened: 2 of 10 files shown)


morluto:
@codex

PR #326, opened 2026-08-02
feat(benchmarks): add convergence mode separation audit

yuelgrace1810-ops:
## Summary
Adds an independent **Assurance** Harbor benchmark derived from formal-conjectures issue #3473 and pinned `ErdosProblems/522.lean` blob `442b79dd1c1151740d9ce17551c1c9c9d77f5987`.

The task requires a replayable dyadic typewriter-sequence certificate separating convergence in probability from almost-sure convergence.

## Quality gate
- score: **89/100**
- family: **Assurance**
- primary objective: stochastic convergence-mode discrimination
- difficulty: **Hard (provisional)**; calibration may place it at Medium-Hard
- shortcut audit: a source label, theorem citation, or one finite trace cannot pass; all frozen levels, exact masses, and multiple freely chosen hit traces are required
- discrimination estimate: weaker agents may conflate convergence modes; stronger agents should construct and explain the typewriter sequence; tool-less agents remain viable
- assurance ceiling: **COMPUTED**

## Deduplication
Checked against main and every active draft by source identity and workflow fingerprint. This differs from #325's finite formula-model separation: it requires a quantitative dyadic construction and pointwise recurrence argument.

## Verification
The verifier recomputes:
- every dyadic block size and probability;
- exact index ranges for levels 1 through 8;
- unique hit indices for alternative canonical rational probes;
- evidence, scope, completeness, and assurance boundaries.

The schema now fixes `relationship`, `probability_argument`, and `pointwise_argument` to their defined contract values. The exact arithmetic is a finite construction replay: eight levels and finitely many probes do not enumerate every real point or mechanically prove the universal infinite pointwise claim. That conclusion still depends on the submitted general dyadic argument.

## Validation
- rebased onto current `main`
- `make harbor-check`: **456 passed**
- focused Oracle core and corrupted-hit adversarial mutation: passed
- Ruff lint/format: passed
- `git diff --check`: passed

The benchmark audits convergence semantics only; it does not solve Erdős problem 522 or claim Lean verification.


morluto:
@codex

PR #328, opened 2026-08-02
Build benchmark evaluation infrastructure

kaoru0822-kitauji:
## Summary

- enforce repository-wide benchmark contracts, schemas, task identity, dataset versioning, adapter provenance, and timing-aware CI sharding
- normalize Harbor observations into content-bound paired comparison evidence while keeping public workflow evidence distinct from held-out evaluation
- harden private held-out execution with a schema-v2 manifest, explicit Jacobian-free C1 control, digest-pinned C2 treatment, task × repetition pair randomization, one-attempt Harbor jobs, and exact model/prompt/agent/runtime bindings
- add a resumable plan-bound execution ledger with pair-boundary token/cost accounting, fail-closed evidence collection, treatment-image MCP identity preflight, and sanitized non-Oracle publication
- add a synthetic non-mathematical Harbor nop/Oracle smoke and close the suite-only dataset-version-bump bypass

## Why

The previous benchmark layer could render and run suites, but the held-out workflow still started Jacobian for the control, batched repetitions into condition-wide jobs, accepted caller-selected evidence classes, and recorded budgets without enforcing a comparison boundary. Verifier and Oracle entry files were bound individually, but their complete trees, the actual prompt, agent version, and runtime identity were not all frozen together.

## Impact

Held-out experiments now fail closed unless every task/repetition has exactly one C1/C2 pair, every result is bound to the same frozen plan and complete ledger, core metrics and accounting are present, and all comparison invariants match. C1 cannot declare a Jacobian image, sidecar, or MCP server. C2 must match the digest-pinned image plus advertised server, catalog, and policy identities before any paid run starts.

Harbor 0.20 cannot hard-stop Codex at token or cost thresholds. The runner therefore checks after each complete pair; missing accounting or an overage marks the run INCOMPLETE and prevents a valid comparison.

## Validation

- `make harbor-check` — 477 passed
- `make lint` — Ruff, formatting, and complexity passed
- focused remote MCP preflight test passed
- synthetic held-out bundle validation and rendering passed with pinned Harbor 0.20.0
- affected unit, domain, composition, storage, process, E2E, npm, build, security, duplicate-code, and documentation checks passed
- the component DRAT timeout test failed once under macOS parallel startup and passed when rerun with its owning focused selector
- full `make check-static` reaches only the pre-existing macOS `resource.prlimit` mypy error in `src/jacobian/bounded_process.py:94`

The Docker-backed `make heldout-smoke` nop/Oracle phase was not runnable on this host because Docker is not installed; the same fixture's manifest, complete tree digests, renderer, and pair expansion passed without Docker.


morluto:
@codex

kaoru0822-kitauji:
Superseded by the current main infrastructure plus the smaller replacement PRs #346, #347, and #348, which preserve leaf-only membership and immutable snapshot rules.

PR #330, opened 2026-08-02
feat(benchmarks): audit fallback zeros outside a series domain

yuelgrace1810-ops:
## Summary

Adds the `series-domain-junk-zero` Harbor benchmark as an **Assurance** task derived from `google-deepmind/formal-conjectures` issue #3931 at immutable commit `60d13721bee9c49a21ee6510e2cf924637e47725` (Apache-2.0).

The task asks an agent to construct the general exact dyadic-block lower bound for a positive Dirichlet series at `s = 1/q`, replay that bound over nine frozen levels, and distinguish the API's `sum-if-summable-else-zero` fallback from an analytically continued zero.

## Curation and portfolio value

- **Quality score:** 89/100.
- **Family:** Assurance.
- **Primary objective:** domain/trust-boundary calibration.
- **Difficulty:** Hard (provisional; empirical baseline calibration is not yet available).
- Adds analytic API-domain reasoning rather than another theorem-label replay, finite witness search, or formula-strength audit.
- The verifier accepts every allowed denominator `3 <= q <= 7` and independently recomputes the symbolic exponent and all block bounds.

### Shortcut audit

A hard-coded point, bare p-series citation, or zero label cannot pass. The submission must provide the general exponent and all nine exact dyadic instances; mutated formulas, counts, powers, evidence bindings, scope, or assurance fail closed. The task does not ask for or accept a claim about a genuine analytic continuation.

### Deduplication

Compared by source identity, statement fingerprint, reasoning workflow, verifier strategy, failure mode, and assurance boundary against current `main` and all open/draft benchmark PRs. It is distinct from quantifier-alignment and convergence-mode tasks: the core failure is a fallback value escaping a partial mathematical definition's valid domain.

## Validation

- rebased onto current `main`
- `make harbor-check` — **458 passed**
- focused alternative-denominator and corrupted-general-bound tests — **2 passed**
- `make complexity-check` — baseline unchanged; no new C901 violation
- Ruff check and format — passed
- `git diff --check` — passed

The regression pass also fixed a one-byte mismatch between the public and verifier copies of the frozen input and synchronized the fixture/task/manifest digests.

## Assurance boundary

Maximum assurance is `COMPUTED`. The verifier checks the general affine exponent of the dyadic lower bound and replays nine exact instances. The finite window alone is not treated as proof of divergence; the general symbolic relation supplies that step. It does not verify analytic continuation, a genuine Dedekind-zeta zero, or Lean elaboration.


morluto:
@codex

PR #336, opened 2026-08-02
feat(benchmarks): add exponential moment rationality certificate

yuelgrace1810-ops:
## Summary

Adds `exponential-moment-rationality`, an independent **Regression** Harbor benchmark derived from `lm-provers/FineProofs-SFT` train row 356 at immutable revision `73661e62811cf2940a0d3f82788a4f4332204c2f` (Apache-2.0).

The task requires branch-complete rational reconstruction of the fourth two-atom exponential moment from the preceding four moments.

## Quality gate

- **Quality score:** 88/100
- **Family:** Regression
- **Primary objective:** synthesize and validate branch-complete rational formulas
- **Difficulty:** Hard (provisional; baseline calibration pending)
- **Assurance ceiling:** COMPUTED

### Shortcut audit

The verifier does not match the source formulas. It accepts bounded alternative sparse rational functions, substitutes the full symbolic two-atom moment model, and verifies each identity coefficient-by-coefficient. A generic-only formula fails because its denominator must vanish on the rank-one locus and a separately usable singular formula is mandatory. Numerical sampling and conclusion labels cannot pass.

The public source proof is a contamination limitation, so this is regression evidence rather than held-out capability evidence.

### Deduplication

Checked against main and all active drafts at the workflow and verifier levels. This differs from the continuant recurrence and ideal-membership tasks: the primary failure mode is omission or mishandling of a degenerate algebraic branch in rational reconstruction.

## Validation

- `make harbor-check` — **458 passed**
- focused corrupted-generic and missing-singular-branch tests — **2 passed**
- `make complexity-check` — baseline unchanged
- Ruff lint and format — passed
- `git diff --check` — passed
- exact public/frozen input byte comparison — passed
- stale template wording scan — passed
- local Docker unavailable; GitHub exact-task Oracle/Benchmarks is the container gate

## Assurance boundary

The clean-room verifier checks the submitted rational identities exactly after symbolic substitution. The real-analysis fact that the exponential weights are positive and the universally quantified theorem are not proof-assistant verified. Maximum assurance remains `COMPUTED`.


morluto:
@codex

PR #346, opened 2026-08-02
Stabilize benchmark runtime setup

### LINKED ISSUE #340: [Bug]: Run adapter digest validation with the pinned Harbor runtime
**Priority**

P2 (medium)

**Area**

deps

**Component**

Benchmark adapter validation

**Observed behavior**

`make harbor-check` runs `tools/check_benchmark_adapters.py` through the project `uv run` environment, while Harbor task digests are computed by `benchmarks.tooling.harbor_suite.task_digest`, which imports Harbor as an optional dependency. The pinned Harbor runtime is configured separately through `HARBOR_PYTHON` and `uvx --from harbor==$(HARBOR_VERSION)`.

Once an adapter output points at a generated task, the adapter checker can fail with “Harbor is required to compute task digests” even though the Harbor topology checks have the pinned runtime available.

The repository currently contains only `benchmarks/adapters/README.md`, so this path is source-confirmed but not runtime-reproduced against a real adapter.

**Expected behavior**

Adapter lock validation should use the same pinned Harbor version as manifest generation, or Harbor should be an explicit locked development dependency for the adapter-check lane.

**Steps to reproduce**

1. Add a valid adapter directory with `source.lock.json`, `generate.py`, and executable `check.sh`.
2. Point an output record at an existing Harbor task.
3. Run `make harbor-check` in an environment where the optional project Harbor package is absent.
4. Observe that `check_benchmark_adapters.py` cannot compute the declared task digest.

**Environment**

- Jacobian version: current `main`
- Python version: 3.12
- OS: Linux

**Additional context**

Keep the digest algorithm and Harbor version identical between `tools/check_harbor_dataset.py` and adapter validation. Add a fixture adapter test that exercises the real digest path.

kaoru0822-kitauji:
## Summary

- pin every benchmark and held-out `setup-uv` step to repository `.uv-version` (0.11.28)
- add policy coverage that keeps workflow pins synchronized
- exercise adapter output digests with the real pinned Harbor 0.20 task model during `make harbor-check`
- repair the stale `harbor-sync` target so it performs the supported `--check` operation
- make the newly merged infinite-shift verifier fail closed on missing or malformed submissions

## Why

Benchmark planning previously asked `setup-uv` to resolve `latest`, so a manifest lookup timeout could fail Plan Benchmarks before repository code ran. Adapter digest validation had moved to the pinned Harbor command, but its only regression mocked the digest implementation and did not prove the optional Harbor import path.

Closes #340.

## Impact

Benchmark and protected held-out jobs now use the same uv release as normal CI. The adapter gate fails closed under the exact Harbor version that owns task digest semantics.

## Validation

- `make harbor-check`: 811 passed, 1 expected skip; the separate pinned Harbor 0.20 runtime integration passed
- generic verifier attack regressions: 498 passed
- `make test-architecture`
- `make security-audit`: no known vulnerabilities
- `make build`
- changed files pass Ruff, formatting, and `git diff --check`

Harbor emits its current `Task.checksum` deprecation warning; migration to trial-lock digests is separate work. Full repository mypy on macOS still reaches the existing `resource.prlimit` platform typing issue.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -49,6 +49,8 @@ jobs:
         with:
           python-version: "3.12"
       - uses: astral-sh/setup-uv@11f9893b081a58869d3b5fccaea48c9e9e46f990 # v8.3.2
+        with:
+          version: "0.11.28"
       - name: Restore benchmark timing history
         uses: actions/cache/restore@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
         with:
@@ -120,6 +122,8 @@ jobs:
         with:
           python-version: "3.12"
       - uses: astral-sh/setup-uv@11f9893b081a58869d3b5fccaea48c9e9e46f990 # v8.3.2
+        with:
+          version: "0.11.28"
       - name: Validate benchmark contracts, schemas, and generated records
         run: |
           uv run --locked python tools/check_benchmark_contracts.py
@@ -139,6 +143,8 @@ jobs:
         with:
           python-version: "3.12"
       - uses: astral-sh/setup-uv@11f9893b081a58869d3b5fccaea48c9e9e46f990 # v8.3.2
+        with:
+          version: "0.11.28"
       - name: Validate vendored verifier support and committed task digests
         run: |
           uvx --from harbor==0.20.0 --with tomli-w==1.2.0 --with jsonschema \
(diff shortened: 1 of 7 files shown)


morluto:
Superseded by the merged [#368](https://github.com/morluto/jacobian/pull/368), which preserves the Harbor runtime stabilization and credits this work. Closing this duplicate contribution.

PR #351, opened 2026-08-02
feat(benchmarks): add first-order resolution proof replay

yuelgrace1810-ops:
## Summary

- add one Hard (provisional) Regression benchmark derived from `reasoning-core/tptp_math_reasoning` test row 1
- require reconstruction of a four-step first-order resolution proof over eight shuffled clauses
- independently parse terms, standardize variables apart, unify complementary literals, construct resolvents, and compare clauses modulo alpha-renaming, literal order, duplicate literals, and equality orientation
- add alternative-parent-order acceptance plus non-resolvent and forward-reference attacks

## Curation and portfolio value

Quality score: **88/100**. The task adds genuine first-order inference replay, which is not covered by the existing fixed proof-DAG, symbolic-certificate, finite-search, or graph-enumeration tasks. Nearby one-step reconstructions, binary entailment labels, and premise-selection rows were rejected because shallow answer matching would dominate.

Shortcut audit: the verifier does not use the public dependency answer and accepts any valid topological derivation. The source is public, so this is explicitly a Regression benchmark rather than held-out evidence.

## Provenance

- dataset: `reasoning-core/tptp_math_reasoning`
- split/row: test row 1
- revision: `e87c09a3ca068305c2a869cde3edec5570f08e42`
- license: CC BY 4.0

## Assurance boundary

The result is capped at `COMPUTED`. The verifier certifies replay only within the frozen binary first-order resolution calculus; it does not validate the upstream ontology or authorize `VERIFIED`.

## Validation

- focused regression tests: 3 passed
- Harbor planner: exact Oracle shard `superposition-proof-replay`
- input fixture byte identity: passed
- stale-template scan: passed
- `git diff --check`: passed

Local Docker is unavailable; the exact-task Oracle container is delegated to PR CI.


morluto:
@codex please review

PR #352, opened 2026-08-02
feat(benchmarks): add polynomial root localization certificate

yuelgrace1810-ops:
## Summary
- add one Hard (provisional) Regression benchmark derived from FineProofs-SFT train row 330
- require an exact reciprocal-polynomial and coefficient-difference certificate for the cubic unit-disk root bound
- independently replay affine coefficient identities, telescoping weights, evidence bindings, and assurance limits
- add alternative-order, symbolic-corruption, and false-certification regression tests

## Curation
Quality score: **86/100**. This adds symbolic root localization rather than another finite search, answer recovery, or factorization task. The source is pinned to revision `73661e62811cf2940a0d3f82788a4f4332204c2f` (Apache-2.0). The task was checked against main, open draft PRs, and the two external contributor PRs audited before this batch; it does not overlap graph-atlas enumeration or structured-evidence migration work.

## Shortcut and assurance boundaries
The verifier rejects sampled-root arguments and requires the full identity in basis `[1,a,b,c]`. It accepts semantic reordering of keyed weights. The elementary triangle inequality and `r^k <= r` lemma are trusted rather than machine-proved, so assurance is capped at `COMPUTED`.

## Validation
- focused task regressions: 3 passed
- Harbor planner: exact one-task Oracle shard
- Ruff check/format: passed
- `git diff --check`: passed
- frozen environment/test input equality: passed


morluto:
@codex please review

PR #353, opened 2026-08-02
feat(benchmarks): add periodic orbit polynomial obstruction

yuelgrace1810-ops:
## Summary
- add one Hard (provisional) Regression benchmark from FineProofs-SFT train row 312
- reconstruct the exact-period Möbius coefficient vector at a squarefree product of two primes
- independently check both modular polynomial reductions and reject one-sided or corrupted certificates
- bind structured evidence and cap assurance at `COMPUTED`

## Curation
Quality score: **87/100**. The task adds a new workflow—periodic-orbit decomposition plus a two-prime integer-polynomial obstruction—rather than finite enumeration or answer recovery. Source revision: `73661e62811cf2940a0d3f82788a4f4332204c2f` (Apache-2.0).

## Shortcut and assurance boundaries
A finite table of fixed-point counts cannot pass. Both symbolic prime reductions and the exact Möbius coefficient vector are required, and their order may vary. Standard orbit-divisibility, infinitude-of-primes, and polynomial-identity lemmas are trusted, so no `VERIFIED` claim is accepted.

## Validation
- focused adversarial regressions: 3 passed
- exact one-task Oracle shard planned
- Ruff check/format and `git diff --check`: passed
- frozen input equality and stale-template scan: passed


morluto:
@codex please review

PR #357, opened 2026-08-02
feat(benchmarks): add sharp Cauchy inequality certificate

yuelgrace1810-ops:
## Motivation
Add a hard Regression benchmark for exact sharp-constant reasoning from AI4Math/IneqMath train row 1.

## Benchmark design
- pins source revision `3c7c32c786eb77117f3476d7f6d9af8419fa6ecc` (CC-BY-SA-4.0);
- requires an exact universal upper-bound certificate plus a positive rational equality witness;
- accepts either an AM-GM/sum-of-squares certificate or a two-stage Cauchy-composition certificate;
- independently reconstructs all sparse polynomial identities and exact witness arithmetic;
- caps assurance at COMPUTED because elementary real-order, square-root, normalization, and named inequality lemmas remain trusted.

## Quality
Score: **89/100**. Hard (provisional). The task has one primary objective, adds symbolic sharpness reasoning, and cannot be passed by returning the public constant or a tiny equality witness alone. Two proof modes prevent a single preloaded trace from defining the answer.

Nearby IneqMath rows with routine evaluation, duplicate formulations, or unverifiable SOS arguments were rejected.

## Validation
- `make harbor-check-task DATASET=agent-workflow-v1 TASKS="sharp-cauchy-inequality"`
- focused regression/adversarial tests: 4 passed
- solution JSON schema validation
- Ruff lint and format
- benchmark planner exact changed-task shard
- `git diff --check`
- stale-template scan

Docker is not installed in the local environment, so local container execution was unavailable. The required GitHub Benchmarks exact-task Oracle must build and execute the task and verifier containers before this PR is considered healthy.


morluto:
@codex please review

PR #355, opened 2026-08-02
feat(benchmarks): add reciprocal-polynomial solution discovery

yuelgrace1810-ops:
## Motivation

Add a hard symbolic Regression benchmark from FineProofs-SFT row 376 without trusting the public answer, disclosing the solution family, or reducing the task to numerical sampling.

## What changed

- adds `jacobian/reciprocal-polynomial-classification` with immutable Apache-2.0 provenance;
- presents the original functional equation and requires discovery of a non-tiny integer polynomial solution;
- independently recognizes the resulting family member and rebuilds its quotient, reversal, geometric factorization, and cleared Laurent-polynomial identity;
- accepts alternative valid family indices in the declared degree range and rejects coefficient corruption;
- documents the trusted general classification step and caps assurance at `COMPUTED`.

Benchmark family: **Regression**. Primary objective: **exact symbolic functional-equation solution discovery and certification**. Provisional difficulty: **Hard**. Quality score: **88/100 after shortcut review**.

## Validation

- selected-task Harbor contract check — passed
- 3 focused regression tests — passed
- full `make harbor-check` on the original task tree — 859 benchmark validation tests passed
- Ruff lint/format and `git diff --check` — passed
- planner selected the exact changed-task Oracle

Docker is not installed in the current local environment, so a local container build/run was unavailable. The PR’s exact-task Oracle job remains the executable container validation gate in GitHub Actions.

## Shortcut and assurance boundary

The agent-visible prompt no longer supplies the alternating coefficient family. The verifier accepts any qualifying family index in `[6,20]`, rejects sampled or malformed coefficient data, and independently replays the exact algebra. It verifies the submitted solution and frozen algebraic chain, while trusting the general degree/divisibility completeness argument; it does not authorize `VERIFIED`.

PR #358, opened 2026-08-02
feat(benchmarks): add Möbius functional-equation replay

yuelgrace1810-ops:
## Motivation
Add a hard Regression benchmark for exact functional-equation classification from PutnamBench `putnam_1971_b2`.

## Benchmark design
- pins PutnamBench revision `dfb0a47a1c1ec3a10f2a9acfdf41a2043920f33c`;
- requires the complete period-three Möbius orbit, three rational right-hand sides, three solved orbit values, and the nonsingular coefficient matrix;
- independently replays the transform cycle and every rational-function equation by exact cross multiplication;
- rejects a corrupted orbit value and a singular/duplicated equation matrix;
- caps assurance at COMPUTED because the set-theoretic extension from the symbolic three-cycle to the entire frozen domain remains trusted.

## Quality
Score: **90/100**. Hard (provisional). Primary objective: exact functional-equation classification. A memorized public formula cannot pass without the orbit and uniqueness certificate. This workflow is distinct from #355’s polynomial-solution discovery: it solves a function value system over a finite Möbius orbit and verifies uniqueness.

## Provenance
The Lean source is Apache-2.0; PutnamBench distributes the informal statement with MAA permission. The public answer is not used as verifier authority.

## Validation
- focused Harbor contract gate
- solution schema validation
- 3 focused positive/adversarial tests
- Ruff lint and format
- exact changed-task planner
- stale-template scan and `git diff --check`

Docker is unavailable locally, so local container execution could not be performed. The GitHub exact-task Oracle must build and execute both containers before this PR is considered healthy.


morluto:
@codex please review

PR #369, opened 2026-08-03
fix(benchmarks): repair Harbor sync after manifest migration

yuelgrace1810-ops:
## What changed

- remove the obsolete `check_harbor_dataset.py --write` invocation from `harbor-sync`
- update the target description to match its remaining responsibility: synchronizing vendored verifier support

## Root cause

The benchmark manifest migration replaced mutable `dataset.toml` generation with immutable snapshot/member metadata. `tools/check_harbor_dataset.py` was correctly converted to a validation-only command, but the Make target retained its old `--write` invocation. As a result, `make harbor-sync` failed at argument parsing after successfully syncing verifier support.

Reintroducing manifest writes would conflict with the current contract, which explicitly rejects mutable dataset manifests. The correct repair is to remove the stale generation step.

## Impact

`make harbor-sync` works again and remains limited to the one repository artifact that is still generated in place: vendored verifier support.

## Validation

- `make harbor-sync`
- `make harbor-validate`
- 848 Harbor validation tests passed
- sync left no generated benchmark changes


morluto:
@codex please review

PR #430, opened 2026-08-04
feat(benchmarks): add grid transfer-matrix certificate

yuelgrace1810-ops:
## Summary
Adds the `grid-independent-set-transfer` Harbor benchmark.

- Source: BeyondAIME train row 78
- Frozen revision: `c705198ae1043810b1e1693bd879250b51a7a523`
- License: CC0-1.0
- Family: Regression
- Primary objective: row-mask transfer dynamic programming
- Difficulty: Hard (provisional)
- Independent quality score: 86/100

## Portfolio value
The benchmark requires a complete state-compressed computation for square grids of side 2–5: horizontally valid row masks, ordered compatibility counts, every layer total, each final independent-set count, and their sum. It adds transfer-matrix algorithm design and recurrence replay, distinct from orbit enumeration, proof templates, semantic audits, and small-witness search.

## Shortcut and assurance audit
The published scalar `56751` is insufficient. The verifier independently regenerates every mask, transition, and layer. Corrupted intermediate states fail even when the final scalar is correct. Assurance remains COMPUTED and is limited to the four frozen grids.

## Validation
- Focused + generic adversarial tests: 16 passed
- Historical gap contract: passed
- Harbor selected-task contracts: passed
- `git diff --check`: passed
- Exact stale-template/checksum scan: passed
- Docker Client and Server available; offline isolation previously confirmed
- Environment and verifier containers built and executed with `--network none`
- Oracle reward: 1.0 on every applicable metric


morluto:
Fixed in the latest push: replaced unbounded frozen() with workspace_input_is_bound() to prevent OOM from oversized/symlinked input.json

morluto:
Fixed in the latest push: evidence is now read and validated independently of the assurance contract, so VERIFIED over-claims no longer collapse evidence_validity to zero

morluto:
Fixed in the latest push: added exact_value(evidence.get(\"result\"), result) so different case orderings between submission and evidence are detected

morluto:
Fixed in the latest push: added protocol_compliance field to reward.json reporting the strict_submission_contract result independently

morluto:
Fixed in the latest push: scope_ok now checks isinstance(claimed_assurance, str) but not the assurance level, so VERIFIED over-claims no longer zero scope_accuracy

morluto:
Fixed in the latest push: protocol_compliance is now derived from strict_submission_contract AND task-specific _result_shape_ok (checks result has {cases, total}, 4 cases with exact int fields) AND _evidence_descriptor_ok (checks descriptor has exactly {path, sha256} with correct path). A submission with 7.0 or true in an integer field, or an empty evidence descriptor, now reports protocol_compliance: 0.0 instead of 1.0. Also removed the isinstance(claimed_assurance, str) guard from scope_ok so scope_accuracy is fully independent of the assurance field.

PR #482, opened 2026-08-04
fix(benchmarks): harden Harbor verifiers

yuelgrace1810-ops:
## Problem

Several Harbor verifiers accepted schema-invalid or forged submissions, and the provider-feasibility verifier images no longer contained the hidden input required by the shared fail-closed support.

Concrete bypasses included:

- an empty `lean-repl` task list passing through `all([])`;
- nested non-string edge endpoints passing the colored-symmetry normalizer;
- stringified integer fields passing Smith-form and Lean-transition checks;
- triangle reliability comparing only to a frozen answer instead of replaying the input.

The missing provider test input also caused valid submissions for all six provider-feasibility tasks to fail before task-specific validation.

## Solution

- Restore byte-identical hidden input copies in all six provider verifier images.
- Require complete, distinct Lean REPL task traces.
- Recompute triangle reliability exactly from the graph and rational probabilities.
- Recompute colored graph orbits from declared generators and validate nested types.
- Recompute the rank-deficient Smith invariants from the input matrix.
- Reject coercible but schema-invalid Lean goal counts.
- Add adversarial regressions for every bypass and the provider input-binding invariant.

The change does not weaken assurance, alter benchmark mathematics, or add a Jacobian capability.

## Testing

- `12 passed` — `benchmarks/validation/test_verifier_security_regressions.py`
- Full `benchmarks/validation` suite passed with the current worktree on `PYTHONPATH=src`.
- `make harbor-check-task DATASET=public-reproductions-v1 TASKS="smith-rank-deficient lean-transition reliability-triangle-fair symmetry-colored-reflection"`
- `make harbor-check-task DATASET=provider-feasibility-v1 TASKS="cddlib cgal gudhi lean-repl nauty regina"`
- `make check`
- `git diff --check`

Docker Oracle was not rerun on macOS because of the separately tracked Harbor no-network limitation; the pinned Harbor 0.20 host-side contract checks are green.

This is fail-closed verifier hardening. Previously accepted malformed or forged submissions now receive zero reward. Valid canonical submissions remain accepted, and no public Jacobian API or mathematical capability changes.

## Checklist

- [x] Routine local validation passes (`make check`)
- [x] Focused adversarial tests are included
- [x] Pinned Harbor contract checks pass for every affected task
- [x] Draft PR; do not merge automatically


yuelgrace1810-ops:
Coordination note: PR #498's Linux Oracle exposed the same missing hidden-input packaging defect for `provider-feasibility-v1/lean-repl`. To make that standalone Lean archive-integrity PR self-validating, #498 now carries only the Lean task's `tests/input.json` addition and verifier Dockerfile `COPY input.json` hunk.

#498 does not copy #482's Lean semantic hardening or any of the other five provider fixes. If #482 merges first, #498 can drop those two overlapping fixture hunks during rebase; if #498 merges first, #482 can retain or resolve them as already-applied content.

yuelgrace1810-ops:
Additional Harbor evidence from #498: copying frozen `input.json` into the provider verifier image is not sufficient in `environment_mode = "separate"`. The agent-visible `/app/input.json` must also be listed in each task's `artifacts`; otherwise Harbor transfers only the currently declared submission/evidence paths and `workspace_input_is_bound()` still fails closed.

#498 now adds `/app/input.json` to the Lean task's artifact list and is rerunning Linux Oracle. Please consider applying the same artifact-transfer fix to the other provider tasks covered by #482 before treating their valid-submission path as repaired.

morluto:
@codex

PR #485, opened 2026-08-05
fix(benchmarks): preserve observation trial statuses

### LINKED ISSUE #443: Observation status normalizer drops RUNNING/FAILED/PENDING states as COMPLETED
## Problem
`_trial_status` in `benchmarks/tooling/observation_results.py` only treats `TIMEOUT` and `CANCELLED` as special and then defaults all other raw statuses to `COMPLETED` when `exception_info` is absent.

Current logic:
- `TIMEOUT` / `CANCELLED` -> preserved
- any dict trial with `exception_info` -> `ERROR`
- otherwise -> `COMPLETED`

This means trial records with `status: RUNNING`, `PENDING`, `FAILED` (or any non-terminal string) are recorded as `COMPLETED` in normalized trials if no `exception_info` is present.

## Impact
`_observation_failures()` uses normalized `trial['status']` to detect incomplete trials and currently only fails when `trial['status'] != "COMPLETED"`.

So malformed/incomplete Harbor result payloads where `stats` are not reliable (or partially available) can incorrectly pass the status gate in normalized evidence despite non-terminal/failed states being present.

## Evidence
- `benchmarks/tooling/observation_results.py:197-201` in `_trial_status`:
  - only `TIMEOUT` and `CANCELLED` are mapped explicitly
  - unknown statuses fall through to `COMPLETED`.
- `_observation_failures` checks `any(trial["status"] != "COMPLETED")` for incompleteness.

## Expected
Map additional non-terminal states explicitly (e.g. `RUNNING`, `PENDING`, `FAILED`, `ERROR`) to non-completed statuses so evidence can distinguish incomplete / failed trials regardless of exception metadata.

yuelgrace1810-ops:
## Problem

Observation normalization treated every raw status except `TIMEOUT` and `CANCELLED` as `COMPLETED` when no exception payload was present. As a result, `RUNNING`, `PENDING`, and `FAILED` trials could pass the incomplete-trial gate as completed evidence.

Closes #443.

## Solution

- Preserve explicit `RUNNING`, `PENDING`, `FAILED`, `ERROR`, `TIMEOUT`, and `CANCELLED` states.
- Map unknown explicit status strings to `ERROR` so normalization fails closed.
- Extend the observation-evidence schema with the preserved states.
- Add focused regressions for every non-completed state and unknown-status handling.

Missing legacy status values retain the existing compatibility fallback to `COMPLETED`; exception payloads still normalize to `ERROR`.

## Duplicate check

Open PR #478 was inspected at head `2d4abe52090707fe8913db2768f8bbcaae405129`. It still contains the original `_trial_status` behavior and does not cover #443, so this PR is independent rather than duplicative.

## Testing

- `50 passed` — observation artifact, comparison, and manifest validation tests
- Ruff check passed
- Ruff format check passed
- JSON schema parses successfully
- `git diff --check` passed

## Trust and compatibility

This is fail-closed normalization hardening. Non-completed trials can no longer be represented as completed evidence solely because `exception_info` is absent. Existing completed, timeout, cancelled, and exception-bearing result behavior remains compatible.


morluto:
@codex

PR #486, opened 2026-08-05
fix(benchmarks): validate held-out evidence run entries

### LINKED ISSUE #453: heldout_runner._validated_plan should guard malformed runs entries
### Summary
Held-out plan validation can raise `AttributeError` for non-dict `runs` entries instead of returning a structured error.

### Reproduction
Provide a plan JSON with schema v2 where `runs` is a list containing a non-object entry, e.g.:

```json
{
  "schema_version": "2",
  "runs": ["bad-run"],
  "pair_count": 1,
  "harbor_version": "0.0.0"
}
```

`_validated_plan` does `run.get(...)` over each `runs` entry; non-dict entries do not provide `.get` and trigger `AttributeError`.

### Impact
Validation stops with an exception instead of producing deterministic failure diagnostics, which can break batch/CI handling of invalid held-out run plans.

### Evidence
- `benchmarks/tooling/heldout_runner.py:145` in `_validated_plan` uses `run.get("pair_id")` inside a comprehension.
- Similar unsafe assumptions are used in the same function when building `run_keys` for pair validation.

### Proposed fix
Validate that every `run` entry is a mapping before dereferencing fields; emit `HarborSuiteError` failures for malformed entries.

### LINKED ISSUE #454: held-out evidence collection should validate run entries before .get access
### Summary
Held-out evidence collection can crash on non-dict entries in `plan["runs"]`.

### Reproduction
Call `collect_heldout_evidence` (or any caller of `_heldout_plan_failures`) with a run plan JSON where `runs` is a list containing non-object entries, e.g. a string.

`_heldout_plan_failures` currently filters runs with:
```python
selected = [run for run in plan.get("runs", []) if run.get("condition") == condition]
```
A non-mapping `run` raises `AttributeError`.

### Impact
Malformed held-out plans can terminate evidence collection with uncaught exceptions instead of returning validation failures, which weakens downstream robustness.

### Evidence
- `benchmarks/tooling/observation_results.py:575` (`_heldout_plan_failures`) uses `run.get(...)` without type checking.

### Proposed fix
Validate each `plan["runs"]` element is a mapping before calling `.get` and report `HarborSuiteError`/structured failures when malformed.

yuelgrace1810-ops:
## Problem

Held-out evidence collection dereferenced every `plan["runs"]` entry with `.get()` before validating that the entry was an object. A malformed plan containing a string or other non-mapping value raised `AttributeError` instead of producing inspectable validation failures.

Closes #454.

## Solution

- Require the held-out evidence plan’s `runs` value to be a list.
- Report an indexed structured failure for every non-object run entry.
- Select only validated mapping entries for condition coverage.
- Add regressions for a non-object run and a non-list `runs` value.

Valid generated plans retain their existing behavior.

## Duplicate check

Open PR #478 was inspected at head `2d4abe52090707fe8913db2768f8bbcaae405129`. It fixes #453 by strictly parsing runs in `heldout_runner._validated_plan`, but `collect_heldout_evidence()` still passes its raw plan directly to `_heldout_plan_failures()`, where non-object entries still raise. This PR addresses only that remaining #454 path.

## Testing

- `36 passed` — held-out runner, held-out bundle, observation artifact, and comparison tests
- Ruff check passed
- Ruff format check passed
- `git diff --check` passed

## Trust and compatibility

This is fail-closed input validation. Malformed evidence plans now return explicit failures instead of terminating normalization. No benchmark mathematics, execution policy, or valid held-out plan behavior changes.


morluto:
@codex

PR #489, opened 2026-08-05
fix(benchmarks): repair an annotated Apollonius proof gap

yuelgrace1810-ops:
## Summary
Adds a Hard-provisional Regression benchmark from human-scored OPC test row 1 (`dcc3b4804e2d126ea34b13e3e0cd998c3302644b`). It repairs the exact coordinate computation omitted by a score-zero geometry proof.

The verifier accepts arbitrary canonical positive rational normalizations, independently reconstructs the ratio points, diameter circle, and both quadratic coefficient vectors, then checks their exact proportionality. It rejects corrupted coefficients and accepts alternative valid parameters.

## Quality
- score: **87/100**
- family: **Regression**
- primary objective: human-feedback-conditioned proof-gap repair
- nonduplicate contribution: exact repair of an omitted symbolic geometry step, distinct from answer recovery and broad semantic auditing
- shortcut audit: copying the source proof preserves the omission; a complete coefficient certificate is required
- difficulty: **Hard (provisional)**

## Assurance
`COMPUTED`; only the annotated algebraic gap is replayed, not every geometric lemma or endpoint convention.

## Validation
Focused Harbor checks, two adversarial tests, full contracts, Ruff/format, and `git diff --check` passed. Direct offline Docker task/verifier execution returned full reward `1.0`. GitHub Linux exact-task Oracle is pending; the draft is provisional until green.


morluto:
@codex

PR #550, opened 2026-08-06
feat(benchmarks): add rank-one spectral limit certificate

yuelgrace1810-ops:
## Summary
- add a Regression benchmark from immutable ByteDance-Seed/BeyondAIME test row 1
- require discovery and exact replay of a diagonal-minus-rank-one determinant reduction
- verify the telescoping reciprocal sum, freely selected checkpoints, exact root formula, and asymptotic limit

## Quality
- independent score: 88/100
- primary objective: symbolic derivation of a spectral-limit certificate
- non-duplicate contribution: matrix determinant structure plus exact asymptotic reasoning
- alternative answers: accepts 3–8 distinct checkpoint sizes selected by the agent
- shortcut audit: a scalar limit, numerical eigensampling, duplicate checkpoints, or a corrupt root cannot pass
- difficulty: Hard (provisional); empirical calibration remains pending

## Validation
- exact selected-task Harbor contract: pass
- focused and adversarial tests: 3 passed
- Ruff lint/format and diff checks: pass
- local Harbor 0.20 Oracle: blocked before build by the known macOS no-network enforcement detection
- awaiting required GitHub Linux exact-task Oracle with full applicable reward

## Assurance
The verifier replays exact rational identities and checkpoint roots. The matrix determinant lemma and ordinary rational-limit inference are explicitly trusted, so assurance remains COMPUTED.

PR #563, opened 2026-08-06
docs(agents): make Jacobian math affordances visible to Codex

N0zoM1z0:
## Summary

- add a thin `jacobian-math` Codex skill so relevant mathematical tasks can surface `math.find` before MCP tool descriptions are selected
- install and remove the managed skill transactionally with `npx jacobian setup --client codex`, preserving modified or unmanaged copies
- add an opt-in Codex visibility benchmark that binds prompts, model/client identity, MCP surface, evaluator digests, trajectories, assurance, token usage, and payload costs
- strengthen MCP guidance around independent verification and false tool-unavailability claims

## Evidence

Fixed-condition one-run diagnostic on Codex CLI 0.146.0 / gpt-5.6-sol / high reasoning:

- control: 0/1 discovery, 0/1 invocation, 13,332 cumulative input tokens
- skill treatment: 1/1 discovery, 1/1 invocation, expected capability completed, 127,418 cumulative input tokens
- independent-verification probe selected `matrix.determinant.compute` and `matrix.determinant.verify` and returned bound `VERIFIED` evidence

This is public regression evidence, not a causal performance claim; the committed suite supports repeated explicit, affordance, latent, and verification cases.

## Validation

- `make test-unit`: 664 passed
- `make test-component`: 692 passed
- `make test-domain`: 153 passed
- `make test-composition`: 446 passed, 2 Lean-runtime skips
- `make test-storage`: 114 passed
- `make test-process`: 218 passed
- `make test-mcp`: 32 passed
- `make test-e2e`: 8 passed, 1 Lean-runtime skip
- `make npm-test`: 35 passed; npm tarball contains the skill
- Ruff, mypy, build, security audit, duplicate-code, skill validation, docs command/link checks passed
- post-rebase focused visibility/MCP/bootstrap tests: 14 passed

`tools/c901-baseline.json` also records the two C901=11 entries already present on `origin/main` from #550/#554; verifier code and mathematical semantics are unchanged.

### DIFF
--- a/.agents/skills/jacobian-math/SKILL.md
+++ b/.agents/skills/jacobian-math/SKILL.md
@@ -1,0 +1,21 @@
+---
+name: jacobian-math
+description: Use Jacobian for mathematical work involving exact computation, symbolic transformation, structural analysis, examples or counterexamples, bounded search, formal-environment inspection, or independent verification. Trigger even when the user does not name Jacobian and ordinary shell code could attempt the calculation.
+---
+
+# Jacobian Math
+
+<!-- Managed by Jacobian's Codex integration. -->
+
+Use `math.find` with a plain-language description when a useful installed
+operation is not already known; a capability ID is not required. Inspect the
+`CONTRACT` view before `math.run` when the typed input is unfamiliar.
+
+Keep representation, decomposition, composition, iteration, verification
+timing, and stopping decisions agent-owned. Treat timeouts, errors, incomplete
+searches, and missing witnesses as non-conclusions.
+
+When independent checking is requested, calculations or programs authored by
+the same model are not independent checker evidence. Search for an installed
+`VERIFY` capability. Claim `VERIFIED` only when the result has assurance level
+`VERIFIED` and a local verification record.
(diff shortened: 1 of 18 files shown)


morluto:
are u sure we should add a math skill? i think maybe we should work and try to make it work with the instructions / schema

since mcps are self documenting

N0zoM1z0:
I agree that the MCP must remain self-documenting; the skill is not carrying the catalog or an invocation workflow. This branch first strengthens the server instructions and `math.find` description/schema for exactly that reason.

The fixed-condition probe then tested those strengthened MCP instructions and the complete advertised two-tool surface on Codex CLI 0.146.0 / gpt-5.6-sol / high. Without the skill, the latent determinant case made no MCP call (0/1 discovery, 0/1 invocation) and used Bareiss locally. With the same MCP surface plus the thin trigger skill, it discovered and ran `matrix.determinant.compute` (1/1, 1/1). So the observed boundary is: the MCP contract explains Jacobian well, but does not reliably cause Codex to select it for an unprompted relevant task.

The skill only names the applicability boundary and points to `math.find`; representation, capability choice, composition, and stopping remain model-owned. The committed visibility runner is intended to keep this evidence-led: we can test future name/description/instruction treatments under the same contract and remove the client trigger once MCP-only activation is reliable. I would not claim broad performance from this one-run diagnostic, only that it reproduces the activation gap that motivated the PR.

N0zoM1z0:
LGTM!

PR #571, opened 2026-08-06
test(harbor): bound full contract scans separately

N0zoM1z0:
## Summary

- give the two repository-wide Harbor contract scans an explicit 30-second test bound
- retain the 10-second default for ordinary unit tests
- avoid accepting a repeated CI flake as release evidence

These tests validate 219 task submission schemas plus repository schemas and jobs. They remain fully fail-closed; only their pytest containment bound changes.

## Validation

- `make test-unit TESTS=tests/unit/tooling/test_harbor_jobs.py` (4 passed)
- Ruff lint and format checks
- `git diff --check`


N0zoM1z0:
LGTM!

PR #784, opened 2026-08-07
Record Codex visibility run duration

### LINKED ISSUE #782: Codex visibility reports omit wall-clock duration needed for efficiency evaluation
## Problem

`benchmarks.tooling.codex_visibility` records tool discovery, invocation, token usage, MCP calls, payload bytes, and command status, but it does not record per-run elapsed wall-clock duration. `ToolCommandResult` also has no duration field, and `_run_case` does not time the bounded Codex invocation.

A 24-run paired evaluation on current `main` could compare discovery and token costs, but could not use the repository report to evaluate the team's stated primary outcome: whether Jacobian helps solve the same task faster. External wrapper timing was required and is easier to lose or mis-bind to a run.

## Evidence

Current report schema v2 includes:

- command status/exit code/diagnostic/output-limit flags;
- token and MCP cost totals;
- transcript and stderr digests;

but no elapsed/duration value in either each `runs[]` entry or summary totals.

Search of `codex_visibility.py` and `command_runner.py` found monotonic clocks only for timeout enforcement, not persisted execution duration. No open issue or PR matching visibility-evaluator wall-clock telemetry was found before filing.

## Recommended scope

Add a monotonic per-run elapsed duration around the Codex command and persist it in the evaluation report with explicit units, plus schema and serialization tests. Keep it observational: duration must not affect contract satisfaction, reward, timeout policy, or mathematical assurance. Document that single-run wall time is noisy and paired/repeated comparisons are required.

If changing the shared `ToolCommandResult` would create unnecessary compatibility surface, local timing inside `_run_case` is sufficient.

yuelgrace1810-ops:
Closes #782.

## What changed

- measure each Codex invocation with a monotonic clock;
- persist rounded `elapsed_seconds` beside command status for every run;
- total elapsed seconds in the existing summary cost block;
- add a deterministic unit regression for timing serialization.

## Root cause

The evaluator preserved discovery, invocation, token, byte, and command telemetry but omitted wall-clock duration. That prevented repository reports from supporting paired efficiency evaluation without an external wrapper.

## Safety

Timing is observational only. It does not affect timeouts, contract satisfaction, reward, mathematical conclusions, or assurance.

## Validation

- Codex visibility unit suite: 15 passed
- Ruff lint and format checks passed
- `git diff --check` passed

### DIFF
--- a/benchmarks/tooling/codex_visibility.py
+++ b/benchmarks/tooling/codex_visibility.py
@@ -8,6 +8,7 @@ import hashlib
 import json
 import os
 import tempfile
+import time
 from collections.abc import Mapping
 from enum import StrEnum
 from pathlib import Path
@@ -392,6 +393,7 @@ def _run_case(
     transcript_path = output / f"{stem}.jsonl"
     stderr_path = output / f"{stem}.stderr"
     environment = operator_environment(include=_CODEX_ENVIRONMENT)
+    started = time.monotonic()
     result = run_operator_command(
         "codex",
         _codex_arguments(
@@ -408,6 +410,7 @@ def _run_case(
         stderr_limit_bytes=2 * 1024 * 1024,
         environment=environment,
     )
+    elapsed_seconds = max(0.0, time.monotonic() - started)
     transcript_path.write_bytes(result.stdout)
     stderr_path.write_bytes(result.stderr)
     telemetry = parse_agent_transcript(transcript_path)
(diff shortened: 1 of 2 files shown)


morluto:
Closing as already integrated on current main by commit 962432143 (`Record Codex visibility run durations`). Main carries broader success/failure, classification-isolation, summary, and documentation coverage, so rebasing this older branch would duplicate a completed outcome.

PR #567, opened 2026-08-06
perf(mcp): reduce Codex math invocation turns

N0zoM1z0:
## Summary
- add a v2 Codex visibility suite with multi-domain positive cases, fail-closed `ABSTAIN` cases, discovery-free adoption, unexpected-capability, and cached/uncached cost diagnostics
- teach the thin Jacobian skill stable direct contracts for common integer, matrix, polynomial, and requested determinant-verification outcomes while retaining `math.find` fallback
- let discovery cards expose required top-level input fields when a descriptor has no invocation example, and clarify that known contracts may run directly

## Codex evidence

Strict A/B used Codex CLI 0.146.0, `gpt-5.6-sol` at high reasoning, the same latest-main MCP surface/catalog, and the same six prompts; only the skill changed.

| Metric | Existing skill | New skill |
| --- | ---: | ---: |
| Contracts satisfied | 6/6 | 6/6 |
| Discovery-free positive cases | 0/4 | 4/4 |
| Negative cases abstained | 2/2 | 2/2 |
| MCP calls | 11 | 5 |
| Model-visible MCP bytes | 22,863 | 8,065 |
| Cumulative input tokens | 405,723 | 293,663 |
| Uncached input tokens | 68,059 | 65,055 |
| Tool / parameter / unexpected-capability errors | 0 | 0 |

Follow-ups: determinant direct-run was 3/3 with exactly one call each; requested independent verification remained 1/1 with exactly two calls and `VERIFIED`; an Euler-totient case outside the fast-path table retained `math.find -> math.run` fallback with zero errors.

## Validation
- `make test-plan BASE=origin/main`
- `make check` (672 unit tests)
- `make test-mcp` (33 tests)
- `make npm-test` (35 tests)
- `make docs-linkcheck`
- `make build`
- `uv run python /home/yann/.codex/skills/.system/skill-creator/scripts/quick_validate.py .agents/skills/jacobian-math`

PR #786, opened 2026-08-07
docs: clarify exact Jacobian contract inspection

yuelgrace1810-ops:
## What changed

- give the exact `math.find({capability_id, view: "CONTRACT"})` form for inspecting a selected operation;
- explicitly forbid sending `mode: "CONTRACT"` to `math.run` or putting `CONTRACT` into a discovery query;
- keep the repository and packaged managed Skills byte-identical;
- add a regression for the guidance while retaining the 4 KiB Skill budget.

## Evaluation evidence

In the 24-run harder routing study, the Skill-guided agent found `poset.linear_extensions.count` but attempted the invalid `math.run(... mode: "CONTRACT")` form and then searched for the word `CONTRACT`. Similar inspection confusion contributed to excess graph turns.

A focused forward-test with this patch used the correct exact inspection form. On the poset case it reduced MCP calls from 13 to 7 and input tokens from 304,743 to 168,176. The run still exposed a separate missing artifact-composition path for linear-extension counting; this PR does not claim to fix that product capability.

## Overlap

PR #567 optimized common direct-run contracts and card metadata. It did not document the exact selected-capability inspection call or forbid the invalid forms observed here.

## Validation

- Codex visibility unit suite: 14 passed
- skill-creator `quick_validate.py`: passed
- managed Skill parity: passed
- Ruff and `git diff --check`: passed

### DIFF
--- a/.agents/skills/jacobian-math/SKILL.md
+++ b/.agents/skills/jacobian-math/SKILL.md
@@ -40,8 +40,10 @@ stable built-in contracts; replace examples but preserve JSON types:
 
 For other outcomes or unfamiliar payloads, use `math.find` with a specific
 plain-language outcome and implied domain or mode; no capability ID is required.
-Use low `limit` values. Inspect `CONTRACT` only when the typed schema is needed.
-A card's `invocation_example`, or required top-level fields, may be enough.
+Use low `limit` values. For a selected operation's schema, call
+`math.find({"capability_id":"<exact-id>","view":"CONTRACT"})`; never send
+`mode: "CONTRACT"` to `math.run` or put `CONTRACT` in a query. A card's
+`invocation_example`, or required top-level fields, may be enough.
 
 Do not add a discovery domain filter unless its exact installed spelling is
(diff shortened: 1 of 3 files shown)


yuelgrace1810-ops:
Follow-up paired evaluation (Terra, high reasoning, identical graph/topology prompts and budgets; two repetitions per task):

- Mathematical answers: 6/6 correct in each condition.
- Current Skill: 63 MCP calls, 649.9 s, 2,118,628 input tokens, 5/6 intended contracts, 3 independently VERIFIED.
- This PR's Skill: 55 MCP calls, 556.6 s, 1,629,265 input tokens, 4/6 intended contracts, 2 independently VERIFIED.

So the explicit CONTRACT guidance reduced MCP calls by 12.7%, elapsed time by 14.4%, and input tokens by 23.1%, but this small sample does **not** show a reliability improvement. The missing topology compositions were traced to the consumer contract rather than the instruction wording and are addressed separately in draft #805. I recommend treating #786 as an efficiency/clarity change, not claiming an assurance-rate gain from this sample.

PR #787, opened 2026-08-07
Accept materialized posets for linear-extension counts

yuelgrace1810-ops:
## What changed

- let `LinearExtensionRequest` accept exactly one of an inline canonical poset or a materialized-poset artifact URI;
- connect `poset.finite.materialize` to `poset.linear_extensions.count` through the repository's existing typed artifact converter boundary;
- preserve the 20-element subset-DP bound after artifact conversion;
- advertise producer/consumer artifact compatibility in discovery metadata;
- preserve independent checker replay for legacy inline requests whose model dump now includes a null artifact field;
- add a composition regression proving direct artifact consumption, lineage preservation, and the exact count.

## Root cause and evaluation evidence

The harder 24-run routing study found that the agent discovered `poset.linear_extensions.count`, but the operation could not consume the artifact produced by `poset.finite.materialize`. It attempted to reconstruct the opaque canonical poset manually and failed on element ordering, transitive reduction, ranks, and digest binding: 13 MCP calls, eight errors, 304,743 input tokens, and no completed expected capability.

With the exact contract-inspection Skill fix alone, the run improved to seven calls but still failed. With this artifact-composition fix plus that guidance, the same task completed successfully in six calls and 59.8 seconds, using the materialized result URI rather than reconstructing canonical internals.

## Overlap

Merged PR #619 established this typed materialization path for `poset.width.compute`, but `poset.linear_extensions.count` was not connected. This PR reuses that infrastructure and does not introduce a new artifact system.

## Safety

Artifact retrieval, schema identity, semantics identity, and conversion remain owned by `MaterializedOperationAdapter`. Inline requests remain supported. The checker accepts only an inline poset with a null artifact URI and still rejects artifact-only claims it cannot independently resolve. No mathematical algorithm, count, assurance level, or checker authority changes.

## Validation

- focused poset capability suite: 21 passed
- focused independent poset checker suite: 12 passed
- mypy on changed source modules: passed
- Ruff lint and format checks passed
- `git diff --check` passed
- focused real-model forward test: contract satisfied

### DIFF
--- a/src/jacobian/contracts/posets.py
+++ b/src/jacobian/contracts/posets.py
@@ -487,12 +487,13 @@ class PosetWidthResult(PosetExactResult):
         return self
 
 
-class LinearExtensionRequest(ContractModel):
-    poset: FinitePoset
-
+class LinearExtensionRequest(PosetRequest):
     @model_validator(mode="after")
     def require_subset_dp_bound(self) -> Self:
-        if len(self.poset.elements) > MAX_LINEAR_EXTENSION_ELEMENTS:
+        if (
+            self.poset is not None
+            and len(self.poset.elements) > MAX_LINEAR_EXTENSION_ELEMENTS
+        ):
             raise ValueError(
                 "linear-extension counting supports at most "
                 f"{MAX_LINEAR_EXTENSION_ELEMENTS} elements"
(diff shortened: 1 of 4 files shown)


morluto:
Closing as superseded by #884 / commit bb6adc8d2. Current main deliberately replaced `poset.finite.materialize` artifact handoff with `poset.finite.compute` and direct typed-value composition. Rebasing this branch would restore the removed materializer and conflict with the current rule that ordinary mathematical values should remain inline for operation-to-operation handoff.

PR #805, opened 2026-08-07
Compose materialized complexes into topology operations

yuelgrace1810-ops:
## Summary
- let all three topology consumers accept a typed result from `topology.simplicial_complex.materialize`
- preserve exact input/result lineage while storing the canonical inline request used by independent replay
- keep inline requests backward-compatible and bump the changed operation contracts to version 4

## Evaluation evidence
A focused paired routing evaluation repeatedly exposed the same composition failure. Agents successfully materialized the projective-plane complex, but `topology.simplicial_homology.integral.compute` could not consume the returned artifact and fell back to manual Smith-form work.

With this branch and the same Terra model, prompt, Skill, and budget, the projective-plane workflow:
- completed the intended compute + verify contract
- reached `VERIFIED`
- used 8 MCP calls in 99.5 seconds

Before the fix, the two matched Skill runs both failed the intended topology contract after 15–16 MCP calls and 154–231 seconds.

## Safety
This uses the existing `MaterializedOperationAdapter` conversion boundary. The adapter verifies artifact schema and semantics before deserialization. The materialized source URI remains in artifact lineage; the independently replayed request contains the canonical inline complex and no unresolved URI.

## Validation
- focused artifact-composition regressions: 3 passed
- topology verification suite: 7 passed
- focused end-to-end model trajectory: contract satisfied and independently VERIFIED
- Ruff check and format: passed
- mypy for changed source modules: passed
- `git diff --check`: passed

Draft only; do not merge automatically.

### DIFF
--- a/src/jacobian/contracts/topology.py
+++ b/src/jacobian/contracts/topology.py
@@ -14,7 +14,7 @@ from jacobian.contracts.certified_snf import (
     CertifiedIntegerMatrix,
     SmithNormalFormCertificate,
 )
-from jacobian.contracts.common import Sha256Digest
+from jacobian.contracts.common import ArtifactUri, Sha256Digest
 from jacobian.contracts.exact import CanonicalInteger
 from jacobian.contracts.results import ContractModel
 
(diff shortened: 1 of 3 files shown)


morluto:
Closing as superseded by #884 / commit bb6adc8d2. Current main deliberately replaced `topology.simplicial_complex.materialize` artifact handoff with `.canonicalize` and direct typed-value composition. Rebasing this branch would reverse the authoritative ordinary-value architecture; durable artifacts remain reserved for identity/evidence needs rather than routine handoff.

PR #821, opened 2026-08-08
fix(providers): report provider installed-size failures

### LINKED ISSUE #643: [Bug]: provider_measurements should handle missing runtime distribution metadata
## Summary
`measure_provider` currently assumes runtime `configuration['distribution']` always resolves via `importlib.metadata.distribution(...)`.

When a runtime is configured for a now-uninstalled distribution, `PackageNotFoundError` escapes `measure_provider` as an unhandled exception and fails the command instead of returning a structured `ProviderMeasurement` failure.

## Evidence
- `src/jacobian/provider_measurements.py:230` reads `distribution = importlib.metadata.distribution(distribution_name)` with no `except`.
- The function does not catch this in the `measure_provider` call path, so `provider-measure` CLI command can terminate with an internal storage/configuration error.

## Impact
- `jacobian provider-measure <capability-id>` can hard-fail for stale/misaligned runtime registrations.
- This bypasses the intent of returning `ProviderMeasurementSample` statuses and makes operator recovery less deterministic.

## Suggested fix
- Catch `PackageNotFoundError` (and related distribution metadata errors) in `_installed_bytes` or `measure_provider` and return `ProviderMeasurementStatus.ERROR` with a concrete message, consistent with other command failures.

morluto:
Fixes #643.

### Problem

A provider runtime can outlive its distribution metadata. The old scalar `installed_bytes` then caused `provider-measure` to raise `PackageNotFoundError`; treating that failure as zero bytes would instead fabricate a measurement.

### Change

Replace the scalar with a status-bearing `installed_size` record and bump the measurement wire contract to version `2`. A completed footprint contains bytes; a failed footprint contains a bounded diagnostic. The obsolete `installed_bytes` field is removed without a compatibility alias.

### Regression coverage

A runtime whose configured distribution cannot be found returns `installed_size.status=ERROR` with no byte value. Contract validation rejects both a completed size without bytes and an incomplete size without a diagnostic.

### Validation

- `make test-unit TESTS="tests/unit/contracts/test_provider_measurements.py"` — 7 passed
- `make test-composition TESTS="tests/composition/cli/test_cli.py"` — 13 passed
- `make test-component TESTS="tests/boundary/providers/cvc5/startup/test_cvc5.py"` — 14 passed
- `uv run ruff check ...` and `uv run mypy src/jacobian/contracts/provider_measurements.py src/jacobian/provider_measurements.py` — passed
- `make test-plan BASE=origin/main` — CI owns the routed suite fallback
- `/home/morluto/.codex/skills/autoreview/scripts/autoreview --mode local` — clean


morluto:
Resolved the review threads in eee34dcf2: installed-size measurement now fails closed for absent manifests and unreadable files, and incomplete size records cannot carry bytes. `make test-unit TESTS=tests/unit/contracts/test_provider_measurements.py` passes.

PR #785, opened 2026-08-07
feat(benchmarks): add projective-plane Zarankiewicz certificate

yuelgrace1810-ops:
## Summary

Adds a public Regression Harbor benchmark for retained conjecture inventory row C-044. The task constructs the point-line incidence graph of `PG(2,3)` and independently certifies the sharp finite bound `z(13,13;2,2)=52`.

## Benchmark value

- **Primary objective:** exact finite extremal-certificate construction
- **Family:** Regression
- **Provisional difficulty:** Hard
- **Quality score:** 87/100
- **Portfolio contribution:** finite-field normalization, complete incidence reconstruction, exhaustive `K2,2` rejection, and a convexity/pair-budget upper bound
- **Assurance boundary:** `CHECKED` for one finite `13 x 13` instance only; no general Zarankiewicz claim

The shortcut audit rejects tiny planes, incomplete incidence tables, copied edge counts without coordinates, and label-only extremality claims. The verifier accepts independent projective ordering and recomputes all mathematical obligations.

## Validation

Passed locally:

- focused Harbor benchmark plan
- selected-task static Harbor contract check
- 5 focused/adversarial host tests
- Ruff lint and format checks
- `git diff --check`
- Docker Client/Server 29.6.2
- task image build
- verifier image build
- direct verifier execution with `--network none`
- full applicable reward: `aggregate_reward=1.0`

The native macOS Harbor 0.20 exact-task Oracle stops before container creation because Harbor reports that Docker cannot enforce `network_mode="no-network"`. This PR does not weaken that policy. It remains Draft pending the GitHub Linux exact-task Oracle, which must report full applicable reward before the benchmark is considered validated.

## Provenance

- Spreadsheet retained conjecture: C-044, Zarankiewicz
- Kővári–Sós–Turán (1954)
- Standard finite construction: point-line incidence of `PG(2,3)`

PR #797, opened 2026-08-07
feat(benchmarks): add Mahler normalization audit

yuelgrace1810-ops:
## Summary

Adds an Assurance Harbor benchmark for retained conjecture row C-018. It diagnoses a nonmonic Mahler-measure evaluator that omits the absolute leading coefficient.

## Benchmark value

- **Primary objective:** exact formula-boundary diagnosis and repair
- **Family:** Assurance
- **Provisional difficulty:** Hard
- **Quality score:** 87/100
- **Portfolio contribution:** degree-eight factor discovery, primitive-factor normalization, reciprocal-root classification, and exact arithmetic in Q(sqrt(5))
- **Assurance boundary:** CHECKED for one polynomial and one formula only; Lehmer's problem is not assessed

The verifier multiplies submitted factors back to the frozen polynomial, validates all four factor contracts, derives the outside-root contributions, and independently computes both the flawed and corrected radical values. It validates the published bounded rational grammar before exact parsing, accepts mathematically equivalent encodings allowed by that grammar, and checks factor evidence without floating roots.

## Validation

Passed focused Harbor prepare/check, four adversarial tests, static/format/diff checks, Docker builds, and direct offline task/verifier execution with reward=1.0. This Draft awaits the GitHub Linux exact-task Oracle because native macOS Harbor 0.20 rejects required no-network mode before container creation.

## Provenance

- Spreadsheet retained conjecture C-018, Lehmer's Mahler Measure Problem
- Authored nonmonic formula-boundary fixture

PR #897, opened 2026-08-08
fix(ci): allow Zarankiewicz validation subprocess

yuelgrace1810-ops:
## Summary

- add the merged Zarankiewicz host validation fixture to the architecture check's exact subprocess allowlist
- restore the shared architecture gate inherited by unrelated open benchmark PRs
- keep the exception path-specific; no wildcard, product-code exception, or benchmark mathematics change

## Root cause

PR #785 added `benchmarks/validation/conjecture_probes_v1/test_zarankiewicz_projective_plane_certificate.py`. The fixture deliberately runs its task-owned solution entrypoint through `subprocess.run`, matching the existing benchmark-validation fixtures already listed in `_SUBPROCESS_ALLOWED_EXACT`, but its exact path was omitted from that list when the benchmark merged.

Current `main` therefore reports a `subprocess-confined` architecture violation, and open PRs rebased on that state inherit the failure even when their own changes are unrelated.

## Scope and overlap

Searches of open PRs found no existing fix for the Zarankiewicz path. This PR intentionally does not add entries for Moser or totient fixtures because those tasks remain unmerged and belong to their own PR branches.

## Validation

- full `tools/check_architecture.py` gate passed
- `tests/unit/tooling/test_architecture_process_policies.py` passed
- `tests/unit/tooling/test_architecture_policy.py` passed
- Ruff lint and format checks passed for `tools/check_architecture.py`
- `git diff --check` passed

### DIFF
--- a/tools/check_architecture.py
+++ b/tools/check_architecture.py
@@ -198,6 +198,10 @@ _SUBPROCESS_ALLOWED_EXACT: frozenset[PurePosixPath] = frozenset(
             "benchmarks/validation/conjecture_probes_v1/"
             "test_yang_mills_gauge_invariance_certificate.py"
         ),
+        PurePosixPath(
+            "benchmarks/validation/conjecture_probes_v1/"
+            "test_zarankiewicz_projective_plane_certificate.py"
+        ),
         # Repository-command integration tests deliberately invoke CLI entrypoints.
         PurePosixPath("benchmarks/validation/test_benchmark_plan_validation.py"),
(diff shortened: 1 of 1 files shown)


yuelgrace1810-ops:
Closing as redundant after PR #797 merged. Current `main` now contains the exact same Zarankiewicz subprocess allowlist change. The #897 patch and the corresponding branch-local commit have identical stable patch IDs, and `tools/check_architecture.py` has no remaining diff between #897 and current main. No merge is needed.

PR #898, opened 2026-08-08
fix(mcp): repair object-rooted math.find MCP output schema

yuelgrace1810-ops:
## Summary

- publish `math.find`'s discriminated structured result as an object-rooted MCP output schema
- preserve the existing `oneOf` and discriminator contract
- add a latest-protocol `tools/list` serialization regression test

## Why

A corrected recently-solved-conjecture treatment run reached Jacobian over HTTP but failed before tool use. The live MCP client rejected `tools/list` because `math.find.outputSchema` lacked the required top-level `type`. Pydantic's `RootModel` emitted the union definitions and discriminator without the shared object root.

The currently negotiated older protocol path can normalize this schema, which is why existing conformance coverage remained green. The regression test exercises the latest `2026-07-28` serializer boundary that exposed the incompatibility.

## Safety

This does not change tool behavior, routing, mathematical results, or isolation. It adds only the common object-root declaration already satisfied by every union member.

## Validation

- focused MCP SDK v2 conformance tests: 2 passed
- Ruff lint: passed
- Ruff format check: passed
- diff whitespace check: passed

### DIFF
--- a/src/jacobian/adapters/mcp/tools.py
+++ b/src/jacobian/adapters/mcp/tools.py
@@ -118,6 +118,11 @@ class CapabilityDiscoveryResponse(
 ):
     """Closed, discriminated structured output for math.find."""
 
+    # MCP tool output schemas describe structured content objects. Pydantic's
+    # RootModel preserves the discriminated union but omits the common object
+    # root from JSON Schema, which stricter MCP clients reject during tools/list.
+    model_config = ConfigDict(json_schema_extra={"type": "object"})
+
 
 CapabilityDiscoveryToolResult = Annotated[
     CallToolResult,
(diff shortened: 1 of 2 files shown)


yuelgrace1810-ops:
Cross-domain canary validation (finite graph theory, 2026 Randić/matching equality counterexample):

- On current `main` (`3684af60`), the latest MCP client reached the server but `tools/list` failed with `ListToolsResult tools.0.outputSchema.type: Field required`; the grounded agent received no tools and made 0 MCP calls.
- On this PR (`b9954980`), the identical frozen prompt completed 7 MCP calls (`math.find` ×3, `math.run` ×4).
- `graph.invariant.maximum_matching.compute` returned cardinalities 1 for `K_{1,9}` and 2 for `K_{2,8}`.
- `graph.invariant.maximum_matching.verify` independently accepted both Tutte–Berge certificates with `VERIFIED` assurance.
- The final answer correctly derived `R(K_{1,9})-α'=2` and `R(K_{2,8})-α'=2`, while preserving the finite-scope caveat.

This confirms the schema fix restores tool discovery and verified execution beyond the original conjecture domain.

PR #783, opened 2026-08-07
Preserve typed domain validation diagnostics

yuelgrace1810-ops:
Closes #781.

## What changed

- enrich fail-closed domain-operation validation failures with a bounded JSON-pointer path, validation reason, error count, and validation type;
- never include the rejected input value in returned diagnostics;
- identify stale topology content bindings as `/complex/complex_digest` while retaining the existing topology error code and assurance boundary;
- add a regression proving the invalid integral-homology request produces no artifacts and exposes the recoverable field.

## Root cause

The capability-dispatch JSON Schema boundary preserved precise schema constraints, but later Pydantic/domain-model failures were replaced with the static domain diagnostic. The topology contract knew that the digest did not bind the canonical complex, but the agent only saw generic topology guidance.

## Safety

This does not accept any previously invalid input, change mathematical results, or weaken assurance. It only adds bounded diagnostic metadata after validation has already failed.

## Validation

- focused topology composition suite: 20 passed
- Ruff lint and format checks passed
- `git diff --check` passed

PR #923, opened 2026-08-09
Bound exact verifier validation diagnostics

yuelgrace1810-ops:
## What changed

- bound exact-domain validation error messages to the `CapabilityDiagnostic.message` contract limit
- apply the same protection to inline candidate validation and stored-result resolution
- add a regression for a malformed Jacobian-syzygy candidate that produces many validation errors

## Why

A weak-model canary copied a valid syzygy candidate with malformed zero-term padding. The verifier should reject that candidate with `INVALID_EXACT_DOMAIN_INPUT`, but the full Pydantic validation text exceeded the diagnostic field's 1024-character limit. Constructing the diagnostic then raised a second validation error, which was surfaced as the misleading `ADAPTER_EXECUTION_FAILED` response.

The bounded message preserves the specific fail-closed validation diagnostic and appends an explicit truncation marker.

## Validation

- `python -m pytest tests/composition/runtime/test_frontier_capabilities.py -k 'graded_jacobian_syzygy or oversized_candidate_validation or syzygy_checker_rejects' -q` — 6 passed
- `python -m ruff check src/jacobian/exact_domain_checkers.py tests/composition/runtime/test_frontier_capabilities.py`
- `git diff --check`


morluto:
Second-pass architecture review: **rebase + likely merge**. This is a focused fail-closed adapter fix: malformed producer/candidate validation must not overflow the public diagnostic contract and degrade into a misleading adapter failure. It does not alter mathematical semantics or checker authority, and it complements rather than conflicts with #783's richer typed validation diagnostics.

morluto (commented):

**Disposition: supersede this branch; port the regression into #1242 or a fresh tiny PR.**

The bug is valid: a second validation failure must never replace the original exact-checker diagnostic. But the current branch spans hundreds of pre-cutover files, including deleted workflow, Skill, relationship, assurance, and test infrastructure. It is not safe to rebase mechanically.

The surviving change should be one shared bounded/redacted validation projection used by current checker candidate validation and stored-result resolution, with a current `CapabilityResult` regression. #1242 is already the natural owner for that policy. Do not retain the branch history or old frontier-composition test path.

PR #924, opened 2026-08-09
fix(polynomials): handle zero Jacobian partial derivatives

yuelgrace1810-ops:
## What changed

- ignore explicit zero coefficients emitted by SymPy while assembling graded Jacobian coefficient maps
- add an end-to-end compute-and-verify regression for `h=x^2 z`, whose `y` partial derivative is zero

## Why

For a valid homogeneous polynomial that omits one variable, SymPy represents the corresponding zero partial with a zero term at exponent `(0,0,0)`. The coefficient-map builder attempted to place that term in the positive-degree target basis and raised `KeyError`, surfacing as `ADAPTER_EXECUTION_FAILED`.

Skipping a zero coefficient preserves the exact linear map and lets the producer return the expected degree-0 kernel. The independent verifier accepts the resulting candidate.

## Validation

- `python -m pytest tests/composition/runtime/test_frontier_capabilities.py -k 'graded_jacobian_syzygy or syzygy_checker_rejects' -q` — 6 passed
- `python -m ruff check src/jacobian/domains/polynomial/jacobian_syzygy.py tests/composition/runtime/test_frontier_capabilities.py`
- `git diff --check`

### DIFF
--- a/src/jacobian/domains/polynomial/jacobian_syzygy.py
+++ b/src/jacobian/domains/polynomial/jacobian_syzygy.py
@@ -127,6 +127,8 @@ def _coefficient_matrix(
         for basis_index, multiplier_exponents in enumerate(source_basis):
             column = component * len(source_basis) + basis_index
             for partial_exponents, coefficient in partial.terms():
+                if coefficient == 0:
+                    continue
                 target_exponents = cast(
                     tuple[int, int, int],
                     tuple(
--- a/tests/composition/runtime/test_frontier_capabilities.py
+++ b/tests/composition/runtime/test_frontier_capabilities.py
@@ -289,6 +289,44 @@ def test_graded_jacobian_syzygy_finds_and_verifies_the_first_kernel(
     assert verified.output["status"] == "VERIFIED"
 
 
+def test_graded_jacobian_syzygy_handles_a_zero_partial_derivative(
+    frontier_services: DomainTestServices,
+) -> None:
+    input_payload = {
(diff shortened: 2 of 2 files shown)


yuelgrace1810-ops:
Independent held-out reproduction on current main (`614fdc40511d41bf88607cb0c6cc7c1e6cb91667`):

- Source artifact: the exact homogenized curve coordinate `H(x,y,z)=-x^3+x^2*z` from Gao's 2026 tangent-sweep account of the Jacobian-conjecture counterexample.
- Request: `polynomial.jacobian_syzygy.minimum_degree.compute`, `max_degree=0`, `coefficient_map_detail=CERTIFICATES`.
- Two identical executions failed with `ADAPTER_EXECUTION_FAILED`.
- Server traceback reaches `_coefficient_matrix` and ends with `KeyError: (0, 0, 0)` while processing `H_y=0`.
- Independent oracle: rows/cols/rank/nullity `(6,3,2,1)`, witness `(0,1,0)`.

I replayed this exact polynomial against the patch in #924: producer returned the expected degree-0 kernel and the independent verifier returned `VERIFIED`.

Additional local validation on the PR-equivalent fix:
- 869 unit passed
- 185 domain passed
- 759 component passed, 3 platform skips
- focused old + new syzygy producer/verifier regressions passed
- Ruff lint/format and mypy passed

This confirms #924 covers another non-monomial polynomial with a zero partial, not only the existing `x^2 z` regression.

PR #927, opened 2026-08-09
Verify materialized Jacobian syzygy ledgers

yuelgrace1810-ops:
## Summary

- add an independent verifier for `polynomial.jacobian_syzygy.coefficients.materialize`
- bind verification directly to the materialized producer's stored `result_uri`
- advertise the correct verifier from sparse-ledger results and replay the materializer's forced sparse-detail semantics
- preserve the existing compact syzygy producer/verifier path

## Reproduction

A weak-model audit of Satriano–Usatine's 2026 Batyrev counterexample used the exact degree-14 homogenization of their published stringy E-polynomial through multiplier degree 8. The model selected the evidence-complete coefficient-ledger producer, which completed and returned a stored result, but could not verify it: discovery exposed no related ledger verifier, while the advertised minimum-degree verifier requires an inline `{input, candidate}` envelope. Large stored ledgers have opaque previews, so the candidate cannot be recovered inline.

This PR gives the materialized producer its own provenance-bound verifier accepting `{result_uri}`.

## Tests

- `ruff check` on all changed files
- focused materialized-ledger end-to-end verification regression
- existing sparse-ledger and compact nine-line producer/verifier tests
- complete frontier capability, checker installation, syzygy invariant, and polynomial bundle test files

All passed locally.

### DIFF
--- a/src/jacobian/domains/topology/operations.py
+++ b/src/jacobian/domains/topology/operations.py
@@ -505,6 +505,21 @@ _CIRCLE = {
     "facets": [["a", "b"], ["b", "c"], ["a", "c"]],
 }
 
+_CANONICAL_CIRCLE = {
+    "vertices": ["a", "b", "c"],
+    "maximal_simplices": [["a", "b"], ["a", "c"], ["b", "c"]],
+    "faces_by_dimension": [
+        {"dimension": 0, "faces": [["a"], ["b"], ["c"]]},
+        {"dimension": 1, "faces": [["a", "b"], ["a", "c"], ["b", "c"]]},
+    ],
+    "dimension": 1,
+    "f_vector": [3, 3],
+    "closure_size": 6,
+    "complex_digest": (
+        "sha256:6f797991bac967e2a8e572707df487061655df0f094cbde0f52f82c5401fc043"
+    ),
+}
+
 type TopologyOperation = (
     InstalledOperation[
         SimplicialComplexRequest, SimplicialComplexCanonicalizationResult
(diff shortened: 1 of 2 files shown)


morluto:
Second-pass architecture review: **rebase + likely merge**. The review cycle appears to have fixed the important trust-boundary issues: canonical >4,300-digit conversion, verifier discoverability, checker cost gating, linear-factor support, and digest formatting. The final shape is appropriate: a materialized evidence-producing capability gets its own provenance-bound verifier, while the compact inline producer/verifier path remains separate.

morluto (commented):

**Disposition: retain the need, redesign the verifier on current `main`.**

Large materialized ledgers are a legitimate durable-artifact case that typed runtime-local value references do not replace. However, the checker should not be framed as trusting producer provenance or a legacy `result_uri`. An artifact URI is a carrier, not proof.

Define a current checker request that binds the expected typed syzygy subject/input and the exact artifact digest/content, independently parses and replays the ledger, and emits a verification record bound to that subject, candidate/evidence, operation, and checker. Remove catalog relationships and “advertise from producer” logic. Use `artifact_uri`/current storage contracts, preserve the compact inline checker separately, and rebase all tests onto current result/verification paths.

morluto:
Post-#1256/#1265 review: **keep the materialized syzygy verifier, but split out and delete the old discovery-relationship mechanism.**

The missing checker is still a coherent feature: a retained coefficient ledger can have an independent exact verifier with its own operation ID. That fits the current verification model.

The `adapters/mcp/projections.py` hunk does not: it extends the old `_RELATED_CAPABILITIES` map with producer↔verifier navigation. #1256 deliberately removed relationship graphs and current discovery is factual lexical search/exact inspection, not a workflow navigator.

Recommendation: restack from current main with the checker declaration/implementation/tests only. Make it discoverable through its own title/tags and a valid invocation example if necessary. Drop the projections change entirely, then rebase the frontier/catalog test onto the current explicit portfolio.

morluto:
Superseded by #1292. The replacement keeps the durable `result_uri` verifier, clean-process replay, canonical >4,300-digit integer handling, and an explicit replay-complexity gate. It intentionally deletes the `_RELATED_CAPABILITIES` projection and all workflow relationship tests.

PR #917, opened 2026-08-09
fix(checkers): repair syzygy verifier handoff metadata

yuelgrace1810-ops:
## What changed

- correct the graded Jacobian-syzygy result metadata to say that verification consumes the exact producer `input` and inline `candidate`, rather than a nonexistent `result_uri`;
- update the independent verifier description to state the same contract;
- add a regression covering both the producer hint and the installed verifier description.

## Root cause and observed impact

A held-out evaluation of Dimca and Pokora's 2026 nine-line counterexample (arXiv:2607.01985) successfully discovered and ran `polynomial.jacobian_syzygy.minimum_degree.compute` for both arrangements. The model then inspected the independent verifier, but declined to call it because:

- the computed result advertised `verification_input_field: "result_uri"`;
- the compute operation produced no result artifact URI because this is an ordinary inline computed operation; and
- the verifier description said it consumed a "stored producer result".

The installed verifier actually accepts the original producer `input` and its inline `candidate`. The conflicting metadata converted two available independent replay paths into COMPUTED-only final claims.

## Scope and safety

This changes only agent-facing handoff metadata and the result's self-description. It does not alter the polynomial computation, candidate schema, independent checker implementation, authorization, assurance semantics, or verification record behavior. The checker still reconstructs every bounded coefficient map independently and remains fail-closed.

## Validation

Focused regression:

- graded Jacobian-syzygy compute/verify test: 1 passed.

Planner-selected validation:

- unit: 868 passed;
- component: 759 passed, 3 expected platform skips;
- domain: 185 passed;
- composition: 475 passed, 2 expected Lean skips;
- storage: 125 passed;
- process: 234 passed, 1 expected platform skip;
- MCP: 46 passed;
- end-to-end: 8 passed, 1 expected Lean skip;
- Ruff, formatting, complexity, dependency, dead-code, mypy, test architecture, product architecture, package build: passed.

## Overlap

Issue #883 tracks the wider audit of historical artifact-backed verification boundaries. This PR does not perform that audit or alter transport architecture; it fixes one directly observed contradictory inline handoff.

PR #929, opened 2026-08-09
Clarify inline verifier invocation envelopes

yuelgrace1810-ops:
## What changed

- Advertise a concrete `math.run` invocation envelope for inline replay verifiers in both discovery results and full contract inspection.
- State explicitly that `input` and `candidate` are sibling fields inside `payload`.
- Correct the graded Jacobian-syzygy verifier description, which previously claimed it consumed a stored producer result even though its contract accepts an inline producer input/result pair.
- Cover the structured and model-visible text projections with MCP boundary tests.

## Why

A held-out weak-model evaluation using the exact degree-8 binary partition-polynomial log-concavity counterexample found the correct Jacobian syzygy immediately, but then spent 408 seconds and failed two verifier calls. The model first nested `candidate` inside `input`, then placed it beside `payload`. Both invalid shapes are plausible readings of a contract that had no invocation example and contradictory stored-versus-inline wording.

This change exposes transport shape without prescribing mathematical strategy, making producer-to-verifier replay easier and reducing avoidable validation retries.

## Validation

- `make test-unit` — 869 passed
- `make test-component` — 759 passed, 3 skipped
- `make test-domain` — 185 passed
- `make test-mcp` — 46 passed
- `make test-e2e` — 8 passed, 1 skipped
- `make check-static` — passed, including lint, format, dependency, type, architecture, and build checks
- Focused Jacobian syzygy producer/verifier test — passed

`make npm-test` could not run because `npm` is not installed on this host.


yuelgrace1810-ops:
Additional held-out reproduction on current main (2026-08-09): a `gpt-5.4-mini`/low run audited the exact Dimca–Pokora nine-line Generalized Terao counterexample (arXiv:2607.01985). It successfully discovered, executed, and independently verified both `geometry.projective_line_arrangement.flats` artifacts. For `polynomial.jacobian_syzygy.minimum_degree`, it computed the correct `mdr(f)=4` and `mdr(g)=5`, but made four invalid verifier attempts: bare `result_uri` twice, then `{input, candidate:{result_uri}}`, then `{input, candidate:<materializer wrapper>}`. It stopped at `COMPUTED` assurance despite the verifier contract being inspected first. Valid grounded runtime was 185.84s / 20 MCP calls / 413,125 input tokens, versus control 112.40s / 171,792 input tokens. This is the same invocation-envelope ambiguity addressed by this PR, not a separate defect.

yuelgrace1810-ops:
Post-#924 recovery evidence (2026-08-09): the same weak-model configuration (`gpt-5.4-mini`, low) successfully replayed the frozen zero-partial Gao coordinate task on current main. It found the producer/verifier pair, produced the exact degree-0 certificate for `H=-x^3+x^2z`, and passed the producer's clean inline input/result to the verifier on the first attempt; the verifier returned `VERIFIED`. End-to-end: 61.50s, 5 MCP calls, 137,364 input tokens. This does not isolate the effect of this PR because the branch was not applied, but it confirms that a clearly understood sibling-field envelope is sufficient and provides a clean comparison point for the earlier invalid shapes.

yuelgrace1810-ops:
Branch validation against the original frozen `gpt-5.4-mini`/low partition-polynomial case exposed an important dependency boundary. The #929-only branch discovered the producer/verifier and produced the correct degree-0 result, but current main's producer still advertised `verification_input_field: "result_uri"`; the model then made an invalid verifier call and the 600 s arm timed out. This does not contradict the envelope fix here—the stale producer hint is the separate defect already fixed by draft PR #917. I refreshed #917 onto current main and validated its focused syzygy tests (2 passed) plus the full MCP lane (47 passed). A combined post-fix model replay should wait until #917 is present alongside this branch.

morluto:
Second-pass architecture review: **rebase + likely merge**. This survives the post-#956 architecture well: it exposes the concrete transport shape of an already-declared producer/verifier relationship without prescribing verification timing or research strategy. The earlier reasoning-ID concern is obsolete because #956 removed the production reasoning-mode schema. The later review hardening around typed invocation-protocol models, explicit producer relationships, and `response.output.result` targeting is the right direction.

morluto (commented):

**Disposition: revise substantially; retain the invocation-example goal.**

The observed payload-shape failure is valid, but this implementation derives a new MCP `invocation_protocol` projection from a deleted producer relationship and schema heuristics. That recreates transport-specific metadata and another source of truth.

On current `main`, put a concrete `{capability_id, payload: {input, candidate}}` invocation example directly on each applicable checker `OperationSpec`/descriptor. Exact inspection already returns the full descriptor, so no synthetic projection or relationship test is needed. Correct the syzygy description at the checker declaration, add one generic invariant that advertised examples validate against `math.run`, and delete all relationship/projection code from this branch.

morluto:
Architecture review after #1256/#1265: **superseded as an implementation; keep the usability problem and solve it through current descriptor examples.**

The observed failure is valid: inline replay checkers with sibling `input` / `candidate` fields are easy for agents to wrap incorrectly. But this PR derives a special `invocation_protocol` from `descriptor.related_capabilities` / `CapabilityCatalogRelationshipKind`. Those relationship models were deliberately removed in #1256 along with workflow-like catalog navigation.

Current exact inspection already has authoritative request schemas and bounded, operator-authored `invocation_examples`. That is the simpler owner for transport-shape guidance: give affected checker declarations one schema-valid example showing `{capability_id, payload: {input, candidate}}`, and let exact inspection expose it normally. No relationship graph or verifier-specific MCP projection is needed.

Recommendation: do not rebase this patch. Replace it with a much smaller current-main change adding validated invocation examples to the affected checker descriptors (and an MCP inspection regression), then close #929 as superseded.

morluto:
### Superseded by typed ports, operation examples, and current lineage cleanup

Closing this pre-cutover implementation rather than rebasing its generic projection heuristic.

The reproduced failure was real, but the surviving solution is now split across current ownership boundaries:

- #1234 owns correction of inline-vs-stored lineage wording;
- #1262 carries whole typed producer results to compatible checkers through declared value ports, avoiding candidate retranscription;
- current operation inspection owns authoritative schemas and operation-owned invocation examples;
- #1267 strengthens guidance to copy inspected examples instead of probing wrappers.

This branch inferred an `invocation_protocol` from catalog relationships and request-field names. That recreates transport policy in the discovery projector and depends on relationship machinery deliberately removed by #1256/#1265. New verifier examples or ports should be declared by the operation/checker owner, not synthesized by MCP projection.

**Disposition: root cause retained; implementation superseded.**

Issue #900, opened 2026-08-09
[Feature]: Add bounded exact verification for variable-coefficient recurrences


yuelgrace1810-ops:
## Area

Exact combinatorics; bounded recurrence evaluation and verification

## Observed reliability gap

A held-out audit based on Tong Niu's 2026 proof of Mathar's OEIS A176677 recurrence exposed a gap between Jacobian's nearby contracts:

- `combinatorics.recurrence.linear.evaluate` / `.verify` accept constant coefficients only;
- `combinatorics.generating_function.coefficients.compute` / `.verify` accept rational generating functions only.

The audited sequence is defined by the nonlinear convolution

```
a(0) = a(1) = 1
a(n+1) = sum_{p=0}^n a(p)a(n-p) - 1
```

and the proved relation is the variable-coefficient P-recurrence

```
(n+1)a(n)
+ 2(-3n+1)a(n-1)
+ (9n-13)a(n-2)
- 4a(n-3)
+ 4(-n+4)a(n-4) = 0.
```

Source: https://arxiv.org/abs/2605.04369

## Reproduction evidence

A paired run used `gpt-5.4-mini` with low reasoning on the exact finite scope `0 <= n <= 16`.

- Control: 90.35 s, 0 MCP calls. It generated the correct table, used a separate exact calculation, and obtained zero residuals for every `4 <= n <= 16`.
- Jacobian-grounded: 153.98 s, 2 `math.find` calls. It correctly discovered that the nearby constant-coefficient and rational-GF capabilities were inapplicable and did not force either contract. It then performed unsupported hand arithmetic, returned `a(16)=3,950,089` instead of `3,949,969`, and falsely reported residual `2040` instead of `0`.

The correct tail is:

```
a(13)=115940
a(14)=373709
a(15)=1211740
a(16)=3949969
```

All residuals through `n=16` are exactly zero.

This is not a defect in the existing constant-coefficient capabilities: discovery and refusal behavior were correct. It is a missing exact fallback that becomes visible with weaker agents.

## Proposed capability

Add a narrow bounded exact capability pair for variable-coefficient linear recurrences, for example:

- `combinatorics.recurrence.p_recursive.evaluate`
- `combinatorics.recurrence.p_recursive.verify`

Suggested input contract:

- exact initial integer/rational terms;
- coefficient polynomials `p_j(n)` over `ZZ` or `QQ`;
- declared index range;
- optional candidate term table for verification;
[...]

PR #934, opened 2026-08-09
feat(math): expose P-recursive evaluation under sequence domains

yuelgrace1810-ops:
## What changed

- Tag the exact polynomial-coefficient recurrence evaluator with the intuitive `sequence` and `polynomial` domain aliases.
- Add discovery regressions proving that variable-coefficient recurrence queries under either domain route to `combinatorics.recurrence.p_recursive.evaluate` first.

## Why

A frozen weak-model replay of Tong Niu's 2026 proof of Mathar's OEIS A176677 recurrence searched Jacobian twice using natural domain filters:

1. `domain="polynomial"` with a variable-coefficient recurrence query;
2. `domain="sequence"` with an exact finite recurrence-residual query.

Both filters were recognized, but both excluded the newly merged P-recursive evaluator because its descriptor only advertised the `combinatorics` domain. The model therefore fell back to hand arithmetic even though the exact operation was installed. This is a routing/discoverability defect, not a computation defect.

The capability remains owned by the combinatorics bundle; these tags only expose two mathematically accurate entry vocabularies already used by nearby installed operations and agents.

## Validation

- `uv run --locked pytest -q tests/composition/runtime/test_recurrence_series_capabilities.py` — 19 passed
- `make test-mcp` — 47 passed

## Scope

No recurrence arithmetic, schemas, assurance, checker behavior, or capability identifiers change.

### DIFF
--- a/src/jacobian/domains/combinatorics/recurrence.py
+++ b/src/jacobian/domains/combinatorics/recurrence.py
@@ -137,6 +137,8 @@ RECURRENCE_CAPABILITIES = (
         evaluate_polynomial_coefficient_recurrence,
         "combinatorics",
         "recurrence",
+        "sequence",
+        "polynomial",
         "p-recursive",
         "polynomial-coefficients",
         "exact-rational",
--- a/tests/domain/combinatorics/test_recurrence_series_capabilities.py
+++ b/tests/domain/combinatorics/test_recurrence_series_capabilities.py
@@ -184,6 +184,27 @@ def test_polynomial_coefficient_recurrence_exposes_exact_terms_and_residuals(
     }
 
 
+@pytest.mark.parametrize("domain", ("sequence", "polynomial"))
+def test_polynomial_coefficient_recurrence_is_discoverable_by_intuitive_domain(
+    combinatorics_services,
+    domain: str,
+) -> None:
+    discovered = combinatorics_services.core.capabilities.discover(
+        CapabilityDiscoveryRequest(
+            query=(
+                "exactly compute terms of a sequence and residuals of a "
+                "variable-coefficient polynomial recurrence"
+            ),
+            domain=domain,
+            limit=5,
+        )
+    )
+
+    assert "combinatorics.recurrence.p_recursive.evaluate" in {
+        match.capability_id for match in discovered.matches
+    }
+
+
 def test_polynomial_coefficient_recurrence_rejects_singular_required_step(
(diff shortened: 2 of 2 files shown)


yuelgrace1810-ops:
Post-PR weak-model replay confirms the routing change: with this branch, `combinatorics.recurrence.p_recursive.evaluate` became the first result for the same intuitive recurrence search (previously absent under both `sequence` and `polynomial`). The model did not invoke it because the frozen Mathar task requires replay of a separately convolution-derived candidate table, while the current evaluator generates its own table from the P-recurrence. That remaining contract gap is tracked by reopened issue #900. Accordingly, this PR should be reviewed only as a general domain-alias discoverability fix; it does not claim to complete #900 or make the Mathar audit tool-verifiable.

morluto:
Second-pass architecture review: **likely merge after rebase**. Adding `sequence` and `polynomial` as mathematically accurate discovery aliases is consistent with agent-owned search/composition and does not prescribe a workflow. The PR also correctly keeps the remaining candidate-table replay gap separate rather than claiming this discoverability change solves #900.

morluto (commented):

`sequence` and `polynomial` are useful factual search vocabulary for a P-recursive evaluator without changing its combinatorics ownership. This is consistent with current lexical discovery and does not require domain aliases as a separate registry concept.

Rebase onto current `main`, attach the terms to the surviving operation declaration, and update tests to current `math.find` behavior. Avoid freezing “ranked first” when presence/strong lexical applicability is the real invariant, and remove old assurance wording.

morluto:
Post-simplification review: **keep and rebase.** Adding `sequence` / `polynomial` vocabulary to the existing exact recurrence operation is aligned with the final `math.find` model: bounded lexical retrieval over factual operation metadata.

This does not create another capability, planner edge, or result abstraction; it just makes an existing atomic tool discoverable under language agents actually use. Preserve the operation as one typed computation and keep the focused discovery regression on the current catalog/search surface.

Recommendation: rebase and merge as a small descriptor-quality fix.

morluto (commented):

This PR was refreshed after the typed-tool simplification and has landed as `3ff985b`.

The final change is limited to factual discovery aliases and their regressions: the combinatorics-owned P-recursive evaluator is discoverable through `sequence` and `polynomial` vocabulary without changing capability ownership, arithmetic, schemas, verification behavior, or the fixed MCP surface.

PR #935, opened 2026-08-09
Fix MCP tool discovery over streamable HTTP

yuelgrace1810-ops:
## Summary

- make root-union structured output schemas valid MCP 2 object schemas on the HTTP boundary
- clarify `math.find` input-kind filtering so intent searches omit the filter and exact JSON claims use `STRUCTURED_REQUEST`
- cover both contracts with MCP boundary tests

## Reproduction

A clean `uv sync --frozen` environment on current main starts successfully, but a streamable-HTTP client cannot complete `tools/list`:

```
ValidationError: tools.0.outputSchema.type Field required
```

`CapabilityDiscoveryResponse` is a Pydantic `RootModel` union, whose generated schema has a root `oneOf` but no explicit top-level `type`. MCP 2's wire validator requires that type, so clients receive no Jacobian tools.

This was found by a frozen weak-model audit of the proof-critical rational identity in the 2026 resolution of Conjecture 1 of Kulenović, Ladas and Overdeep (2003). Before the fix the model saw zero tools. After the fix it discovered `polynomial.rational_function.identity.verify`, invoked it, and reported a `TRUE` result with `VERIFIED` assurance, `COMPLETE` coverage, the certificate and verification-record URIs, and the fraction-field/pointwise-definedness scope boundary.

## Validation

- `.venv/bin/pytest tests/boundary/mcp -q` — 19 passed
- frozen `gpt-5.4-mini`, low-reasoning audit — 6 MCP calls, valid verified result in 81.72 s

The adapter normalization only adds `type: object` when an SDK-generated output schema lacks a top-level type; existing ordinary model schemas are unchanged.


yuelgrace1810-ops:
Post-refresh validation against current `main` (`6aea0129`): merged cleanly, refreshed the locked environment, and reran the full MCP boundary suite: **17 passed**. Head is now `003e7203`.

Frozen canary evidence remains valid for the functional change: before the output-schema fix, streamable-HTTP `tools/list` failed with `tools.0.outputSchema.type Field required` and the weak model received zero MCP tools. With the fix, the same case discovered `polynomial.rational_function.identity.verify`, recovered from sparse-term ordering, and obtained a `TRUE` / `VERIFIED` / `COMPLETE` result with certificate and verification-record URIs.

morluto:
Second-pass review: the **underlying bug is still important, but this should be restacked/reimplemented on current post-#956 MCP SDK 2.0 architecture rather than mechanically rebased**. Current `JacobianMCPServer.list_tools()` hardens `input_schema` but still does not normalize a root-union `output_schema`, so the HTTP `tools/list` failure remains relevant. Please transplant the minimal output-schema normalization and its boundary regression onto current `main`, avoiding old pre-#956 adapter seams.

morluto:
Tracking issue opened: #1017 — normalize root-union MCP output schemas on the current SDK 2.0 boundary.

yuelgrace1810-ops:
Addressed in 0435d826: the branch is restacked around the current MCP SDK 2.0 server boundary. It retains the minimal root-union `output_schema` object normalization and removes the unrelated legacy discovery-description changes. MCP SDK 2 conformance tests pass (2 passed).

yuelgrace1810-ops:
Follow-up: current `main` is now merged in 95ef47a2. The schema fix now extends the current public SDK-facing `list_tools()` compatibility seam, normalizing root-union output schemas alongside strict input schemas. GitHub now reports the PR mergeable; SDK conformance tests pass (2 tests).

morluto (commented):

**Disposition: close as already absorbed by current `main`.**

The current MCP tool construction already normalizes root-union output schemas to an MCP-compatible top-level object schema, so the streamable-HTTP `tools/list` failure addressed here no longer reproduces on the maintained path. The accompanying input-kind/view guidance also targets discovery projections removed by #1256.

Retain a streamable-HTTP boundary regression if one is not already present, but do not rebase this branch or restore the old projection code.

PR #943, opened 2026-08-09
fix(polynomials): clarify canonical Gaussian polynomial term order

yuelgrace1810-ops:
## Summary
- document that sparse Gaussian polynomial terms must be ordered lexicographically by the complete exponent vector
- make validation errors identify the first offending adjacent exponent pair
- add a regression test covering the recovery guidance

## Motivation
A weak-model audit based on Long's resolved Gaussian Moments conjecture constructed a mathematically valid sparse polynomial but ordered its terms by an intuitive variable-first convention. The schema only said “canonical exponent order,” and the resulting error did not explain the required ordering well enough to recover efficiently.

This change improves discoverability and recovery only; it does not alter mathematical semantics or resource bounds.

## Validation
- `ruff check src/jacobian/contracts/probability.py tests/domain/probability/test_finite_probability.py`
- `pytest -q tests/domain/probability/test_finite_probability.py` (21 passed)

### DIFF
--- a/src/jacobian/contracts/probability.py
+++ b/src/jacobian/contracts/probability.py
@@ -127,6 +127,10 @@ class GaussianPolynomial(ContractModel):
     terms: tuple[GaussianPolynomialTerm, ...] = Field(
         min_length=1,
         max_length=MAX_GAUSSIAN_POLYNOMIAL_TERMS,
+        description=(
+            "Nonzero sparse terms ordered lexicographically by their complete "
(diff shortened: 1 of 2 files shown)


morluto:
Second-pass architecture review: **likely merge after rebase**. This is contract clarity rather than workflow policy: it makes an existing canonical sparse-term ordering explicit and improves recovery by naming the first offending adjacent pair. No mathematical semantics, assurance, or capability composition behavior changes.

PR #981, opened 2026-08-09
Add exact finite-table mutual information certificates

### LINKED ISSUE #978: [Capability]: Add exact finite-table mutual-information certificates
## Capability gap

Jacobian supports several exact finite-probability primitives but lacks a checker-backed operation for Shannon mutual information / KL divergence of explicit rational tables with exact symbolic logarithmic certificates.

## Held-out reproduction

Wang-Wang-Chen's arXiv:2608.03828v2 publishes two qutrit joint tables:

`PX=[[0,1/4,1/4],[1/4,0,0],[1/4,0,0]]`

`PZ=[[1/6,1/6,1/12],[1/6,1/6,1/12],[1/12,1/12,0]]`.

Exact reconstruction gives:

- PX marginals `(1/2,1/4,1/4)`; every positive-cell likelihood ratio is 2; with `L=4`, the product certificate is 16, hence `I_X=1`.
- PZ marginals `(5/12,5/12,1/6)`; ratios are four copies of `24/25` with mass `1/6` and four copies of `6/5` with mass `1/12`; with `L=12`, the product is `(3456/3125)^4`, hence `I_Z=(1/3)log2(3456/3125)>0`.
- Therefore `I_X+I_Z>1` by exact integer comparison `3456>3125`.

A grounded weak-model run made three discovery calls and zero executions; no compatible mutual-information contract surfaced. It computed correctly locally but more than tripled input context and misstated its call count. Routing is tracked on #845.

## Proposed bounded contract

Consider `probability.joint.mutual_information.compute/verify` and/or `probability.relative_entropy.compute/verify` for finite rational tables. For exact rational inputs, return:

- normalization and exact marginals;
- every positive-support ratio, with zero-support conventions explicit;
- a canonical symbolic result as a rational linear combination of logs or an integer-power product certificate `L*I=log_b(R)`;
- sign/equality comparisons proven by exact rational/integer comparison where possible;
- independent checker record bound to the exact table, support, log base, and normalization policy.

## Acceptance criteria

1. Reproduce both frozen tables, marginals, ratios, and product identities.
2. Return `I_X=1` exactly and the canonical qutrit `I_Z` expression.
3. Prove positivity without floating-point logarithms.
4. Handle zero cells without evaluating `0 log 0` or dividing by zero; reject invalid marginal support.
5. Bind row/column ordering and log base explicitly.
6. Keep scope to supplied tables; do not infer quantum-state provenance or resource properties.

This needs maintainer review because it adds symbolic-log canonicalization and checker semantics.

yuelgrace1810-ops:
## Summary
- add `probability.joint.mutual_information.compute` for bounded normalized rational joint tables with explicit ordered row/column labels
- return exact marginals, every positive-support likelihood ratio, and a canonical certificate `L·I = log_b(R)` with rational `R`; simplify to an exact rational value only when `R` is an exact power of the declared base
- add `probability.joint.mutual_information.verify`, backed by independent standard-library `Fraction` replay
- omit zero-mass terms explicitly and never evaluate logarithms using floating point

## Assurance boundary
The producer remains `COMPUTED`. Only the operator-authorized independent checker can return `VERIFIED`. The operation certifies only the supplied finite table; it does not infer quantum-state provenance or broader resource claims. Strict positivity is represented by exact comparison `R > 1`, with Gibbs nonnegativity as the stated mathematical basis.

## Validation
- focused finite probability, verifier, and qutrit reproduction tests: 34 passed
- reproduced both frozen #978 tables, including `I_X=1`, the `P_Z` product `(3456/3125)^4`, and exact positivity
- tampered likelihood ratio rejected by the independent checker
- mypy on affected source: passed
- `git diff --check`: passed
- hosted CI will run the broad lanes selected for the shared probability contract

Closes #978.


morluto:
Second-pass review: **request changes for an aggregate resource-bound hole**. The request bounds cells and each rational component, but certificate construction takes the LCM of all positive probability denominators and then materializes `ratio ** exponent` with exponents proportional to that LCM. A valid small table with many distinct large denominator factors can therefore create astronomically large exponents/products despite satisfying the per-cell bounds. Please add a pre-exponentiation aggregate scale/replay/output-cost bound, or keep the logarithmic product certificate factorized instead of materializing the huge rational product.

morluto:
Tracking issue opened: #1013 — bound aggregate growth in exact mutual-information certificates.

yuelgrace1810-ops:
Addressed in 4d943924: certificate construction now rejects excessive aggregate LCM scale and projected rational-power output cost before exponentiation. The independent checker mirrors both bounds. Focused mutual-information tests pass (4 passed), including pre-exponentiation rejection coverage.

morluto (commented):

**Disposition: retain, but simplify and rewrite around domain-owned values.**

The exact finite-table mutual-information certificate is useful, especially where the value is genuinely logarithmic rather than rational. Define a normalized labelled joint distribution as the authoritative value, with row/column identities structurally attached. The result should expose the exact marginals and the minimal canonical `L·I = log_b(R)` certificate/decision needed by callers.

Avoid publishing every derived likelihood ratio merely to make the checker convenient unless it is independently useful; the checker can recompute intermediates. Install a thin producer and separate standard-library replay checker, remove old probability bundle/generic assurance metadata, and keep exact zero-support omission, positivity, forged-certificate, and label-order regressions.

morluto:
Post-#1256/#1265 review: **keep the mutual-information capability, but rewrite it onto the current typed-operation architecture before merging.** The operation is still absent from current main, so this is not superseded mathematically.

The branch is written against the older `ComputedOperation` / `ComputedSuccess` / `relation_id` design and implements the mathematics directly inside `domains/probability/operations.py`. Those are exactly the competing representations #1256 removed.

Recommended restack: put the finite joint-table semantic value/function under the current probability math owner, keep Python-FLINT conversion/backend logic private, expose one canonical semantic input, then bind the boundary request/result with an `OperationSpec`. The sign/exact-value/log-product facts can remain in the **domain result**; they do not need a generic completeness/assurance wrapper. Keep an independent checker as a separate installed operation if the replay remains valuable.

Also retain the pre-computation bounds on scale/power cost, but enforce them through the authoritative request/preflight path before expensive exponentiation.

Recommendation: substantial revise/rebase, preserving the math and tests but deleting old operation/result/relationship plumbing.

morluto:
Superseded by #1300 after the math-layer and checker-runtime ownership rewrite. The original finite-table mutual-information certificate and replay here remain credited to @yuelgrace1810-ops; #1300 is a current-tree port, not a rejection or reassignment.

morluto:
Superseded by #1300, which carries the current-ownership port and explicitly credits @yuelgrace1810-ops for the original investigation, reproduction, and implementation work. Closing this older branch so there is one authoritative implementation.

Issue #947, opened 2026-08-09
Add canonical graph6 decoding and bound graph-invariant certificates


yuelgrace1810-ops:
A weak-model audit of Fischer’s 2026 zero-forcing counterexample exposed a severe input-decoding gap. The paper publishes H24 as graph6 `W{CGW_@?Y??@?@?@_@??@??K_????G??C??B??@????_??B`. No installed graph6 operation was found, and the grounded model used the wrong graph6 bit ordering, turning the connected subcubic source graph into a fictitious disconnected degree-8 graph and falsely declaring the paper invalid.

Correct decode: 24 vertices, edges `(0,1) (0,2) (1,2) (0,3) (3,4) (4,5) (4,6) (5,6) (3,7) (7,8) (7,9) (8,9) (1,10) (10,11) (11,12) (11,13) (12,13) (10,14) (14,15) (14,16) (15,16) (2,17) (17,18) (18,19) (18,20) (19,20) (17,21) (21,22) (21,23) (22,23)`. It is connected, has maximum degree 3, `alpha=9`, and zero-forcing number `Z=11`; exhaustive scans find no independent or zero-forcing 10-set.

Proposed capability split:

1. A small deterministic graph6 codec that follows the standard header and upper-triangle bit ordering, rejects sparse6/digraph6 unless explicitly supported, returns a canonical sorted edge list/degree sequence/digest, and has independent round-trip fixtures including H24.
2. Separately reviewed exact bounded producers/verifiers for maximum independent set and zero forcing, with witnesses, chronological forces, lower-bound certificates, and binding to the decoded edge-list digest.

The codec should not depend on optional nauty. PR #205 explored a nauty provider but explicitly deferred production capability registration. This reproduction shows a concrete reliability failure from that remaining gap.

Review is needed because exposing a graph format and choosing lower-bound certificates for NP-hard invariants affect contracts, limits, artifacts, and verifier independence.

yuelgrace1810-ops:
Replication from a distinct 2025 resolved-conjecture case: Clow–Haxell–Mohar, arXiv:2505.05339v2, use the 102-vertex Biggs–Smith graph to refute Lovász's hypergraph conjecture. An independent exact reconstruction gives 102 vertices, 153 edges, cubic, girth 9, diameter 7, alpha=43, and 12 canonical endpoint-distance classes across all 11,628 unordered edge pairs; exact MILP witnesses show alpha remains 43 after deleting the endpoints of a representative from every class.

[...]

PR #983, opened 2026-08-10
Add canonical graph6 decoding and verification

yuelgrace1810-ops:
## Summary
- add `graph.encoding.graph6.decode.compute` for standard small-order graph6 (orders 0–62), with optional `>>graph6<<` header
- implement the standard column-major upper-triangle bit order and return canonical sorted edges, degree sequence, and a digest of the exact decoded graph
- reject sparse6, digraph6, extended-order headers, invalid lengths, invalid characters, and nonzero padding
- add `graph.encoding.graph6.decode.verify` with independent standard-library bitstream replay

## Reproduction
The H24 string from #947 decodes to 24 vertices, 30 edges, and maximum degree 3. The regression fixture checks the first canonical edges and independent verifier, and a wrong-edge candidate is rejected.

## Scope boundary
This PR implements only the deterministic codec slice of #947. It does not add or broaden maximum-independent-set or zero-forcing optimization contracts. Producer output remains `COMPUTED`; only the authorized independent replay is `VERIFIED`.

## Validation
- focused graph producer/verifier and checker-installation suite: 25 passed
- mypy on affected source: passed
- `git diff --check`: passed
- hosted CI will run broad lanes selected for shared graph contracts/checkers

Related to #947; does not close it.


morluto:
Second-pass architecture review: **rebase + likely merge**. This is a clean atomic deterministic codec capability with explicit format rejection and an independent replay. It does not bundle the motivating NP-hard invariant work into the codec and stays within the domain-owned capability model. Existing issue #947 already tracks the broader graph6/invariant follow-up, so I would not open a duplicate issue for this PR.

morluto (commented):

**Disposition: retain the mathematical slice, rewrite around `jacobian.math`.**

A bounded exact graph6 decoder is a useful atomic codec operation and is not superseded by the cutover. Move decoding/canonicalization into a domain-owned graph encoding module and return a strict graph value whose vertex order/labels and edges are structurally bound; do not make the digest a substitute for that identity.

Install the producer as a thin `OperationSpec` and the standard-library replay as a separate authorized checker. Remove old graph bundle/exact-domain dispatcher/result metadata, relationships, and generic assurance fields. Keep the rejection cases for sparse6/digraph6/extended headers/padding and the H24 regression.

morluto:
Post-#1265 review: **keep the graph6 decoding/checking slice, but rebase it as a domain-owned graph conversion and remove unrelated stack residue.** The operation is still absent from current main.

Graph6 is a mathematical/domain encoding, so an explicit graph-owned conversion is compatible with the current rule against generic codec frameworks. It should live with graph semantics (or a private graph encoding adapter used by the public graph function), produce the canonical typed graph value, and not create a universal serialization/coercion layer.

Keep the independent checker as a separate operation if it provides value. Do not carry old relationship metadata, and drop the unrelated discrete-log edits from the current branch when restacking.

Recommendation: revise/rebase rather than close. Preserve exact decoding/canonicalization attack tests and ensure malformed/oversized encodings fail before graph backend work.

morluto:
Superseded by #1284. The replacement moves graph6 value/decoding semantics into `jacobian.math.graphs`, exposes one thin `OperationSpec`, and uses a dedicated standard-library checker runtime. It deliberately omits the old invariant mini-framework and reciprocal relationship machinery.

PR #984, opened 2026-08-10
fix(number-theory): accept large integer nth-root inputs

yuelgrace1810-ops:
## What changed

- Accept arbitrary-size canonical integer strings as `integer.compute.nth_root` radicands while retaining the bounded positive integer root degree.
- Parse the canonical radicand through Jacobian's canonical conversion API before calling SymPy.
- Update the catalog example and contract-representation fixtures to the canonical wire format.
- Add regression coverage for a realistic exact cube, its adjacent inexact value, and a negative odd root.

## Why

The implementation already uses SymPy's arbitrary-precision `integer_nthroot`, but the request schema restricted radicands to integers no larger than 1,000. A resolved-conjecture certificate required checking `729000000=900^3`; Jacobian discovered the exact operation but could not execute it because of that artificial contract bound. The bound also made the existing negative-odd-root implementation branch unreachable.

## Impact

Agents can now use the exact nth-root operation on realistic integer certificates and on negative radicands of odd degree. The wire representation becomes consistent with Jacobian's other arbitrary-size exact integer operations. Root degree remains bounded at 1,000.

## Validation

After rebasing onto current `main`:

- focused domain and composition suite: 16 passed
- Ruff check: passed
- `git diff --check`: passed

Earlier broader official `make test-changed BASE=origin/main` validation:

- passed: static/build, unit (863), component (765), domain (188), composition (493), storage (126), MCP (52), e2e (7), deploy (12)
- environment-only unavailable: npm and Markdown link checks (`npm`/`npx` are not installed)
- process lane: 242 passed, 1 skipped, 2 unrelated macOS failures (`-n` substring appears inside the checkout path `jacobian-nth-root`; system Bash 3.2 lacks `mapfile`)


morluto:
Second-pass review: **request changes for a mathematical semantics bug**. `integer.compute.nth_root` promises the floor nth root, but the negative odd-radicand path computes `integer_nthroot(abs(value))` and then negates it. That is correct only when the root is exact. For example `value=-9, degree=3` returns `-2, exact=False`, while `floor(cuberoot(-9)) = -3`. Please handle inexact negative odd roots with the extra decrement and add a regression such as `-9,3 -> -3,false`.

morluto:
Tracking issue opened: #1012 — fix floor semantics for inexact negative odd nth roots.

yuelgrace1810-ops:
Addressed in eb2d32e2: inexact negative odd roots now apply the additional decrement required for floor semantics, with a regression covering `-9, degree 3 -> -3, exact=False`. Focused nth-root tests pass (5 passed).

PR #985, opened 2026-08-10
Add a passive external reasoning observer

### LINKED ISSUE #955: [Design]: Add a harness-side observer for self-reported agent decisions
## Problem

Jacobian can observe host and MCP events, but those events do not explain why an
agent selected, skipped, retried, or stopped using a mathematical capability.
Some future evaluations may need bounded, model-authored decision summaries to
test a concrete hypothesis about tool selection or evidence use.

That observation must not return to the production MCP contract. A production
`reasoning.write` tool or reasoning fields on `math.run` change the catalog and
the behavior being measured, couple mathematical execution to a prescribed
workflow, and create configuration-dependent schemas. They also cannot expose
private chain-of-thought: any recorded text is an explicit self-report.

## Implementation gate

Do not implement the observer until an evaluation defines all of the following:

- a named, frozen experiment and the decision-level hypothesis it tests;
- control and treatment conditions, including whether the observation channel
  itself is part of the treatment;
- the exact event points and fields needed for scoring;
- how missing, malformed, or refused self-reports affect observation
  completeness; and
- retention, redaction, and publication rules for model-authored text.

Without that contract, adding an observer would be speculative infrastructure
and would risk recreating the removed production reasoning protocol in another
location.

## Proposed boundary

When the gate is met, add the observer to the agent/evaluation harness rather
than the Jacobian MCP server:

- expose an evaluation-only, host-supported event channel or tool for bounded
  decision summaries;
- keep its schema versioned and independent of `math.find`, `math.run`, and
  `capability://catalog`;
- let the harness attach trial, turn, and tool-call correlation from observed
  host identifiers instead of requiring the agent to copy opaque IDs;
- store events with the trial artifacts and bind normalized observations to the
  frozen experiment configuration;
- treat summaries as self-reported evaluation data, never as mathematical
  evidence, assurance, or checker authorization; and
- enforce any required event sequence in the evaluation normalizer or verifier,
  not in production capability execution.

A minimal event should contain a schema version, bounded summary, event kind,
sequence number, and harness-owned correlation. Experiment-specific fields
belong in the experiment contract; avoid another flat optional-field bag whose
valid shape changes by phase.

## Acceptance criteria

- A concrete experiment and scoring contract are linked before implementation.
- Control and treatment runs use frozen, comparable configurations, and the
  observer's behavioral effect is explicit.
- The observer is available only through the evaluation harness.
- Production MCP tool names, input schemas, guidance, and annotations are
  unchanged when observation is enabled.
- Events are bounded, typed, ordered, correlated to observed host/tool events,
  and retained under an explicit privacy policy.
- Missing or invalid events fail the observation contract closed without
  changing the mathematical result or its assurance.
- Tests cover malformed events, missing required events, correlation mismatch,
  cross-trial isolation, redaction, and proof that the production MCP catalog is
  identical with and without the experiment.

## Non-goals

- Capturing or inferring private chain-of-thought.
- Restoring a production reasoning mode or `reasoning.write` tool.
- Requiring one decomposition, tool order, verification order, or stopping
  strategy in ordinary Jacobian use.
- Treating a self-report as evidence that a decision was correct.

## Related work

- #944 owns the stable production `math.run` contract and identifies the
  evaluation harness as the correct enforcement boundary.
- #948 documents the phase-dependent schema and guidance problems in the
  superseded production reasoning tool.

kaoru0822-kitauji:
## Summary

- add a passive, harness-side observer for Codex JSONL or Harbor ATIF agent messages plus Jacobian MCP runtime telemetry
- retain only bounded explicit agent messages and exact server-emitted event fields; exclude prompts, hidden reasoning, tool arguments, and tool results
- fail observation closed on missing, malformed, cross-trial, or symlink-escaping sources without affecting mathematical execution or assurance
- freeze and run an authenticated weak-model pilot, with typed tests and operator documentation

## Why

The retired production reasoning workflow interfered with ordinary tool use: it added `reasoning.write`, changed `math.run` schemas by mode, required model-managed IDs, and encouraged a PLAN/BEFORE_TOOL/AFTER_TOOL/FINAL state machine. That made logging compete with mathematical discovery and invocation.

This change puts observation after the agent and verifier have exited. It makes no changes under `src/`, adds no MCP tool or CLI runtime mode, and leaves `math.find` and `math.run` untouched.

Closes #955.

## User and developer impact

Operators can derive a schema-versioned JSON record with:

- server-observed `math.find` / `math.run` metadata and capability-attempt outcomes;
- optional user-visible agent progress messages;
- source digests, trial-root binding, redaction/truncation metadata, coverage, and fail-closed diagnostics.

The record is observational only. It cannot authorize a checker, promote assurance, gate discovery, or block a capability invocation.

## Pilot evidence

The frozen post-run paired pilot used authenticated Codex CLI `0.147.0`, `gpt-5.4-mini`, medium reasoning, three public workflow tasks, and one fresh local MCP runtime per task.

- verifier outcomes: 3/3 pass in both baseline and post-run treatment
- false certifications: 0/3
- server event coverage: 6/6
- explicit agent messages retained: 25/25, with two bounded truncations
- agent-side calls: six `math.find`, two client-cancelled `math.run`, zero parameter errors
- production source delta: none
- observer wall time: 0.07-0.09 seconds per trial including `uv run` startup

Docker and Podman were unavailable, so the pilot is explicitly documented as a host-run exploratory workflow observation, not a Harbor container or causal-performance claim. Raw traces remain host-local; checked-in documentation binds their digests.

## Validation

- `make test-plan BASE=origin/main`
- `make lint typecheck`
- `make docs-linkcheck`
- `make test-unit` — 875 passed
- focused observer tests — 9 passed
- authenticated weak-model pilot — 3/3 path-adapted independent verifier replays passed, zero false certifications

### DIFF
--- a/benchmarks/config/external-reasoning-observer-v1.json
+++ b/benchmarks/config/external-reasoning-observer-v1.json
@@ -1,0 +1,63 @@
+{
+  "schema_version": "1",
+  "study_id": "external-reasoning-observer-v1",
+  "status": "FROZEN_BEFORE_IMPLEMENTATION",
+  "evidence_class": "EXPLORATORY_PAIRED_POST_RUN_OBSERVATION",
+  "causal_claim_authorized": false,
+  "frozen_against": {
+    "git_sha": "9653dd991a47c1be3cf8ad478450b69c8be01482",
+    "git_ref": "origin/main"
+  },
+  "research_question": "Can a passive harness-side observer retain useful server-observed Jacobian events and optional explicit agent progress summaries without changing capability discovery, invocation, or mathematical outcomes?",
+  "decision_hypothesis": "Keep the observer only if it leaves the model-facing MCP surface and every paired trajectory outcome unchanged, fails observation closed without affecting mathematics, and records all parseable math.find/math.run server events.",
+  "intervention": {
+    "baseline": "The completed agent trajectory and Jacobian runtime log without observer output.",
+    "treatment": "Run the observer only after the agent process and verifier have completed, over immutable copies of the same trajectory and runtime log.",
+    "model_visible_change": "NONE",
+    "production_mcp_change": "NONE",
+    "prompt_change": "NONE",
+    "observer_effect": "Post-run CPU time and one additional derived artifact only. The observer cannot affect the paired agent or verifier execution."
+  },
+  "runtime": {
+    "agent": "codex",
+    "agent_authentication": "locally authenticated ChatGPT session",
+    "model": "gpt-5.4-mini",
+    "reasoning_effort": "medium",
+    "web_search": "disabled",
+    "repetitions_per_task": 1,
+    "task_timeout_seconds": 900,
+    "sampling_seed": null,
+    "sampling_deterministic": false
+  },
+  "tasks": [
+    "jacobian/graph-counterexample",
+    "jacobian/finite-field-irreducibility-repair",
+    "jacobian/symbolic-block-determinant-decomposition"
+  ],
+  "primary_safety_outcomes": [
+    "canonical MCP tool names and input schemas are identical before and after the implementation",
+    "paired math.find count is unchanged",
+    "paired math.run count is unchanged",
+    "paired tool-error and parameter-error counts are unchanged",
+    "paired verifier result and false-certification classification are unchanged",
+    "missing or malformed observer input does not change or block any agent, verifier, or mathematical result"
+  ],
+  "secondary_observation_outcomes": [
+    "server event coverage",
+    "explicit user-visible agent-message coverage",
+    "observer status and diagnostics",
+    "observer wall time",
+    "derived artifact bytes",
+    "agent prompt, cached, and completion tokens",
+    "agent wall time and tool-call count"
+  ],
+  "interpretation": {
+    "success": "All primary safety outcomes pass and the observer preserves every parseable server event. Partial or zero explicit-summary coverage is acceptable and must be reported.",
+    "failure": "Any product-surface delta, changed paired outcome, false certification, or observer error that affects agent, verifier, discovery, invocation, assurance, or mathematical execution.",
+    "limitations": [
+      "The small non-deterministic sample estimates usability, not a general performance effect.",
+      "The post-run paired design establishes non-interference structurally; it does not estimate the effect of prompting a model to write summaries.",
+      "Only explicit agent messages are eligible summaries. Hidden chain-of-thought and ATIF reasoning_content are excluded."
+    ]
+  }
+}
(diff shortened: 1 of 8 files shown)


morluto:
Second-pass architecture review: **the production boundary is right, but please do not close #955 with this PR**. Keeping the observer under benchmark/evaluation tooling and out of `src/`, `math.find`, and `math.run` is aligned with #956. However #955 is specifically about a gated harness-side *decision self-report* channel (why an agent selected/skipped/retried/stopped) with harness-owned correlation. This PR passively harvests ordinary user-visible agent messages and allows zero summaries as complete, which is useful telemetry but does not fully implement that design. Please change `Closes #955` to `Related to #955` (or equivalent).

morluto (commented):

**Disposition: retain and rebase; architecturally aligned.**

This is the right observation boundary: it changes no `src/` code, adds no MCP tool or mode, and runs only after the agent/verifier process exits. Preserve that strict passivity—observer failure, scoring, or promotion must never affect execution, checker authority, or the recorded mathematical result.

Before merge, rebase onto #1256 and update telemetry extraction to the exact current two-tool/result contract (`math.find`, `math.run`, `execution`, domain-owned `output`, `diagnostics`, `verification_record_uri`, artifact/value-reference facts actually emitted). Keep prompts, hidden reasoning, tool arguments/results, and model-event payloads excluded; retain the symlink/cross-trial and bounded-size fail-closed tests.

The pilot evidence predates the final cutover, so either label it historical or rerun it before drawing current-product conclusions. Also remove generic assurance/coverage-promotion language where it could be confused with deleted runtime metadata; this should remain external evaluation telemetry only.

morluto:
Post-#1256/#1265 review: **keep the passive-observer idea, but revise/rebase the implementation and docs against the current wire surface.**

This PR is correctly outside `src/`: observing model-visible MCP traffic is evaluation infrastructure, not something the runtime should own. That boundary still makes sense.

What is stale is the contract it observes/documents. Current main uses nested discriminated `math.find` requests, result v2 with top-level `verification_record_uri`, a fixed two-tool surface, and separate local/remote hosts. Rebase the parser/tests/docs to those facts and make the observer fail closed on unknown current event shapes instead of preserving old assurance/relationship fields.

Recommendation: update and keep as evaluation tooling if it still answers a distinct question from #1259. Do not let the observer schema become a new product/telemetry authority; it should describe what the current surface emitted.

morluto:
### Superseded by the current observable-trajectory study

Closing this older passive-observer implementation in favor of #1259.

#985 established the right product boundary—post-run observation only, no production reasoning protocol—but its pilot was host-local, three tasks, and based before the current MCP/runtime/telemetry shape. #1259 now provides the stronger current-tree evidence:

- a genuinely held-out 12-task / 24-trial replication;
- current top-level verification-record evidence semantics;
- family-held-out diagnostics, uncertainty, baselines, and field ablations;
- no product/runtime behavior change;
- explicit `INCONCLUSIVE_RESEARCH_ONLY` disposition.

Keeping two passive observer/analyzer implementations would duplicate privacy, redaction, telemetry, and report contracts. The original pilot and implementation remain recoverable from this PR's immutable history.

**Disposition: superseded by #1259; historical pilot retained.**

PR #1007, opened 2026-08-10
Extend bounded independence number to order 128

kaoru0822-kitauji:
## Summary

- give `graph.invariant.independence_number.compute` a domain-owned request, result, and obligation contract through order 128;
- replace the complement-clique threshold loop for this capability with one wall-clock-bounded Z3 optimization seeded by a NetworkX feasible witness;
- return `EXACT` only when Z3's objective lower and upper bounds coincide with the emitted witness, otherwise return `UNKNOWN` with the incumbent, conservative upper bound, and open optimality obligation;
- preserve the existing capability ID and budget fields, bump its contract version to `2`, and leave clique-number behavior unchanged.

No checker is added. Results remain `COMPUTED`, and exact claims retain an open `GRAPH_INDEPENDENCE_NUMBER_OPTIMALITY` obligation.

## Public reproduction

Clow, Haxell, and Mohar's 2025 Biggs–Smith counterexample uses the exact finite claim `alpha(B) = 43` on a 102-vertex cubic graph: https://arxiv.org/abs/2505.05339

On a fresh install of current main (`50ea5457`), the authenticated Codex pilot selected the existing independence-number capability but received `UNSUPPORTED_INPUT` because the shared graph contract stopped at order 32. On a fresh install of this branch, the same frozen task returned an exact 43-vertex witness in 678 ms. A separate SciPy/HiGHS MILP oracle confirmed optimum 43 and independently found the size-44 feasibility problem infeasible.

The regression reconstructs only the finite graph needed for that public case. It does not add a benchmark suite or encode a research workflow.

## Validation

- `make test-plan BASE=origin/main` selected static, docs, unit, component, domain, composition, storage, process, MCP, and E2E lanes.
- `make check-changed BASE=origin/main`: lint/typecheck, docs, unit, domain, composition, storage, MCP, and E2E passed on the final tree.
- The process lane's sole failure reproduces unchanged on a clean `50ea5457` worktree: macOS `/bin/bash` 3.2 lacks `mapfile`, which `scripts/setup-agent` uses.
- The component lane's sole final-run failure was the unrelated one-second `test_carcara_timeout_fails_closed` marker race. The same complete lane passed in the prior gate; focused repetitions produced both 3/3 and 2/3 passes, confirming host-timing nondeterminism.
- Focused graph domain file: 16 passed, including the 102-vertex exact result and a 129-vertex fail-closed request with no artifact writes.

Lean-dependent tests remained skipped because the pinned Lean runtime is not installed, as expected for the local environment.

### DIFF
--- a/docs/reference/capabilities/graphs/graph-invariant-batch.md
+++ b/docs/reference/capabilities/graphs/graph-invariant-batch.md
@@ -78,7 +78,7 @@ budgets, and in some cases use a different backend:
 
 | Standalone invariant | Batch registry relationship |
 | --- | --- |
-| `graph.invariant.independence_number.compute` | Same mathematical value; different input and artifact contract. |
+| `graph.invariant.independence_number.compute` | Same mathematical value; a dedicated bounded Z3 optimization accepts inline graphs through order 128 and returns a witness plus an open optimality obligation. |
 | `graph.invariant.girth.compute` | Same invariant; the standalone operation uses `0` for acyclic graphs while the batch uses `null`. |
 | `graph.invariant.diameter.compute` | Same invariant; the standalone operation uses `-1` for disconnected graphs while the batch reports `NOT_APPLICABLE`. |
 | `graph.invariant.clique_number.compute` | Not in registry version `1`. |
--- a/src/jacobian/contracts/graph_invariant_operations.py
+++ b/src/jacobian/contracts/graph_invariant_operations.py
@@ -30,6 +30,36 @@ class GraphMaximumMatchingRequest(ContractModel):
     graph: GraphMaximumMatchingGraph
 
 
+class GraphIndependenceNumberGraph(ChromaticGraph):
+    """A simple graph bounded for the dedicated independence-number search."""
+
+    vertices: tuple[GraphVertex, ...] = Field(max_length=128)
+    edges: tuple[tuple[GraphVertex, GraphVertex], ...] = Field(max_length=8_128)
+
+
+class GraphIndependenceNumberBudget(ContractModel):
+    """Explicit wall-clock and order limits for independence-number search."""
+
+    wall_seconds: StrictInt = Field(default=5, ge=1, le=120)
+    max_solver_calls: StrictInt = Field(default=1, ge=1, le=33)
+    max_order: StrictInt = Field(default=128, ge=0, le=128)
+
+
+class GraphIndependenceNumberRequest(ContractModel):
+    """One bounded simple-graph independence-number request."""
+
+    graph: GraphIndependenceNumberGraph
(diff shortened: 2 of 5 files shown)


morluto:
Second-pass architecture review: **draft looks sound**. This keeps the server responsible for a bounded mathematical invariant: one domain-owned request/result/obligation contract, explicit wall budget, exact only when optimizer bounds coincide, otherwise UNKNOWN with an incumbent and open optimality obligation. No workflow policy leaks into discovery and no checker authority is invented. I would proceed with the stated hosted-CI validation.

morluto (commented):

**Disposition: retain the bounded-search improvement, rewrite the result contract.**

The operation-specific order-128 scope and single bounded optimization run are reasonable. Delete the generic optimality obligation and represent outcomes as a domain-owned discriminated union: `Exact` must contain a witness whose size equals coincident lower/upper bounds; `Bounded`/`Unknown` must contain an incumbent, conservative upper bound, termination reason, and no exact claim. Those impossible-state checks belong in the graph domain value.

Rebase onto current `main`, keep the capability ID, make the graph/order/time budgets operation-owned, and avoid a new generic optimization-result framework. Preserve the 102-vertex regression, 129-vertex rejection, timeout/unknown cases, and independent MILP evidence as tests—not runtime assurance metadata.

morluto:
Post-#1256/#1265 review: **revise/rebase; keep the bounded independence-number capability, simplify the result model.**

The useful semantic slice remains compelling: support graphs up to the intended order, return incumbent/lower/upper bounds plus witness, and distinguish exact optimum from bounded/incomplete search. That fits the current rule that a completed bounded operation may return a domain-owned `UNKNOWN`/`INCOMPLETE` result.

What I would not carry forward is the extra `GraphIndependenceNumberObligation` wrapper/artifact if it merely restates the same status, bounds, witness, and required checks already present in the typed result. #1256 deliberately removed generic obligation/completeness machinery; for this operation the result itself can make the bounded state unambiguous, and a separate checker operation can verify a claimed optimum/witness when needed.

Recommendation: port the operation to the current graph semantic/backend ownership, collapse duplicated obligation state into the domain result, keep preflight bounds before solver work, and rebase the tests/docs accordingly.

morluto:
Superseded by #1297 after the typed-result ownership rewrite. The original order-128 independence-number capability and Biggs–Smith reproduction here remain credited to @kaoru0822-kitauji; #1297 is a current-tree simplification/port, not a rejection or reassignment.

morluto:
Superseded by #1297, which carries the current-architecture port and explicitly credits @kaoru0822-kitauji for the original investigation, reproduction, and implementation work. Closing this older branch so there is one authoritative implementation.

PR #1020, opened 2026-08-10
Extend exact determinant bound to order 64

kaoru0822-kitauji:
## What changed

- give `matrix.determinant.compute` a determinant-owned exact `QQ` matrix input capped at order 64 while leaving the shared matrix contract at 32
- apply the same order-64 and 256-digit source bounds in the independent Python-FLINT replay, plus the existing 32,768-digit candidate bound
- bump the experimental determinant producer to version 3 and document the operation-specific scope
- add the p=71 truncated Legendre-symbol determinant (`33 x 33`, determinant `529`) as a compute-and-verify regression, with order-65 fail-closed producer/checker cases

## Why

Yang, Yang, and Zhang's arXiv:2606.22548v1, Theorem 1.1 gives a proof-critical exact order-33 determinant. The existing producer and checker use exact polynomial-time backends, but the producer inherited the shared order-32 matrix boundary used by heavier rank, RREF, and multiplication operations. A fresh control agent found the correct producer/verifier and received `INVALID_REQUEST` at `matrix/entries` with `maxItems=32`.

Local timings were about 0.055 s (SymPy Bareiss) and 0.0016 s (Python-FLINT) for the order-33 source. A seeded order-64 small-integer matrix took about 0.505 s and 0.0014 s, with exact agreement.

## Evaluation caveat

This is an experimental capability-scope fix, not evidence of autonomous-agent improvement. In a frozen `gpt-5.4-mini`, low-reasoning, MCP-only pilot, treatment made producer version 3 discoverable but completed 0/3 attempts: one model timeout before invocation, one ragged matrix request, and one 81-row request. Jacobian failed closed each time and produced no false conclusion or verification record. The direct runtime regression did compute `529` and the independently authorized Python-FLINT replay returned `VERIFIED`.

## Validation

- focused matrix producer/checker tests: `127 passed`
- `make test-plan BASE=origin/main`
- `make check-changed BASE=origin/main`: 8/10 commands passed on the rebased
  tree. The selected component lane had two unrelated one-second SAT/SMT
  timeout failures (the SMT node passed serially; the DRAT marker still missed
  its startup window), and the process lane's source-bootstrap test requires
  Bash `mapfile` while this macOS host only provides Bash 3.2. The immediately
  preceding selected run passed the complete component lane; the Bash failure
  reproduced unchanged.

The broader selected gate was rerun after rebasing onto the then-current upstream `main`.

### DIFF
--- a/docs/reference/capabilities/matrix/matrix-rational-determinant.md
+++ b/docs/reference/capabilities/matrix/matrix-rational-determinant.md
@@ -7,17 +7,19 @@ separate trust boundaries.
 
 ## Input and result contracts
 
-The producer accepts one [`RationalMatrix`](../index.md#shared-matrix-values)
-from `jacobian.contracts.matrices`: a nonempty square matrix with at most 32
-rows and columns. Every entry is a canonical reduced rational:
+The producer accepts one determinant-owned exact rational matrix: a nonempty
+square matrix with at most 64 rows and columns. This operation-specific bound
+does not widen the shared 32-row and 32-column [`RationalMatrix`](../index.md#shared-matrix-values)
+used by rank, RREF, multiplication, and other matrix operations. Every entry is
+a canonical reduced rational:
 
 ```json
 {"num": "-3", "den": "7"}
 ```
 
-The shared `RationalMatrix` model permits up to 32,768 canonical digits per
-scalar component; the determinant request model tightens this to 256 decimal
-digits via its own `require_matrix_scalar_digits` validator.
+The determinant input value has the same canonical `QQ` matrix semantics. The
+determinant request model limits every scalar component to 256 decimal digits
+via its own `require_matrix_scalar_digits` validator.
 
 `matrix.determinant.compute` uses SymPy's exact matrix determinant API with the
 fraction-free Bareiss method. It returns the bounded canonical determinant
--- a/src/jacobian/contracts/matrix_operations.py
+++ b/src/jacobian/contracts/matrix_operations.py
@@ -19,6 +19,7 @@ from jacobian.contracts.matrices import (
 from jacobian.contracts.results import ContractModel
 
 MAX_INPUT_SCALAR_DIGITS = 256
+MAX_DETERMINANT_MATRIX_DIMENSION = 64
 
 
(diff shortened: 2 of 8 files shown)


morluto (commented):

**Disposition: retain the bound change, rewrite/rebase on the authoritative matrix path.**

The evidence supports an operation-specific determinant bound above the heavier shared matrix limits. After #1252/#1256, express that directly in the determinant request/value contract under `jacobian.math`, not by reviving a shared contract override or legacy versioned adapter.

Use the same order/digit limits in the separately authorized checker, keep order-65 fail-closed behavior, and preserve the order-33/order-64 timing and exact-agreement regressions. Update all result/verification assertions to current records and remove generic assurance/semantics metadata. The determinant operation should own this budget; rank/RREF/multiplication should remain unaffected.

morluto:
Post-#1256/#1265 review: **keep the larger determinant capability, but revise/rebase it around the current matrix semantic type and operation budget.**

The motivation still makes sense: determinant can safely support a larger square-matrix bound than heavier normal-form/lattice operations, so forcing one shared matrix request ceiling under-represents the backend. Operation-specific admissible bounds are preferable to a lowest-common-denominator global cap.

On rebase, avoid solving this by proliferating parallel matrix representations. The matrix value itself should retain one canonical semantic identity under `jacobian.math`; the determinant request/preflight may impose its own larger admissibility bound, and the private backend adapter can delegate to the maintained library. Keep the exact checker separate and make sure the larger bound is enforced before expensive allocation/provider work.

Recommendation: revise/rebase, preserve the 64×64 boundary/attack tests, and merge without reintroducing generic capability-result or relationship machinery.

morluto:
Superseded by #1302 after the current-tree matrix ownership cleanup. The original determinant-bound expansion, order-33 Legendre-symbol reproduction, and validation here remain credited to @kaoru0822-kitauji; #1302 is a port, not a rejection or reassignment.

morluto:
Superseded by #1302, which carries the current-tree port and explicitly credits @kaoru0822-kitauji for the original investigation, reproduction, and implementation work. Closing this older branch so there is one authoritative implementation.

PR #1182, opened 2026-08-10
Verify caller-supplied P-recursive tables exactly

### LINKED ISSUE #900: [Feature]: Add bounded exact verification for variable-coefficient recurrences

yuelgrace1810-ops:
## Summary

Adds the remaining bounded, independently checkable slice from #900 for recurrence tables produced outside Jacobian.

- adds `combinatorics.recurrence.p_recursive.table_residuals.compute`;
- accepts a complete caller-supplied rational table and never generates or repairs terms;
- records every exact residual, pass/fail status, and first failing index;
- adds `combinatorics.recurrence.p_recursive.table_residuals.verify`, which independently replays the submitted table with standard-library `Fraction`;
- documents that finite-table verification does not prove an infinite recurrence or a generating-function identity.

This is intentionally separate from `combinatorics.recurrence.p_recursive.evaluate`, whose job is to generate terms from initial values.

## Verification boundary

The producer remains `COMPUTED`. Only the separate checker may produce `VERIFIED`, and only for the exact finite table and recurrence relation bound into the checker request. An accurately reported failing table can also be verified as an accurate residual computation; no opposite theorem is inferred.

## Tests

Added coverage for:
- the A176677 recurrence/table through index 16;
- a corrupted late table term;
- exact reporting of a nonzero residual and first failure;
- rejection of a forged zero-residual/success result;
- malformed tables without a checked recurrence step;
- producer → independent verifier integration;
- portfolio representative-payload ownership.

Validation:
- focused recurrence contract/component/domain/verifier tests: passed;
- lint, formatting, complexity, mypy, docs links, and `git diff --check`: passed;
- unit: 891 passed;
- storage: 127 passed;
- MCP: 49 passed;
- e2e: 7 passed, 1 skipped (Lean/Mathlib unavailable);
- component: 808 passed, 3 skipped; one unrelated plugin-worker timeout passed on isolated rerun;
- domain: changed tests passed; two unrelated number-theory cases failed under parallel host contention and passed on isolated rerun;
- composition: the new portfolio-coverage failure was fixed and passed; one unrelated inverse-synthesis timeout passed on isolated rerun;
- process: 244 passed, 1 skipped; one existing macOS Bash 3.2 `mapfile` incompatibility remains outside this change.

Refs #900. This PR does not close #900 automatically so maintainers can confirm the remaining issue acceptance criteria after merge.


morluto (commented):

**Disposition: retain the capability, rewrite on current `main`.**

This is a distinct, nontrivial operation: verify a caller-supplied finite table against a declared P-recursive relation without generating or repairing terms. Model the recurrence, indexed rational table, and residual report as strict domain values/functions under the authoritative math layer; the public operation should be a thin `OperationSpec`, with the independent checker separately installed.

Keep the important semantic distinction that a checker may verify an accurately reported failing residual computation—it is not proving the recurrence or an infinite generating-function identity. Remove generic assurance/completeness/portfolio machinery, use the current verification-record binding, and add ports only if the whole table/report is actually composed elsewhere.

morluto:
Post-#1256/#1265 review: **keep the recurrence-table producer/checker capability, but revise/rebase it to the current typed-operation boundary.**

The mathematical need is still valid: accepting caller-supplied recurrence data and returning exact residual/consistency information is a useful atomic operation, and an independent replay checker is the right verification shape.

On rebase, keep the recurrence table/result as domain-owned typed values, put the mathematical evaluation function under the current math owner, bind it with one `OperationSpec`, and keep the checker separate. Remove any inherited generic computed-result/relationship plumbing from the older bundle architecture. A bounded failure/incomplete state should live in the recurrence result itself rather than a generic obligation wrapper.

Recommendation: revise rather than supersede; preserve the exact arithmetic/checker tests and update docs/examples to current `math.find`/`math.run` inspection semantics.

morluto:
Superseded by #1285. The replacement moves the submitted finite-table request/value/function into `jacobian.math.combinatorics`, keeps the domain wrapper thin, and independently replays every residual with `Fraction`. It retains no backend/assurance slogans or infinite-recurrence claims.

PR #1183, opened 2026-08-10
Allow reviewed approvals in Codex visibility evaluations

yuelgrace1810-ops:
## What changed

- run Codex visibility evaluations with `--approve-for-me`
- remove the hard-coded `-a never` policy
- assert the approval configuration in the visibility tooling unit test

## Why

`math.run` is intentionally annotated as potentially destructive because some capabilities may create artifacts. The visibility harness hard-coded `-a never`, so Codex automatically returned `user cancelled MCP tool call` for every execution attempt. This prevented the evaluation from measuring successful Jacobian adoption or task completion.

The harness already executes Codex in an isolated temporary workspace with a read-only sandbox. Reviewed automatic approval preserves that isolation while allowing the evaluation's required MCP calls to proceed.

## Impact

Model evaluations can now distinguish genuine Jacobian execution behavior from approval-policy cancellation. This makes task-success, call-count, token, and latency comparisons usable.

## Validation

- `pytest tests/unit/tooling/test_codex_visibility.py -q` — 18 passed
- paid smoke test with `gpt-5.4-mini`, high reasoning: `matrix.determinant.compute` completed successfully instead of being cancelled
- remote diff verified to contain only the harness and its unit test

### DIFF
--- a/benchmarks/tooling/codex_visibility.py
+++ b/benchmarks/tooling/codex_visibility.py
@@ -340,8 +340,7 @@ def _codex_arguments(
     tool_mode: ToolMode,
 ) -> tuple[str, ...]:
     arguments = [
-        "-a",
-        "never",
+        "--approve-for-me",
         "exec",
         "--ephemeral",
         "--skip-git-repo-check",
--- a/tests/unit/tooling/test_codex_visibility.py
+++ b/tests/unit/tooling/test_codex_visibility.py
@@ -135,6 +135,8 @@ def test_unified_exec_mode_is_opt_in(tmp_path: Path) -> None:
     direct = _codex_arguments(**common, tool_mode=ToolMode.DIRECT)
     unified = _codex_arguments(**common, tool_mode=ToolMode.UNIFIED_EXEC)
 
+    assert "--approve-for-me" in direct
+    assert "never" not in direct
     assert "unified_exec" not in direct
     assert unified[-3:-1] == ("--enable", "unified_exec")
     assert 'mcp_servers.jacobian.default_tools_approval_mode="approve"' in unified


morluto (commented):

This is evaluation-harness infrastructure, not a product-surface change, and the root cause remains valid: a blanket deny policy turns every required `math.run` attempt into an approval cancellation, invalidating the measurement.

Rebase onto current `main` and the currently pinned Codex CLI. Preserve the safety preconditions explicitly in tests/config: isolated temporary workspace, read-only sandbox, no host credential/file expansion beyond the intended MCP connection, and reviewed automatic approval only for this evaluation process. Update the smoke task and expected telemetry to the simplified two-tool/result contract.

morluto:
Post-simplification review: **keep/update; this is a harness-policy fix, not an architectural conflict.**

The PR removes the hard-coded `-a never` behavior and switches the Codex evaluator to its intended self-approval mode. That is independent of the `math.find`/`math.run` redesign and still matters because an evaluation harness should not silently impose a stricter approval policy than the condition being measured.

Before merge, rebase onto the current `codex_visibility.py` (it has changed materially during the recent tool-surface work) and verify the exact current Codex CLI flag/behavior used by the benchmark image. Keep the regression at the argument-construction boundary so a future CLI change fails loudly.

Recommendation: rebase/update and merge if the pinned runner still supports this flag. Do not mix product guidance changes into this PR.

morluto:
### Superseded by the current Codex MCP approval configuration

Closing this stale CLI-flag patch.

Current `main` no longer hard-codes `-a never`. The visibility harness configures the MCP server directly with:

```text
mcp_servers.jacobian.default_tools_approval_mode="approve"
```

through the current Codex configuration interface. That preserves the intended reviewed automatic approval without depending on the older `--approve-for-me` CLI spelling, and it is already exercised by the newer visibility/evaluation work.

**Disposition: root cause fixed on main through the current configuration path; old flag patch superseded.**

PR #1185, opened 2026-08-11
Default exact capability inspection to runnable contracts

yuelgrace1810-ops:
## What changed

- make an exact `math.find(capability_id=...)` lookup return the runnable `CONTRACT` projection by default
- retain explicit `SUMMARY`, `CONTRACT`, and `FULL` views
- preserve compatibility for discovery calls that explicitly send `view="SUMMARY"`
- align MCP guidance, bundled skills, reference docs, and boundary coverage

## Why

Agents currently need a compact exact lookup followed by a second exact lookup with `view="CONTRACT"` before invocation. For a known capability ID, the first lookup does not provide the validated input schema needed to run the operation. Returning the runnable contract directly removes that promotion step while keeping compact fit-only and full audit projections available as explicit overrides.

## Evaluation

A corrected paired evaluation used `gpt-5.4-mini` at high reasoning across integer GCD, polynomial GCD, and independently verified determinant cases, with three repetitions per case and identical harness behavior in both conditions.

| Metric | Baseline | Pilot | Change |
|---|---:|---:|---:|
| Successful contracts | 5/9 | 7/9 | +2 |
| Verified runs | 0/9 | 3/9 | +3 |
| MCP calls | 35 | 23 | -34.3% |
| Tool errors | 11 | 3 | -72.7% |
| Model-visible MCP bytes | 94,288 | 67,081 | -28.9% |
| Uncached input tokens | 154,266 | 104,161 | -32.5% |
| Output tokens | 21,135 | 12,762 | -39.6% |
| Elapsed time | 504.7 s | 349.0 s | -30.9% |

Polynomial success remained 3/3. Independent verification improved from 0/3 to 3/3. Integer-GCD adoption varied from 2/3 to 1/3, so the evaluation does not claim universally higher tool adoption on simple tasks.

A deterministic matched pilot also measured 50% fewer calls, 31.2% fewer model-visible response bytes, and 27.2% lower summed median latency.

## Validation

- 23 focused unit and MCP boundary tests passed
- Ruff lint passed
- Ruff formatting check passed
- mypy passed for the changed MCP source files
- packaged and repository skills remain byte-identical

This PR is independent of #1183; it does not include the evaluation-harness approval fix.

### DIFF
--- a/.agents/skills/jacobian-math/SKILL.md
+++ b/.agents/skills/jacobian-math/SKILL.md
@@ -44,8 +44,8 @@ Payloads:
 For other outcomes, query `math.find` by plain-language outcome; no ID is
 required. Use low `limit` only for query search, never with `capability_id`. For
 a selected operation's schema, call
-`math.find({"capability_id":"<exact-id>","view":"CONTRACT"})`; never put
-`CONTRACT` in a query. A card's
+`math.find({"capability_id":"<exact-id>"})`; exact lookup returns the runnable
+contract by default; never put `CONTRACT` in a query. A card's
 `invocation_example`, or required top-level fields, may be enough.
 
 Add no domain filter unless its installed spelling is known. Follow exposed
--- a/docs/reference/tools.md
+++ b/docs/reference/tools.md
@@ -117,26 +117,23 @@ arguments, or inspect `capability://catalog`. Array position is not a preferred
 next step. These paths do not imply that the requested operation is
 mathematically impossible or absent in principle.
 
-Passing one `capability_id` returns the default `SUMMARY` exact projection. It
-contains the one-line outcome, tags, provider availability, input/output field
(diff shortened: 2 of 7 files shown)


morluto:
wow looks good

morluto:
taking a look at the approach

morluto (commented):

**Disposition: close as superseded by the current `math.find` contract.**

The underlying ergonomics goal landed in a simpler form. Current exact inspection returns the full runnable descriptor directly; the old `SUMMARY`/`CONTRACT`/`FULL` promotion model, compact projections, and managed-Skill guidance were removed by #1252/#1256.

Rebasing this PR would restore view-state and compatibility branches that the cutover intentionally deleted. Keep the historical evaluation as evidence for exact inspection, but there is no remaining production patch to merge.

Issue #28, opened 2026-07-25
[Feature]: Make capability contracts easier for agents to discover and compose


morluto:
**Priority**

P2 (medium)

**Area**

mcp, kernel

**Problem**

Agents can call `capability.describe` and `capability.invoke`, but successful mathematical composition still depends on understanding exact scope, evidence type, artifact flow, and failure recovery from compact tool descriptions. The initial paired transcripts show that opaque artifact outputs and weak provenance cues can cause an agent to reconstruct or relabel an object instead of using the exact returned artifact.

Adding more capability IDs will increase parameter mistakes and ambiguous composition unless the shared agent-facing contract becomes clearer first.

**Proposed solution**

Evaluate and implement the smallest shared ergonomics improvements that reduce measured transcript failures:

- concise valid examples and one representative invalid example in `capability.describe`;
- explicit exact/approximate and bounded/exhaustive fields where relevant;
- actionable parameter errors that name the invalid field, accepted shape, and safe retry;
- concise inline output summaries alongside durable artifact URIs;
- explicit provenance links for ordered calls and consumed/produced artifacts;
- batch invocation only for repeated homogeneous calls demonstrated in transcripts.

Keep these fields versioned and machine-readable. Domain plugins continue to own mathematical schemas and semantics; the kernel owns common execution, scope, completeness, assurance, artifact identity, and provenance.

Success should be measured on held-out multi-call tasks. Compare parameter-error rate, unnecessary calls, artifact substitution, correctness, false certification, runtime, and token use before and after each contract change. Consolidate fields that agents ignore and avoid duplicating the same information in incompatible locations.

**Alternatives considered**

- Add a large set of top-level MCP tools with long descriptions: rejected because capability IDs behind the existing describe/invoke interface are easier to version and compose.
- Add a generic artifact-storage browser as the first fix: deferred; returning the small typed payload with its artifact URI solved the observed graph provenance problem with less surface area.
- Add batching unconditionally: rejected because batching can obscure per-operation status and provenance unless transcripts demonstrate repetitive homogeneous calls.

**Impact on trust boundaries**

Ergonomic summaries and examples are descriptive, not evidence. They must not override durable artifacts, checker bindings, execution status, completeness, or assurance. Parameter recovery must remain fail-closed, and a batch result must preserve per-item status and verification records.

**Additional context**

PR #22 contains the first relevant transcript evidence and shared telemetry. Coordinate this issue with #24 and #25 so contract changes are evaluated rather than added speculatively.

[...]

PR #1187, opened 2026-08-11
Check execution status before using math output

yuelgrace1810-ops:
## What changed

- make the `math.run` tool description tell agents to check execution status before treating `output` as mathematical evidence
- require completed runs to be interpreted through conclusion, completeness, and assurance before their mathematical value is used
- add a focused regression test that preserves the safety ordering

## Why

The current description says to read the mathematical output first and execution status second. That ordering conflicts with Jacobian's fail-closed contract: cancelled, invalid, timed-out, incomplete, or otherwise non-conclusive runs must not be promoted into mathematical evidence.

Recent held-out trajectories repeatedly showed the reliability risk around cancelled calls and unsupported assurance promotion. Those trajectories do not establish a backend mathematics defect, but they make the contradictory instruction unsafe to retain. This is a narrow contract-guidance correction adjacent to issue #28; it does not change mathematical operations, schemas, or assurance semantics.

## Validation

- focused unit and MCP boundary tests passed
- Ruff lint passed
- Ruff formatting check passed
- mypy passed for the changed source file
- `git diff --check` passed

This PR is independent of draft #1185 and does not alter exact capability inspection.

### DIFF
--- a/src/jacobian/adapters/mcp/guidance.py
+++ b/src/jacobian/adapters/mcp/guidance.py
@@ -52,9 +52,10 @@ Examples:
 """
 
 MATH_RUN_DESCRIPTION = """\
-Run one installed math tool by ID with its typed `payload`. Read the mathematical
-value in `output` first, then execution status. If the payload shape is unknown,
-use math.find with view CONTRACT.
+Run one installed math tool by ID with its typed `payload`. Check execution status
+before treating `output` as mathematical evidence. For a completed run, also inspect
+conclusion, completeness, and assurance before using the mathematical value. If the
+payload shape is unknown, use math.find with view CONTRACT.
 
 Ordinary tools return calculations. Independent checking uses a separate checker
 tool ID (for example `polynomial.identity.verify` or `case.partition.finite.verify`),
--- a/tests/unit/tooling/test_mcp_tool_surface.py
+++ b/tests/unit/tooling/test_mcp_tool_surface.py
@@ -56,3 +56,12 @@ def test_guidance_rejects_verification_transfer_to_derived_claims() -> None:
     combined = "\n".join((SERVER_INSTRUCTIONS, MATH_RUN_DESCRIPTION))
     assert "does not verify a model-derived conclusion" in combined
     assert "record must be bound to the exact final claim" in combined
+
+
+def test_math_run_guidance_checks_execution_before_using_output() -> None:
+    guidance = MATH_RUN_DESCRIPTION.lower()
+
+    status_position = guidance.index("check execution status")
+    output_position = guidance.index("treating `output` as mathematical evidence")
+    assert status_position < output_position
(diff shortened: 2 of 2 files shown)


morluto (commented):

**Disposition: retain and revise for the simplified result contract.**

The safety ordering is correct and current guidance still needs it: callers must inspect `execution` before treating `output` as mathematical evidence. However, the proposed second step references deleted generic conclusion/completeness/assurance fields.

Rebase onto current `main` and phrase the rule against the surviving envelope: require `execution.status == COMPLETED`; then inspect the domain-owned output/status and any diagnostics; treat verification as a separate, locally bound verification-record URI rather than a generic assurance level. Keep a focused MCP-description regression for that ordering only.

morluto (commented):
### Live-audit update after #1258 opened

**Updated disposition: supersede this branch with #1258 once the execution-first invariant is ported.**

#1258 is the current-main owner for the MCP titles and model-visible routing guidance, and it edits the same `guidance.py` surface. Carry this PR’s valid safety rule into #1258, but rewrite it for the simplified result contract:

- check `execution.status` before treating `output` as evidence;
- inspect the completed operation’s domain-owned output/status and diagnostics;
- require an exact bound `verification_record_uri` for independent verification.

Do not carry the deleted generic conclusion/completeness/assurance fields or old inspection-view wording. Once #1258 contains the ordering regression, close #1187 as superseded rather than rebasing two overlapping prompt branches.

morluto:
Post-#1256/#1265 review: **the bug is still real, but this PR should be superseded by a revised #1258 rather than merged as-is.**

Current main still tells agents to read `output` first and execution status second, so the safety-ordering correction remains worthwhile. However this branch is written against removed concepts: `conclusion`, generic `completeness`, generic `assurance`, `math.find` `view=CONTRACT`, and the deleted `case.partition.finite.verify` workflow.

The current rule should be much smaller: check `execution.status` first; only a completed invocation makes its typed `output` eligible for mathematical use; then interpret any domain-owned `UNKNOWN`/`INCOMPLETE` status inside that result; independent verification is separate and indicated by a bound `verification_record_uri`.

Because #1258 already edits the same guidance/test surface and needs a current-main rebase, I recommend moving this status-first correction there and closing #1187 as superseded once that lands. It may also be worth ordering the small MCP text projection as `execution` before `output` so the wire presentation and prose agree.

morluto:
### Superseded by #1269

The status-before-output defect is valid and has been ported into #1269, alongside the current local-routing guidance.

The replacement deliberately drops this branch's references to generic `conclusion`, `completeness`, and `assurance` interpretation. Post-#1256/#1265, completed runs are interpreted through operation-owned result fields, while independent verification is established by a bound `verification_record_uri` from a separate checker operation.

#1269 adds a focused regression proving that status appears before output use and that generic assurance/completeness wording is absent. Closing this stale branch as fully superseded.

Issue #1140, opened 2026-08-10
[Superseded]: Reuse concrete mathematical value models; do not add an inline type registry


morluto:
## Disposition

Closed as not planned. Superseded by #1477, #1491, and the direct-composition dispositions #1208/#1216.

This issue correctly identified duplicated exact-integer, rational, matrix, polynomial, graph, label/order, and canonicalization definitions across operation schemas.

Its proposed replacement—`InlineMathematicalTypeDescriptor`, type/semantics/canonicalization IDs and digests, a local content-addressable registry, compatibility versions, bundled references, catalog lint, compact summaries, and value-reference integration—would create another schema/type platform beside the concrete Python models.

## Simpler endpoint

When several operations genuinely share a mathematical value, share the actual concrete Python/Pydantic/dataclass type and parser through ordinary imports:

```text
CanonicalInteger
ExactRational
RationalMatrix
SparsePolynomial
LabelledGraph
```

The owning domain/module defines the value once. Operation request/result models embed or reference that concrete class. Pydantic generates self-contained MCP JSON Schema as needed.

Interpretation-critical facts such as field/ring, labels, axes, variable order, basis, and dimensions remain in the concrete value.

If two values have different semantics or canonicalization, use different concrete types. No registry decision is needed.

## Retained work

- consolidate duplicate scalar/value classes;
- keep bounded resource integers distinct from unbounded mathematical integers;
- keep labels/order attached to positional values;
- remove local recanonicalization that disagrees with the shared type;
- reuse ordinary `$defs` generated by Pydantic where it naturally reduces schema duplication;
- keep compact human discovery summaries as simple static description text/examples.

## Do not introduce

- inline type IDs/URIs/digests as a runtime type system;
- a mathematical type/canonicalizer/compatibility registry;
- content-addressed type descriptors;
- type migration/read/write compatibility machinery;
- catalog-wide type registration/lint infrastructure;
- value-reference or port integration;
- a universal JSON mathematical ontology.

A release may version a public concrete model when compatibility actually requires it. Do not prebuild a registry for every value.

## Canonical owners

- #1491 — one semantic owner per mathematical value.
- #1477 — concrete model/parser is authoritative; generated schema is a projection.
- #1483 — ordinary serialized values compose without references/type registries.
- domain-specific cleanup issues such as #1489, #1478, and #1479.

morluto:
## Type registry role under the composition spine

[...]

PR #1188, opened 2026-08-11
fix(polynomials): canonicalize Gaussian moment polynomial inputs

yuelgrace1810-ops:
## What changed

- canonicalize bounded Gaussian-moment sparse polynomial inputs before capability validation
- sort exponent vectors, combine duplicate monomials exactly over `Q(i)`, and remove zero terms
- preserve the strict canonical `GaussianPolynomial` value contract and all raw-input/resource bounds
- add focused regressions for unordered terms, duplicate/zero handling, canonical stored values, raw term limits, and all-zero rejection

## Why

Merged PR #920 canonicalizes the shared rational sparse-polynomial boundary, but `probability.gaussian_polynomial.moment.compute` uses a separate complex-rational sparse representation and retained its strict caller-side ordering requirement.

A saved held-out trajectory based on Long's resolved Gaussian Moments conjecture supplied a mathematically valid ten-term polynomial in intuitive rather than lexicographic order. Current main rejected it before computation. Replaying that exact saved payload against this branch now canonicalizes exponents from `(0,0,2)` through `(2,0,2)` and computes the exact zero first moment.

This is a localized extension of #920's deterministic-input principle. PR #943 improved the ordering diagnostic but did not remove the caller-side normalization burden. Issue #1140 owns the broader reusable-wire-type architecture; this PR does not attempt that redesign.

## Safety boundary

- raw requests remain capped at 16 terms before any combining
- exponent dimensions, degrees, coefficient digit bounds, and expansion budgets still fail closed
- an all-zero polynomial remains invalid
- direct `GaussianPolynomial` values remain strictly ordered, unique, and nonzero
- producer mathematics and independent-checker semantics are unchanged
- verifier requests use the same canonical request model, preserving exact input binding

## Validation

- probability domain suite: 22 passed
- exact-probability checker/component and capability-contract focused suite: 47 passed
- `git diff --check`: passed
- exact saved Long-trajectory payload: parsed and computed successfully with exact zero moment

The repository planner requests the broad standard lanes because the central probability contract has no narrower ownership. Local `make` orchestration could not create its separate uv environment because this host has no package-network access; hosted CI is expected to run the complete required matrix.

### DIFF
--- a/docs/reference/capabilities/polynomial/gaussian-polynomial-moments.md
+++ b/docs/reference/capabilities/polynomial/gaussian-polynomial-moments.md
@@ -22,25 +22,34 @@ operation, not a Gaussian-identity search or proof workflow.
 
 ## Contract and scope
 
-The request bounds are:
+The request accepts either bounded loose wire terms or the strict canonical
+`GaussianPolynomial` value emitted by parsing those terms. The bounds are:
 
 | Field | Bound |
 | --- | --- |
 | Gaussian variables | 1–8 |
 | Raw sparse terms | 1–16 before canonicalization |
+| Canonical nonzero sparse terms | 1–16 after canonicalization |
 | Total degree of each input term | at most 8 |
 | Fixed moment order \(m\) | 0–16 |
 | Raw ordered expansion paths | `term_count ** m <= 65536` |
-| Input rational numerator/denominator | at most 128 decimal digits |
+| Each raw rational numerator/denominator | at most 128 decimal digits |
+| Each strict canonical rational component | at most 4,096 decimal digits |
 
-Wire terms use exponent vectors of length `variable_count`. They may arrive in
-any order, repeat an exponent vector, or carry an exact zero coefficient. The
-request boundary combines duplicates over \(\mathbb Q(i)\), removes exact zero
-terms, and orders the surviving terms lexicographically before constructing the
-strict `GaussianPolynomial` value used by the producer and checker. A request
-whose terms cancel completely is rejected, as are negative exponents,
-inconsistent dimensions, more than 16 raw terms, and requests beyond the
-complete-expansion bound.
+Loose wire terms use exponent vectors of length `variable_count`. They may arrive
+in any order, repeat an exponent vector, or carry an exact zero coefficient. The
+request boundary checks the raw 16-term and 128-digit limits before combining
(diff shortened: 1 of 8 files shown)


morluto (commented):

**Disposition: retain, but rewrite around the current domain boundary.**

The behavior is correct and follows “parse, don’t validate”: accept bounded unordered/duplicate boundary terms, canonicalize once over `Q(i)`, then expose a strict ordered, unique, nonzero polynomial value downstream. Raw term/count/digit bounds must be enforced before combining.

Rebase onto #1256 and place the canonical value plus normalization function under the authoritative probability/math domain rather than extending another shared wire-type or adapter framework. The producer and checker should consume the same strict domain value without independently re-normalizing loose representations. Update tests to current operation/result paths and remove generic assurance/storage language.

morluto:
Post-simplification review: **keep and rebase; this is strongly aligned with the current parse-to-a-valid-domain-value direction.**

The PR accepts bounded noncanonical Gaussian sparse input, combines duplicate exponents/removes zero terms/orders them at the request boundary, then hands downstream code one canonical `GaussianPolynomial`. That is much better than forcing every consumer to defend against multiple equivalent representations.

On rebase, keep this as one authoritative request parse/canonicalization step. Do not add a parallel canonicalizer elsewhere, and make sure the canonical domain value—not the original wire ordering—owns identity/digests. The oversized/invalid bounds must still fail before expansion/provider work.

Recommendation: rebase onto the current probability contracts, retain the public reproduction regressions, and merge.

morluto:
### Retain, but port at the request boundary on current main

This remains a valid parse-don't-validate fix: mathematically equivalent bounded sparse inputs should canonicalize once before the strict `GaussianPolynomial` value is constructed.

The branch conflicts because the shared probability contracts and result model moved after it was written. The current port should keep the change localized to `GaussianPolynomialMomentRequest` (or a dedicated public parser immediately before it):

- parse at most the existing raw 16-term limit before combining;
- combine duplicate exponent vectors exactly over `Q(i)`;
- remove exact zero coefficients and lexicographically order the surviving terms;
- reject all-zero results and preserve all dimension/degree/coefficient/expansion limits;
- feed the same canonical request into producer and checker binding;
- keep direct `GaussianPolynomial` construction strict and canonical.

Do not introduce a second persistent polynomial representation or loosen the canonical domain value. Update tests to current typed-result paths only.

**Disposition: retain and port; medium priority.**

Issue #1153, opened 2026-08-10
[Architecture]: Publish typed applicability and resource-limit summaries in math.find operation cards


morluto:
## Priority

P1/P2 agent-discovery architecture issue: `math.find` can retrieve the right vocabulary match while still giving the agent too little information to judge mathematical applicability. Critical domain, representation, dimension, and size constraints remain buried in the full input schema or implementation-specific prose.

## Area

`math.find` operation cards, capability descriptors, applicability, mathematical domains, input bounds, execution suitability, contract hydration.

## Summary

A discovery operation card currently exposes useful high-level fields:

```text
capability ID/title/description/modes/tags
accepted input/artifact kinds
input/output schema summaries
scope = exact supplied input or claim
assurance ceiling
provider availability
related capabilities
one bounded invocation example
```

The common schema summary is intentionally compact. It largely reports:

```text
top-level type
required property names
available property names
$ref / oneOf / anyOf counts
```

That is not enough to determine whether an operation applies to the agent's actual mathematical object.

Examples of proof-critical constraints that often remain hidden until CONTRACT inspection or a failed call include:

- coefficient domain `QQ` versus `GF(p)` versus an algebraic/rational-function field;
- fixed three-variable versus dimension-parametric polynomial maps;
- graph order at most 32/64/128;
- finite-field order/extension-degree limits;
- prime versus prime-power assumptions;
- simple graph versus multigraph semantics;
- labelled/ordered basis or tensor convention;
- canonical string versus bounded JSON integer representation;
- maximum polynomial degree/term count/entry magnitude;
- exhaustive versus heuristic/solver-bounded behavior;
- requirement for a producer result URI versus an inline candidate;
- resource/provider/certificate prerequisites.

The card's universal scope statement—“only the exact supplied input or claim is covered”—is a correct anti-overgeneralization rule, but it is not an operation-specific applicability or boundedness summary.

An agent must therefore inspect several candidates or attempt calls simply to learn that they are structurally or mathematically inapplicable.

Audit baseline: current `main` as inspected on 2026-08-10.

### Nearby but inapplicable operations recur across audits

- #964: Jacobian syzygy operations surfaced as weak nearby matches, but their contracts were restricted to `QQ` while the task required characteristic two and a quotient ring.
- #967: exact matrix operations existed over numeric `QQ` matrices but could not represent matrices over `QQ(a,b,...)`.
[...]

Issue #1036, opened 2026-08-10
[Reliability][Deletion]: Add small operation-local bounds for proven aggregate-growth bugs; do not build a cost framework


morluto:
## Summary

Some operations accept fields that are individually bounded but whose combination can still create infeasible work—for example total integer digits, Cartesian products, denominator LCM growth, generated matrix cells, or exact output size.

Fix each proven case with the smallest operation-local preflight before allocation/backend execution. Keep process CPU/memory/time/output limits as a backstop.

Do not introduce `OperationCostEstimate`, execution-envelope models, shared limit registries/contracts, producer/checker compatibility frameworks, generic cost reason taxonomies, or an architecture rule attempting to infer expensive request shapes repository-wide.

### Sequence aggregate magnitude

A sequence may have bounded length while total digits/bits remain infeasible for the selected operation.

### Mutual-information certificate amplification

Individually legal rational denominators may produce a huge LCM/exponent/product.

### Other demonstrated cases

Cartesian enumeration, monomial/basis expansion, dense ledger construction, or checker replay may need a concrete bound when current tests/profiles prove the amplification.

## Target

For each affected operation, write one small pure function over the already parsed request, for example:

```python
def preflight(request: RequestT) -> None:
    if total_digits(request.values) > MAX_TOTAL_DIGITS:
        raise UnsupportedRequest(...)
```

or return the existing concrete unsupported result/diagnostic used by that operation.

Requirements:

- exact/simple arithmetic when possible;
- runs before large allocation, backend call, child process, or permanent write;
- no provider/storage/network/process effects;
- no generic estimate object unless the operation’s current public result genuinely needs those fields;
- bound constant lives beside the concrete operation/checker that uses it;
- checker may enforce its own independent bound when its implementation is slower.

If producer and checker accepted domains differ, state that concrete difference in the checker’s description/request validation. Do not build a compatibility engine.

## Output size

Use #1074’s simple transport rule:

- complete result fits inline -> normal return;
- complete result oversized -> bounded temporary resource.

Do not estimate artifact sizes or preserve durable publication. If computation itself would create an inadmissibly large in-memory result, reject through the operation-local preflight.

## Runtime backstop

Keep existing child-process/wall-clock/memory/output limits where a process boundary is justified. Timeout/resource kill is an operational failure, never a mathematical conclusion.

## Testing

For each concrete preflight add:

- one just-under case;
- one just-over case;
- one adversarial combination that triggered the bug;
[...]

PR #806, opened 2026-08-07
Surface operation validation constraints

morluto:
Fixes #781.

### Problem

A Pydantic model-validation failure was reduced to the domain's generic invalid-request diagnostic. For a stale topology `complex_digest`, the runtime failed closed but did not explain the failing constraint.

### Change

Preserve one bounded, data-minimized validation record when installed computed, materialized, or bounded-search operations reject a Pydantic request. The static domain code and failure semantics remain unchanged; only the diagnostic gains its path when available, a constraint hint, and structural type/path/message data.

The record excludes the rejected input, validation context, and Pydantic URLs. It also caps dynamic text to the existing diagnostic field bounds.

### Regression coverage

Covers a stale topology digest end to end and each operation adapter's Pydantic-validation branch, including a 2,000-character validator message that must remain bounded.

### Validation

- `make test-component TESTS='tests/composition/portfolio/test_operation_installation.py::test_validation_error_preserves_the_rejected_field_and_constraint tests/composition/portfolio/test_operation_installation.py::test_validation_error_respects_diagnostic_text_bounds tests/composition/runtime/test_topology_capabilities.py::test_stale_complex_digest_surfaces_its_constraint'` — 3 passed
- `make test-component TESTS='tests/composition/portfolio/test_operation_installation.py::test_bounded_adapter_preserves_the_rejected_field_and_constraint'` — 1 passed
- `make check-static` — passed

PR #1189, opened 2026-08-11
Clarify polynomial Jacobian expansion limits

yuelgrace1810-ops:
## Problem

`polynomial.map.compute_jacobian` enforces a pre-expansion determinant-product budget of 1,024, but its discovery description only said the operation was bounded. A structurally valid request beyond that aggregate limit then returned only a generic invalid-request message.

This makes a safe resource rejection difficult to understand: an agent cannot tell whether the request was malformed, which aggregate estimate was applied, or how far it was beyond the supported envelope.

## Change

- advertise the 1,024-term Leibniz determinant-product estimate in the capability description, explicitly distinguishing it from simplified output size;
- include the computed estimate and limit in the contract validation error;
- preserve that bounded validation constraint through this bespoke polynomial adapter's public diagnostic;
- keep the request fail-closed and write no artifacts on rejection.

For the focused four-variable regression, the diagnostic now reports:

```text
Jacobian determinant expansion term estimate 3840000 exceeds the exact operation budget of 1024
```

## Overlap

- #1153 and #1036 discuss portfolio-wide applicability and aggregate-cost architecture. This PR does not add a new cost model; it exposes one existing enforced bound on one capability.
- Draft #806 preserves Pydantic constraints for installed domain-operation adapters. `polynomial.map.compute_jacobian` uses the separate bespoke polynomial adapter, so #806 does not cover this runtime path.
- No issue or PR search result contained this exact polynomial-Jacobian diagnostic fix.

## Safety

No input becomes accepted or rejected differently. The mathematical computation, checker, artifact formats, and assurance levels are unchanged. Dynamic diagnostic text is bounded and excludes rejected input values.

## Validation

- focused polynomial contract/runtime tests: 31 passed
- focused discovery/validation tests: 13 passed
- static checks: Ruff, formatting, complexity, and mypy passed
- unit: 891 passed
- component: 808 passed, 3 platform skips
- domain: 379 passed
- composition: 270 passed, 2 unavailable-Lean skips
- storage: 127 passed
- process: 246 passed in the lane; the loopback case passed with local-port permission and the checkout-path false positive passed through a neutral path; one macOS Bash 3.2-only `mapfile` environment failure remains unrelated
- MCP: 48 passed in the lane; the loopback tenant test passed with local-port permission
- end-to-end: 7 passed, 1 unavailable-Lean skip
- source distribution and wheel build passed
- `git diff --check` passed


morluto (commented):

**Disposition: supersede with a consolidated current-main polynomial-budget patch.**

The enforced expansion estimate and bounded, input-redacted diagnostic are still valuable. However, this branch overlaps #1240 on advertising the same Jacobian/product and exponent budgets, while its bespoke adapter path predates the authoritative `OperationSpec`/domain boundary.

I recommend making #1240 (or a fresh replacement) the single owner: keep the operation-owned descriptions and executable collision example, add this PR’s safe computed-estimate/limit diagnostic, and reuse the shared redaction/bounding helper from #1242. Remove references to generic assurance and old adapter-specific diagnostic preservation. Close this branch after those pieces are ported.

PR #1191, opened 2026-08-11
fix(discovery): link polynomial interval verifiers

yuelgrace1810-ops:
## Summary

- publish reciprocal catalog relationships for both exact polynomial-interval producer/verifier pairs:
  - `polynomial.interval.enclose` ↔ `polynomial.interval.enclosure.verify`
  - `polynomial.interval.positivity.decide` ↔ `polynomial.interval.positivity.verify`
- register each relationship only when its authorized checker is installed
- add component, domain, and MCP discovery regressions

## Why

Held-out reliability work exposed two operations whose mathematics and independent checkers already worked, but whose catalog navigation was incomplete. `math.find` returned no related verifier from either producer contract even though each producer and verifier were installed and compatible.

The positivity omission surfaced in a held-out graph-energy replay. A subsequent deterministic portfolio audit found the same localized omission for the neighboring Bernstein-enclosure pair. Keeping both fixes in this draft avoids a competing PR and duplicate edits to the portfolio attachment seam.

This is a discovery-metadata fix. It does not change mathematical execution, request or result schemas, assurance, checker authority, or verification semantics.

## Impact

Agents can navigate directly from either interval producer to its exact independent verifier, and back from the verifier to the accepted producer. The relationships remain absent when checker authority is unavailable.

## Validation

- focused interval/positivity/MCP regressions: 22 passed after rebasing onto the latest PR head
- Ruff and formatting: passed
- strict mypy: 430 source files passed
- unit: 891 passed
- component: 809 passed, 3 expected macOS skips
- domain: 376 passed
- composition: 270 passed, 2 expected Lean skips
- storage: 127 passed
- process: 247 passed, 1 expected platform skip; the sandbox-blocked loopback test passed separately with local permission
- MCP: 47 passed in the parallel lane; the crashed-worker test passed serially and the sandbox-blocked loopback test passed with local permission
- end-to-end: 7 passed, 1 expected Lean skip

One unrelated process test requires Bash 4 `mapfile`; macOS supplies Bash 3.2, so Linux CI remains authoritative for that test. The npm gate is unavailable because this host has no `npm` executable; JavaScript files are untouched.

This PR remains draft and must not be merged by Codex.

### DIFF
--- a/src/jacobian/polynomial_interval_capabilities.py
+++ b/src/jacobian/polynomial_interval_capabilities.py
@@ -40,6 +40,9 @@ from jacobian.checker_operations import CheckerOperation
 from jacobian.contracts.capabilities import (
     CapabilityAssurance,
     CapabilityAssuranceLevel,
+    CapabilityCatalogRelationship,
+    CapabilityCatalogRelationshipKind,
+    CapabilityCatalogRelationshipRegistration,
     CapabilityCompleteness,
     CapabilityCompletenessStatus,
     CapabilityDescriptor,
@@ -115,6 +118,7 @@ class PolynomialIntervalInstallation:
     claim_schema_uri: str
     certificate_schema_uri: str
     checker_id: str | None
+    catalog_relationships: tuple[CapabilityCatalogRelationshipRegistration, ...] = ()
 
 
 @dataclass(frozen=True, slots=True)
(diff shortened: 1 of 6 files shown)


yuelgrace1810-ops:
Independent current-main discovery audit found the same missing relationship presentation in four additional factual producer/verifier pairs:

- `polynomial.expression.normalize` ↔ `polynomial.expression_normalization.verify`
- `polynomial.interval.enclose` ↔ `polynomial.interval.enclosure.verify`
- `polynomial.map.inverse.candidate_synthesize` ↔ `polynomial.map.inverse.verify`
- `polynomial.map.collision.search` ↔ `polynomial.map.collision.verify`

For every endpoint above, the authorized checker is installed, but `math.find(capability_id=..., view="CONTRACT")` currently returns no `related_capabilities`. The producer descriptions already establish the exact relationship: inverse synthesis submits a candidate to the independent two-sided verifier, and collision search emits an unverified witness/checker ID for independent collision replay.

I did **not** include `polynomial.map.compute_jacobian` ↔ `polynomial.map.keller_condition.verify`: the Keller checker consumes the original map and recomputes the determinant, rather than verifying the producer's `PolynomialJacobian` result, so presenting that as a reciprocal producer/result-verifier edge would be misleading.

The expression/interval pairs appear to be the same resource-installation root cause as this PR and are best handled here rather than in a competing patch. The polynomial-map pairs use a separate installer; I am preparing a narrowly scoped, non-overlapping draft fix for those two edges unless you prefer to absorb them here too.

morluto (commented):

**Disposition: close as superseded by the relationship-model deletion.**

The interval producer/checker mathematics remains useful, but this PR contains only reciprocal relationship attachment and tests for that deleted model. #1256 removed those descriptor fields and the registry that owned them.

The semantic-query improvement in #1250 is the correct replacement pattern: make the interval operations independently discoverable through operation-owned wording and examples. Do not rebase this branch or restore relationship metadata conditionally on checker authority.

PR #1194, opened 2026-08-11
fix(discovery): link polynomial map verifiers

yuelgrace1810-ops:
## What changed

- expose reciprocal discovery links between `polynomial.map.collision.search` and `polynomial.map.collision.verify`
- expose reciprocal discovery links between `polynomial.map.inverse.candidate_synthesize` and `polynomial.map.inverse.verify`
- register links only when the corresponding authorized verifier is installed
- add composition regressions for both directions of both relationships

## Why

Current-main `math.find(..., view="CONTRACT")` advertised all four installed operations, but none of their capability cards exposed the factual producer/verifier relationship. Agents therefore had to rediscover the checker by reformulating searches even after finding the useful producer.

This patch does not prescribe a workflow and does not change checker authority or assurance. It only exposes operator-registered factual navigation between already-installed operations.

The adjacent expression/interval relationship gaps were reported on draft PR #1191 because that PR already owns their installation path. The Jacobian/Keller pair is intentionally excluded: the Keller checker consumes the original map rather than verifying the Jacobian producer result.

## Validation

- Ruff: passed
- mypy for changed source files: passed
- focused composition/runtime tests: 22 passed
- unit lane: 891 passed
- `git diff --check`: passed

The full change-aware planner selects broad CI-owned lanes because the polynomial installer has no narrower ownership mapping.


morluto (commented):

**Disposition: close as superseded by the relationship-model deletion.**

This PR is entirely about reciprocal catalog edges. #1256 intentionally removed generic capability relationships and current `math.find` exposes operation-owned facts—schemas, ports, titles, descriptions, tags, examples, and provider availability—rather than a producer/verifier navigation graph.

Do not recreate these edges in another registry or projection. If the collision or inverse verifier remains hard to find, fix that locally through checker-owned names/tags/descriptions and a runnable invocation example, with lexical discovery tests. The current branch has no surviving production change.

Issue #866, opened 2026-08-08
[Bug]: IntegerSequenceRequest has no per-item digit bound


morluto:
## Summary

`IntegerSequenceRequest.values` is bounded to `max_length=256` items, but `CanonicalInteger` has only a pattern constraint with no digit-count limit. Every sequence operation calls `parse_canonical_integer` on each value, so previously guarded oversized operands are now reachable by every sequence operation without a domain-level bound.

## Evidence

`src/jacobian/contracts/sequences.py` L13-19:
```python
class IntegerSequenceRequest(ContractModel):
    """One nonempty finite sequence of canonical integers."""

    values: tuple[CanonicalInteger, ...] = Field(
        min_length=1,
        max_length=_MAX_SEQUENCE_LENGTH,
    )
```

`CanonicalInteger` (`src/jacobian/contracts/exact.py` L13-18) has only `pattern=r"^(?:0|-?[1-9][0-9]*)$"` with no digit-count limit. Other domain contracts (e.g., `CanonicalRational` at exact.py L78-81) enforce `MAX_CANONICAL_RATIONAL_DIGITS`, but `IntegerSequenceRequest` has no equivalent.

## Root cause

The sequence contract bounds item count but not item magnitude.

## Scope

Add a `max_digits` field-level validator to `IntegerSequenceRequest.values` or a cross-field `model_validator` that rejects items exceeding a domain-defined digit bound.

## Source

- https://github.com/morluto/jacobian/pull/807#discussion_r3740003220

Generated with [Devin](https://devin.ai)

morluto:
Confirmed on current `main` and triaged under the aggregate-cost owner **#1036**. The request currently bounds sequence length but not operation-relevant total digits/bits. The fix should be a documented sequence-domain magnitude policy plus aggregate preflight—not necessarily one identical per-item cap for every sequence operation, since sort/frequency/product/recurrence costs differ.

PR #1196, opened 2026-08-11
Add independent LCM replay verification

yuelgrace1810-ops:
## Summary

- install an independent replay checker for `integer.compute.lcm`
- recompute bounded binary LCMs with Python-FLINT and a hand-coded Euclidean recurrence, without importing producer code or calling `math.lcm`
- bind the exact input/result artifact and reject forged schema-valid results
- cover signed and zero boundary cases through the public producer → verifier path

## Why

A saved resolved-conjecture trajectory retained a correct final fraction while displaying a false intermediate LCM (`87787929600` instead of `2520`). The existing binary LCM producer was exact but remained `COMPUTED`; this adds an independent assurance path without introducing theorem-specific logic.

This intentionally does **not** add sequence-wide LCM replay. Aggregate size and integer-magnitude policy for sequence operations remains the separate #866/#1036 boundary.

## Independence and safety

The producer uses Python's standard-library `math.lcm`. The checker uses Python-FLINT integers and an independently implemented Euclidean recurrence. It enforces the existing bounded binary request contract, bounded result size, exact source/result fields and digests, nonnegative LCM semantics, and fail-closed rejection.

## Validation

- lint and full type checking: passed (470 source files)
- unit: 891 passed
- component: 818 passed, 3 platform skips
- domain: 383 passed
- composition: 270 passed, 2 pinned-Lean runtime skips
- storage: 127 passed
- MCP: 49 passed
- E2E: 7 passed, 1 pinned-Lean runtime skip
- process: 248 passed, 1 platform skip; one unchanged macOS environment failure because Apple Bash 3.2 lacks `mapfile`, already used by `scripts/setup-agent` on the base commit

No benchmark mathematics or unrelated capability was changed.

### DIFF
--- a/src/jacobian/domains/topology/operations.py
+++ b/src/jacobian/domains/topology/operations.py
@@ -505,6 +505,21 @@ _CIRCLE = {
     "facets": [["a", "b"], ["b", "c"], ["a", "c"]],
 }
 
+_CANONICAL_CIRCLE = {
+    "vertices": ["a", "b", "c"],
+    "maximal_simplices": [["a", "b"], ["a", "c"], ["b", "c"]],
+    "faces_by_dimension": [
+        {"dimension": 0, "faces": [["a"], ["b"], ["c"]]},
+        {"dimension": 1, "faces": [["a", "b"], ["a", "c"], ["b", "c"]]},
+    ],
+    "dimension": 1,
+    "f_vector": [3, 3],
+    "closure_size": 6,
+    "complex_digest": (
+        "sha256:6f797991bac967e2a8e572707df487061655df0f094cbde0f52f82c5401fc043"
+    ),
+}
+
 type TopologyOperation = (
     InstalledOperation[
         SimplicialComplexRequest, SimplicialComplexCanonicalizationResult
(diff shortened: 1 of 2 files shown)


morluto (commented):

**Disposition: revise; do not merge the producer-coupled replay shape as written.**

A public checker for binary LCM only clears the catalog budget if it can verify an externally supplied claim. Reframe it as a small candidate verifier over typed `{left, right, candidate}` values, independent of producer artifacts, source/result digests, or reciprocal navigation. If its only use is cross-checking Jacobian’s own `math.lcm` result, move the FLINT/Euclidean comparison into differential tests and close the public-operation PR.

If retained, install it as a normal checker `OperationSpec`, make the accepted producer/candidate semantics declaration-owned rather than a central string dispatcher, use the current verification record, and keep signed/zero/forged/boundary regressions.

morluto:
Post-#1265 review: **keep/rebase the missing LCM verifier; drop unrelated stacked changes.** The verifier is still absent on current main, so the functional gap remains.

This is a good fit for the simplified verification model: `integer.compute.lcm` stays an ordinary producer and `integer.lcm.verify` is a separate operator-authorized replay operation. No producer mode switch, relationship graph, or generic assurance wrapper is needed.

The branch also touches discrete-log files that are unrelated to the stated verifier fix. Restack from current main with only the number-theory checker declaration/entrypoint/tests needed for LCM, and rely on normal lexical tags/examples for discovery.

Recommendation: revise/rebase narrowly and merge the checker slice.

morluto:
Superseded by #1280. The replacement is rebuilt directly on post-#1276 `main`, uses a dedicated standard-library Euclidean checker runtime, and keeps only the bounded producer→verifier slice. The unrelated discrete-log startup-budget edit is not included; no reciprocal relationships, generic assurance fields, or shared checker-hub expansion were retained.

PR #1197, opened 2026-08-11
fix(discovery): link polynomial-system verifier

yuelgrace1810-ops:
## What changed

- expose `polynomial.system.solution.verify` from bounded rational solution search as its independent verifier
- expose the reciprocal search producer from the verifier
- register both edges only when the bundled checker is operator-authorized and installed
- add a whole-runtime composition regression for both relationship kinds and checker authority

## Why

Current main installs `polynomial.system.rational_solution.search` and `polynomial.system.solution.verify` as a natural producer/verifier pair. The search returns an exact assignment candidate, and the verifier independently checks that assignment against every equation and inequation. Exact `math.find(..., view="CONTRACT")` inspection nevertheless returned no related capabilities for either endpoint, forcing agents to rediscover the verification boundary separately.

## Impact

This is discovery metadata only. It does not change polynomial search, evaluation, schemas, resource bounds, checker authorization, or assurance. Search remains computed and bounded; only the existing authorized verifier can return verified assurance.

## Validation

- user-facing `math.find` replay shows reciprocal `INDEPENDENT_VERIFIER` / `VERIFIABLE_RESULT_PRODUCER` links
- focused producer, checker, catalog, and whole-runtime tests: 14 passed
- full unit lane: 891 passed
- `git diff --check`: passed

The standard `make test-unit` wrapper attempted dependency synchronization and could not reach PyPI in the sandbox, so the same unit topology was run directly with the existing locked environment.

### DIFF
--- a/src/jacobian/portfolio/core_installation.py
+++ b/src/jacobian/portfolio/core_installation.py
@@ -6,6 +6,10 @@ from dataclasses import dataclass
 
 from jacobian.atomic_capabilities import install_atomic_capabilities
 from jacobian.conjecture_ingestion import ConjectureIngestionInstallation
+from jacobian.contracts.capabilities import (
+    CapabilityCatalogRelationship,
+    CapabilityCatalogRelationshipKind,
+)
 from jacobian.exact_domain_checkers import install_exact_domain_verification
 from jacobian.finite_coverage import install_finite_coverage
 from jacobian.finite_partition import install_finite_partition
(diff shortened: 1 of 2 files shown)


morluto (commented):

**Disposition: close as superseded by #1256.**

The mathematical producer and checker may still be valid operations, but this PR changes only the deleted relationship representation. Current descriptors intentionally do not encode reciprocal producer/verifier edges.

If `polynomial.system.solution.verify` is still difficult to discover, improve the verifier’s own title, description, tags, and operation-owned invocation example, then test the natural query directly. Do not restore a relationship registry, authority-conditioned edge attachment, or MCP projection for this purpose.

PR #1158, opened 2026-08-10
refactor(mcp): remove dual-mode tools and mode from MCP agent surface (#1143, #1148)

morluto:
## Summary

Implements PR1–PR4 of the [#1148](https://github.com/morluto/jacobian/issues/1148) product surface cleanup so the runtime matches the documented product:

```
math.find  → discover/inspect math tools
math.run   → run one tool → mathematical value (or checker verdict)
```

- **No dual-mode tools**; checkers are separate catalog IDs.
- Retire agent-facing `mode` ([#1143](https://github.com/morluto/jacobian/issues/1143)).
- **Math-first** results for agents; demote assurance/mode theater.

### Changes

**PR1 — Forbid dual-mode + split `case.partition.finite` (#1143 1a–1b)**
- `CapabilityDescriptor` validation rejects `len(modes) > 1` at install time (dual-mode ban).
- Split finite partition into a producer (`case.partition.finite`, materializes cases) and a separate checker (`case.partition.finite.verify`, replays coverage/disjointness when checker authorized).
- Updated `core_installation` wiring and catalog relationships.
- Unit test that dual-mode registration fails closed.

**PR2 — Default mode from tool role (#1143 1c)**
- `CapabilityRequest.mode` is now `None`-defaulted; dispatch stamps the sole advertised mode from the descriptor.
- Removed `UNSUPPORTED_MODE` resolution failure path — client mode is ignored, tool identity owns role.
- VERIFY-only tools now work without clients sending `mode=VERIFY`.

**PR3 — Remove mode from MCP agent surface (#1143 1d–1e)**
- Removed `mode` parameter from `capability_describe` and `capability_invoke` (SDK regenerates schemas).
- Removed `mode` filter from discovery request/result contracts and `capability_discovery.py`.
- Removed `mode` from projection cards, invocation examples, cursor hint, recovery path text.
- Removed `--allow-mode` / `--deny-mode` CLI flags and `allowed_modes`/`denied_modes` policy wiring.
- Removed `mode` from `_build_capability_description` in eval telemetry.
- Rewrote guidance: search/execute, producers vs checkers, no mode workflow; `evidence_check_prompt` no longer references `mode="VERIFY"`.

**PR4 — Math-first agent projection**
- `_run_text_projection` now leads with `output` (the value) and execution status; demotes `assurance`/`mode` and omits empty semantic fields from the text projection.
- Full structured `CapabilityResult` preserved via SDK structured content.

### Out of scope (separate PRs per plan)
[...]

PR #1199, opened 2026-08-11
fix(discovery): restore finite-partition verifier link

yuelgrace1810-ops:
## What changed

- restore the discovery link from `case.partition.finite` to its independent verifier
- restore the reciprocal producer link from `case.partition.finite.verify`
- register both edges only when the finite-partition checker is operator-authorized and installed
- add a whole-runtime composition regression for both relationship kinds and checker authority

## Why

PR #1158 split finite partitioning into separate producer and checker IDs and stated that catalog relationships were updated. On current main, however, exact `math.find(..., view="CONTRACT")` inspection returns no related capabilities for either endpoint. Agents can therefore find the producer yet miss the checker that independently replays its coverage/disjointness certificate.

## Impact

This restores factual navigation only. It does not change finite-partition semantics, schemas, equality-only scope, checker authorization, execution, or assurance. The producer remains computed; only the existing checker can return verified assurance.

## Validation

- user-facing `math.find` replay shows reciprocal `INDEPENDENT_VERIFIER` / `VERIFIABLE_RESULT_PRODUCER` links
- focused finite-partition, checker, authority, and whole-runtime tests passed
- full unit lane: 891 passed
- `git diff --check`: passed

### DIFF
--- a/src/jacobian/portfolio/core_installation.py
+++ b/src/jacobian/portfolio/core_installation.py
@@ -6,6 +6,10 @@ from dataclasses import dataclass
 
 from jacobian.atomic_capabilities import install_atomic_capabilities
 from jacobian.conjecture_ingestion import ConjectureIngestionInstallation
+from jacobian.contracts.capabilities import (
+    CapabilityCatalogRelationship,
+    CapabilityCatalogRelationshipKind,
+)
 from jacobian.exact_domain_checkers import install_exact_domain_verification
 from jacobian.finite_coverage import install_finite_coverage
 from jacobian.finite_partition import install_finite_partition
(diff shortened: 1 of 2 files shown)


morluto (commented):

This branch restores only reciprocal catalog relationships, a representation deliberately removed in #1256. The finite-partition producer and checker should remain separate atomic operations; checker authorization is installation state, not a reason to synthesize navigation edges into their descriptors.

If discoverability is still weak, fix the checker declaration itself—plain-language title/description/tags plus a runnable request example—and assert direct lexical search. Do not port the relationship tests or whole-runtime edge registry.

PR #1200, opened 2026-08-11
Fix discrete-log example startup budget

yuelgrace1810-ops:
## What changed

- raise the advertised `modular.compute.discrete_logarithm` example budget from 1 second to 5 seconds
- add a regression test that freezes enough time for isolated-worker startup

## Why

The example launches a bounded worker process. On current `main`, its one-second budget can expire during worker startup before the trivial discrete-logarithm calculation runs. This reproduced locally and caused the domain lane on draft PR #1196 to fail with `DISCRETE_LOGARITHM_TIMEOUT`.

The change affects only the advertised example input. It does not change the operation's mathematics, global request bounds, isolation model, or user-supplied budgets.

## Validation

- focused advertised-example and protocol tests: 4 passed
- Ruff check and format: passed
- mypy: 470 source files passed
- unit lane: 892 passed
- component lane: 945 passed, 3 platform skips; two unrelated parallel workers crashed, and both affected tests passed serially (2 passed)
- domain lane: the changed discrete-log example passed; five unrelated factorization examples timed out only under four-worker load and all five passed serially
- diff check: passed

This is intentionally a draft and should not be merged by Codex.

### DIFF
--- a/src/jacobian/domains/number_theory/discrete_logarithm.py
+++ b/src/jacobian/domains/number_theory/discrete_logarithm.py
@@ -168,7 +168,7 @@ DISCRETE_LOGARITHM_CAPABILITY = BoundedSearchOperation(
                 "base": 2,
                 "target": 1,
                 "modulus": 3,
-                "resource_budget": {"wall_seconds": 1},
+                "resource_budget": {"wall_seconds": 5},
             },
         ),
     ),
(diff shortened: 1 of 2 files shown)


morluto (commented):

**Disposition: keep the fix, supersede this branch into one consolidated example-budget PR.**

The one-second advertised example is genuinely invalid because it budgets mathematical work but not ordinary isolated-worker startup. This is the same defect class as #1204, #1205, and the budget half of #1239.

Rather than maintain four branches and four one-off tests, create one current-main patch that updates all affected operation-owned examples and adds a parameterized invariant: every published example must execute from a cold worker within its declared bounded budget with a reasonable startup margin. Preserve user-supplied/global limits unchanged and remove old assurance terminology.

PR #1201, opened 2026-08-11
Add independent rational arithmetic replay

yuelgrace1810-ops:
## Summary

- add independent Python-FLINT replay checkers for bounded rational sum, difference, product, and quotient
- expose reciprocal producer → verifier navigation through the existing exact-domain installation path
- bind exact inputs and candidate artifacts and reject schema-valid forged results
- preserve `COMPUTED` producer assurance and grant `VERIFIED` only after authorized independent replay

## Why

Resolved-conjecture trajectories across analysis, probability, and number theory repeatedly retained correct final conclusions while displaying incorrect exact intermediate fractions, products, or scaled values. The existing rational producers use standard-library `Fraction` and were exact, but had no independent replay path.

This change adds a reusable trust boundary for the four core binary rational operations. It does not add theorem-specific logic, sequence-wide arithmetic, ordering predicates, or approximate numerics.

## Independence and safety

The producers use standard-library `Fraction`. The checker is a separate entrypoint using Python-FLINT `fmpq`, imports neither producer code nor Jacobian runtime code, requires canonical reduced rationals with positive denominators, enforces 32,768-digit component bounds, binds source/result digests, and fails closed on malformed, cross-operation, or division-by-zero requests.

## Validation

- focused checker and public producer→verifier tests: 23 passed
- changed-file Ruff and formatting: passed
- targeted source mypy: passed
- full static validation: passed (472 source files in the validated checkout)
- unit: 891 passed
- component: 955 passed, 3 platform skips; one unrelated parallel worker crash passed on isolated rerun
- MCP: 47 passed apart from a sandbox-only loopback-bind denial
- composition: 139 passed, 2 pinned-Lean skips
- storage: 126 passed
- E2E: 7 passed, 1 pinned-Lean skip
- full domain validation was blocked only by the reproducible current-main advertised discrete-log startup timeout now isolated in draft PR #1200
- process-only local limitations: sandbox loopback denial and macOS Bash 3.2 lacking `mapfile`

This PR is intentionally draft and must not be merged by Codex.

### DIFF
--- a/src/jacobian/domains/arithmetic/bundle.py
+++ b/src/jacobian/domains/arithmetic/bundle.py
@@ -12,6 +12,7 @@ from __future__ import annotations
 import platform
 
 from jacobian.contracts.capabilities import CapabilityDiagnostic
+from jacobian.domains.arithmetic.checkers import ARITHMETIC_EXACT_REPLAY_CHECKERS
 from jacobian.domains.arithmetic.integers import INTEGER_CAPABILITIES
 from jacobian.domains.arithmetic.rationals import RATIONAL_CAPABILITIES
 from jacobian.operations import (
@@ -39,7 +40,10 @@ def build_arithmetic_bundle() -> DomainBundle:
                 "integer_encoding": "canonical decimal string",
                 "rational_encoding": "canonical reduced num/den with positive denominator",
                 "arithmetic": "exact via stdlib and maintained SymPy APIs",
-                "assurance": "computed; no independent checker",
+                "assurance": (
+                    "producers are computed; core binary rational arithmetic "
+                    "supports operator-authorized independent replay"
+                ),
             },
         ),
         provider_runtime=known_provider_runtime(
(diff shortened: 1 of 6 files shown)


morluto (commented):

**Disposition: supersede as a public checker bundle; keep the independent arithmetic as tests.**

Four producer-coupled replay checkers for rational `+`, `-`, `×`, and `÷` duplicate trivial canonical domain arithmetic and expand the public catalog without adding a nontrivial certificate. Typed rational values already make reduced form/positive denominator structural, and independent FLINT comparisons are valuable as differential/property tests.

I would not port the relationship wiring, source/result digests, or four public checker operations. Move the FLINT replay and boundary/forgery cases into tests of the authoritative rational operations. Reopen a public candidate-verification operation only if a concrete benchmark requires checking externally supplied arithmetic claims and can justify a narrower, non-stringly contract.

yuelgrace1810-ops:
Post-refactor re-audit confirms this public checker bundle should not be ported. Typed rational values already make reduced form and positive denominators structural, while four producer-coupled replay operations would duplicate trivial canonical arithmetic and expand the public catalog without a nontrivial certificate boundary. The FLINT comparisons remain useful as differential/property-test ideas, but not as these public operations, so I am closing this PR rather than reintroducing pre-cutover checker plumbing.

PR #1202, opened 2026-08-11
test(polynomials): align polynomial-map semantics with the public dimension limit

yuelgrace1810-ops:
## Problem

The authoritative `RationalPolynomialMap` contract accepts up to `MAX_POLYNOMIAL_VARIABLES` variables (currently 8), but the installed `jacobian.rational-polynomial-map` semantics descriptor reports `maximum_dimension: 4`.

That makes durable semantics and discovery metadata narrower than the public request schema. Consumers can incorrectly reject supported 5–8 variable maps or report the wrong contract boundary.

## Change

- derive the semantics descriptor's `maximum_dimension` from `MAX_POLYNOMIAL_VARIABLES`;
- add a component regression test that requires the installed semantics and public schema limit to remain aligned.

This changes metadata only. It does not expand accepted inputs, alter polynomial computation, change resource budgets, or change checker assurance.

## Evidence and overlap

The mismatch surfaced while source-gating an explicit 14-dimensional polynomial-map certificate. That request remains unsupported by the current 8-variable schema; this PR does not attempt to support it.

I searched open/closed issues and all PR states for the exact 4-versus-8 semantics mismatch and found no owner. Draft #1189 separately clarifies the determinant-product estimate and diagnostic; it does not change dimension metadata. #1036 owns broader aggregate-cost architecture.

## Validation

- focused regression: 1 passed
- Ruff formatting and lint: passed
- `git diff --check`: passed

A broader file run produced 7 passes and 2 `SOURCE_CHANGED` failures because the isolated worktree reused another checkout's installed checker environment; those two tests intentionally reject a checker whose source digest points at a different checkout. The changed metadata path is not involved.


morluto (commented):

**Disposition: retain the defect, revise the solution on current `main`.**

The 4-versus-8 mismatch still exists, but the deeper problem is duplicate ownership: a generic semantics descriptor repeats a limit already owned by the rational-polynomial-map request/value contract. After #1256, domain values and operation requests should be authoritative.

First decide whether `jacobian.rational-polynomial-map` is still required for durable compatibility. If not, delete the duplicate `maximum_dimension` field/descriptor path. If it is required, derive it directly from the domain-owned bound and add a current regression preventing divergence. Do not introduce another synchronization constant or preserve generic assurance metadata.

morluto:
Post-#1265 review: **keep and rebase.** Correcting the polynomial-map semantics bound is an ownership/contract fix, not something invalidated by the tool simplification.

The main requirement on rebase is that the maximum supported dimension be owned once by the current polynomial-map semantic/request type and reflected mechanically in inspection/schema/provider admission. Avoid duplicating the same numeric limit in installation metadata, prose, and runtime checks if the typed owner can supply it.

Recommendation: restack onto the current polynomial operation installation, keep the regression proving advertised semantics and executable bounds agree, and merge. No new generic semantics registry is needed.

morluto:
### Current-main port required; one-line invariant remains correct

I tested the branch against the current refresh base and confirmed that it conflicts with the post-cutover polynomial installation path, so I did not replace current files wholesale.

The surviving change is still exactly the right invariant:

```python
"maximum_dimension": MAX_POLYNOMIAL_VARIABLES
```

in the current `jacobian.rational-polynomial-map` semantics registration, with a regression comparing the stored descriptor to the authoritative public constant.

Port notes:

- import `MAX_POLYNOMIAL_VARIABLES` into the current `polynomials/installation.py` without restoring old adapter tuple/result types;
- change only the map semantics descriptor, not the separate sparse-polynomial/rational-function descriptors unless their contracts are independently broadened;
- place the regression beside current installation/component ownership and use the current runtime fixture;
- do not change accepted inputs or resource budgets.

**Disposition: retain and port as a two-file metadata fix; high confidence.**

morluto:
@codex address the two unresolved P1 threads together, not independently. Make the polynomial-map boundary coherent at the public eight-variable limit:

- widen `PolynomialJacobian.variable_order` and matrix row/column bounds to `MAX_POLYNOMIAL_VARIABLES`;
- keep the generic polynomial-identity checker at its existing four-variable/127-derived-exponent scope;
- give map parsing/replay its own eight-variable bound and derive its maximum replay exponent from `MAX_POLYNOMIAL_VARIABLES * _MAX_JACOBIAN_SOURCE_EXPONENT - 1` rather than broadening the identity checker accidentally;
- update the installed map semantics' `maximum_derived_exponent` from 127 to the same type-owned map bound;
- add direct 8D Jacobian and Keller/collision replay regressions that construct the typed `PolynomialJacobian`, not only a raw collision payload;
- retain the existing determinant product-term preflight and checker intermediate-term cap.

Please push the focused fix to this PR branch and rerun the owning contract/checker/provider tests.

PR #1203, opened 2026-08-11
fix(matrices): repair one-row lattice reduction

yuelgrace1810-ops:
## Summary

- handle valid one-row integer bases inside the isolated LLL worker
- return the unchanged row basis with the exact 1×1 identity transformation
- cover both the advertised unit-basis example and a rectangular one-row basis

## Root cause

`lattice.basis.reduce` accepts and advertises one-row bases, but Python-FLINT 0.9.0 / FLINT 3.6.0 rejects that degenerate LLL call. Every one-row basis is already LLL-reduced, so asking the backend to reduce it turned a valid public request into `FLINT_LLL_WORKER_FAILED`.

The fix remains inside the pinned isolated worker. It does not change the public contract, backend identity checks, materialization, relation binding, or COMPUTED assurance.

## Validation

- focused matrix-lattice file: 16 passed
- unit lane: 891 passed
- component lane: 947 passed, 3 platform skips
- lint, format, complexity, and mypy: passed
- package build: passed
- full domain lane: 459 passed; one unrelated failure in `modular.compute.discrete_logarithm` because its advertised one-second startup budget timed out, already owned by draft PR #1200
- diff check: passed

### DIFF
--- a/src/jacobian/domains/matrix_lattice/lll_worker.py
+++ b/src/jacobian/domains/matrix_lattice/lll_worker.py
@@ -77,13 +77,20 @@ def _run(worker_request: LllWorkerRequest) -> LllWorkerResponse:
             for row in worker_request.request.basis.entries
         ]
         source = flint.fmpz_mat(entries)
-        reduced, transformation = source.lll(
-            transform=True,
-            delta=0.99,
-            eta=0.51,
-            rep="zbasis",
-            gram="exact",
-        )
+        if source.nrows() == 1:
+            # Every one-row integer basis is already LLL-reduced.  FLINT rejects
+            # this mathematically valid boundary case, so preserve it exactly
+            # with the unique one-dimensional identity transformation.
+            reduced = source
+            transformation = flint.fmpz_mat([[1]])
+        else:
+            reduced, transformation = source.lll(
+                transform=True,
+                delta=0.99,
+                eta=0.51,
+                rep="zbasis",
+                gram="exact",
+            )
         if transformation * source != reduced:
             raise WorkerError(LllWorkerErrorCode.RELATION_INVALID)
     except WorkerError:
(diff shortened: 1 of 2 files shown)


morluto (commented):

**Disposition: retain and rebase as a small backend-totalization fix.**

A one-row integer basis is mathematically already reduced, while the pinned FLINT call is partial on that valid public input. Handling the case inside the isolated backend worker is the right ownership boundary and does not require a new public mode or abstraction.

Rebase onto current `main`, keep the exact unchanged basis plus 1×1 identity transformation, and update tests/result assertions to current domain-owned values. Remove stale relationship and generic `COMPUTED` wording from the PR, but otherwise keep this patch narrow.

morluto:
Post-#1265 review: **keep and rebase; this remains a clean backend-boundary fix.**

Current main still calls `fmpz_mat.lll(...)` unconditionally. The one-row mathematical case is valid but the provider rejects it, so preserving the source with the 1×1 identity transformation before entering FLINT is an appropriate thin-adapter compatibility fix. It does not change the semantic value or weaken the relation check (`transformation * source == reduced`).

This also fits the new backend rule well: Jacobian owns the promised mathematical semantics and adapts a backend edge case rather than exposing FLINT's narrower precondition as the public domain.

Recommendation: rebase, keep the focused one-row + existing multi-row/provider-failure regressions, and merge. No broader LLL abstraction is needed.

morluto (commented):

This branch was refreshed onto the post-#1265 architecture and has landed as `b5826b6`.

The final patch stays at the correct boundary: it totalizes the pinned FLINT adapter for valid one-row integer bases, preserves the source basis with the exact 1×1 identity transformation, and retains focused one-row/multi-row regressions. It introduces no broader LLL abstraction, public mode, or stale result/relationship machinery.

PR #1204, opened 2026-08-11
Fix Gröbner example startup budget

yuelgrace1810-ops:
## Summary

- raise the advertised Gröbner `unit_ideal` example budget from 1 to 10 seconds
- add a regression that executes the catalog example through the public capability

## Root cause

The public example deterministically timed out at about 1.04 seconds on both cold and warm runs because its one-second process budget did not include ordinary isolated-worker startup. With the normal 10-second default, the exact unit-ideal computation completes in about 1.6–1.8 seconds.

This changes only the example's declared budget. It does not change Gröbner mathematics, runtime limits for user requests, completeness, artifacts, or assurance.

## Validation

- focused Gröbner tests: 3 passed
- lint, format, complexity, and mypy: passed
- offline package build: passed
- diff check: passed
- Benchmarks workflow: passed
- Container workflow: passed

## Known base-branch CI dependency

The domain CI lane currently fails only on the unchanged `modular.compute.discrete_logarithm` catalog example, whose one-second startup budget is a current-`main` defect already fixed in draft PR #1200. The changed Gröbner example passes. This PR intentionally does not duplicate or cherry-pick #1200's unrelated fix; its CI should be rerun after #1200 lands.

### DIFF
--- a/src/jacobian/domains/number_theory/discrete_logarithm.py
+++ b/src/jacobian/domains/number_theory/discrete_logarithm.py
@@ -168,7 +168,7 @@ DISCRETE_LOGARITHM_CAPABILITY = BoundedSearchOperation(
                 "base": 2,
                 "target": 1,
                 "modulus": 3,
-                "resource_budget": {"wall_seconds": 1},
+                "resource_budget": {"wall_seconds": 5},
             },
         ),
     ),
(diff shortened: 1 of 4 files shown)


morluto (commented):

**Disposition: keep the fix, consolidate with #1200/#1205/#1239.**

The advertised Gröbner example must not deterministically expire during normal worker startup, and the public-execution regression is valuable. The implementation is otherwise orthogonal to the architecture cutover.

Do not carry this as a separate stale branch. Port the 10-second example change into one post-#1256 example-budget PR with a parameterized cold-start execution invariant covering all affected examples. Update paths/tests to current operation declarations and drop references to deleted completeness/assurance fields.

PR #1205, opened 2026-08-11
Fix published Arb and Gröbner cold-worker examples

yuelgrace1810-ops:
## Summary

Ports the surviving cold-worker example fixes onto the post-#1276 typed ownership model.

- raise the published Gröbner-basis example budget from one second to ten seconds;
- raise the published Arb point-enclosure example budget from one second to ten seconds;
- name every supported Arb function in the operation description and tags so square-root, logarithm, exponential, sine, and cosine queries discover it;
- execute both operation-owned examples through the current typed dispatch path;
- keep one Arb budget owner here and leave #1239 superseded.

## Architecture boundary

This branch was rebuilt as one commit directly on current `main` (`195c4c1`). It changes only two `OperationSpec` declarations and three focused tests. It does not carry forward the old inverse-verifier refactor, generic result construction, reciprocal relationships, workflow navigation, or an alternate MCP surface.

The inverse-synthesis example was deliberately removed from this refresh: #1276 changed that publisher/checker path, and preserving its old branch-local refactor would mix an architectural migration into a metadata fix. Any remaining inverse startup defect should be handled against its current owner in a separate narrow patch.

## Validation

CI is authoritative for the rebuilt head. The focused tests assert that the advertised examples complete and that every supported Arb function remains discoverable.

Supersedes the surviving Arb portion of #1239.

### DIFF
--- a/src/jacobian/domains/analysis/operations.py
+++ b/src/jacobian/domains/analysis/operations.py
@@ -139,12 +139,29 @@ POINT_ENCLOSURE_CAPABILITIES = (
             title="Enclose a real function at a rational point",
             description=(
                 "Use pinned Arb ball arithmetic to enclose one supported real "
-                "function at one exact rational point within a wall-clock budget."
+                "function (square root, logarithm, exponential, sine, or cosine) "
+                "at one exact rational point within a wall-clock budget."
             ),
             request_type=ArbPointEnclosureRequest,
             result_type=ArbPointEnclosureResult,
             execute=_point_enclosure,
-            tags=("analysis", "validated", "arb", "enclosure", "bounded"),
+            tags=(
+                "analysis",
+                "validated",
+                "arb",
+                "enclosure",
+                "bounded",
+                "square-root",
+                "sqrt",
+                "logarithm",
+                "log",
+                "exponential",
+                "exp",
+                "sine",
+                "sin",
+                "cosine",
+                "cos",
+            ),
             invocation_examples=(
                 example(
                     "sqrt_zero",
@@ -153,7 +170,7 @@ POINT_ENCLOSURE_CAPABILITIES = (
                         "function": "SQRT",
                         "argument": {"num": "0", "den": "1"},
                         "precision_bits": 32,
-                        "wall_seconds": 1,
+                        "wall_seconds": 10,
                     },
                 ),
             ),
--- a/src/jacobian/domains/polynomial/groebner.py
+++ b/src/jacobian/domains/polynomial/groebner.py
@@ -149,7 +149,7 @@ POLYNOMIAL_GROEBNER_CAPABILITY = durable_operation(
                         }
                     ],
                     "monomial_order": "lex",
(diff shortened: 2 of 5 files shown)


morluto (commented):

**Disposition: keep the correction, consolidate it into the shared example-budget patch.**

The current example is not runnable under its own advertised timeout, so the 30-second bound and end-to-end example regression are justified. This is not an inverse-synthesis architecture change; it is publication metadata validation.

Port it with #1200/#1204 and the budget portion of #1239 onto current `main`. The shared test should execute every published example from a cold process and fail if startup overhead alone consumes the declared budget. Do not change general request limits or add per-capability timeout infrastructure.

morluto:
Post-#1256/#1265 review: **keep the underlying cold-worker/resource fixes, but revise/rebase and de-duplicate the Arb slice with #1239.**

The principle remains correct: predictable setup/import cost belongs in preflight/admission, not in the first bounded operation call. That is compatible with the new `OperationSpec` pipeline as long as each operation owns a concrete preflight bound and provider startup remains outside mathematical execution.

This branch should now be narrowed to the surviving Groebner/inverse worker fixes plus any Arb fix not already covered by #1239. #1239 also changes the Arb point-enclosure budget/discovery path; choose one owner for that hunk rather than merging two variants.

On rebase, express the limits through current typed request/preflight/provider owners, not the older `ComputedOperation`/generic budget scaffolding, and keep the regression that the limit is enforced before starting the worker/provider.

Recommendation: revise/rebase, split out duplicate Arb work, then merge the remaining provider-boundary fixes.

morluto:
### Make this the single current-main owner for cold-worker example metadata

I tested the branch against the current refresh base; it has real conflicts, so it needs a deliberate six-file port rather than a merge of stale files.

Please absorb the two surviving additions from #1239 while porting:

1. add factual Arb discovery vocabulary for `SQRT`, `LOG`, `EXP`, `SIN`, and `COS` to `analysis.real_function.point_enclosure.compute`;
2. raise the advertised discrete-log example budget from 1 to 5 seconds.

Then keep this PR's existing current-scope changes:

- Arb example: 10 seconds;
- Gröbner example: 10 seconds;
- inverse-synthesis example: 30 seconds;
- operation-owned example execution regressions.

Port tests to current typed result paths and avoid literal generic assurance/completeness assertions. User-supplied budgets, global limits, worker isolation, and mathematical semantics must remain unchanged.

**Disposition: retain as the consolidation owner; revise on current main. #1239 is superseded.**

morluto:
Superseded by #1283. The replacement is rebuilt directly on post-#1276 `main` and contains only the three published cold-worker budget changes, Arb function discovery metadata, and focused executable-example regressions. The old inverse-adapter compatibility migration is not retained.

morluto:
Superseded by #1283, which ports the surviving inverse/Gröbner/Arb cold-worker budgets and the unique Arb discovery vocabulary onto current main without carrying the duplicate or stale branch structure.

PR #1206, opened 2026-08-11
fix(skills): align math.run payload envelope

N0zoM1z0:
## Problem

The managed Jacobian Codex skill advertises stable direct `math.run` calls as `{"capability_id":"<id>","input":...}`. The MCP 2.0 tool schema and `math.find(..., view="CONTRACT")` invocation examples require `payload`, so following the skill fails with `INVALID_INPUT` before the mathematical operation runs.

## Change

- restore `payload` in both repository and packaged skill copies while keeping the mode-less wire contract
- bind the visibility regression to the real `capability_invoke` signature and explicitly reject the stale `input` envelope
- preserve byte identity between both packaged copies and keep the skill below 4 KiB

This changes only agent guidance and its regression test; the MCP server wire contract is unchanged.

## Validation

- skill creator `quick_validate.py` on both skill copies
- `make check-changed PATHS='...'`: 4/4 passed (lint/typecheck, npm, docs, 891 unit tests)
- `make test-mcp TESTS=tests/boundary/mcp/test_mcp_adapter_inventory.py`: 1 passed
- `make npm-test`: 59 passed plus successful package dry-run
- live reproduction on MCP `0.11.0`: `input` rejected with `INVALID_INPUT`; `payload` completed and returned `gcd(84,30)=6`

### DIFF
--- a/.agents/skills/jacobian-math/SKILL.md
+++ b/.agents/skills/jacobian-math/SKILL.md
@@ -22,7 +22,7 @@ Keep decomposition and routing decisions agent-owned; composing already-known
 supporting operations remains allowed when clearer.
 Do not call Jacobian for definitions, formatting, or non-execution tasks.
 
-No discovery for stable producers: `{"capability_id":"<id>","input":<JSON>}`.
+No discovery for stable producers: `{"capability_id":"<id>","payload":<JSON>}`.
 Payloads:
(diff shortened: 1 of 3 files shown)


N0zoM1z0:
LGTM!

PR #1232, opened 2026-08-11
fix(skills): correct determinant verifier payload

yuelgrace1810-ops:
## What changed

- replace the stale `determinant_uri` example for `matrix.determinant.verify` with the live inline `{input, candidate}` envelope
- keep the repository and packaged Jacobian skill copies identical
- validate a concrete determinant verifier payload against the runtime's generic exact-replay request model

## Why

PR #1206 corrected the outer `math.run` argument from `input` to `payload`, but the stable determinant verifier example still described the retired artifact-URI contract. Agents copying the skill would therefore submit a schema-invalid verification request even though the producer and checker are installed and linked.

This changes guidance and regression coverage only. It does not alter determinant mathematics, checker authority, schemas, execution, or assurance.

## Validation

- focused skill/visibility tests: 18 passed
- full unit lane: 891 passed
- complexity baseline: passed
- strict mypy: 470 source files passed
- documentation command validation: passed
- `git diff --check`: passed

The change planner also selects npm packaging and Markdown link checks. This macOS host has neither `npm` nor `npx`, so hosted CI is authoritative for those two environment-only gates.

This PR is intentionally draft and must not be merged by Codex.

### DIFF
--- a/.agents/skills/jacobian-math/SKILL.md
+++ b/.agents/skills/jacobian-math/SKILL.md
@@ -35,7 +35,7 @@ Payloads:
 - For expression normalization, inspect the known
   `polynomial.expression.normalize` contract directly.
 - `matrix.determinant.verify` for an independent check:
-  `{"determinant_uri":"<determinant_uri from compute output>"}`.
+  `{"input":<same determinant payload>,"candidate":<producer output.result>}`.
 - `combinatorics.cyclic_difference_set.extension.decide`:
   `{"base_elements":["1","2","4","8","13"],"target_order":7}`.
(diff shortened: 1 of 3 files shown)


morluto (commented):

**Disposition: close as superseded by #1252.**

The managed repository/npm Skill and its synchronization tests were intentionally deleted in the workflow-runtime cutover. There is no surviving Skill copy to correct, and reintroducing one would restore a second model-facing authority beside the MCP descriptors.

The useful payload lesson should live only in the current verifier’s operation-owned invocation example (see the recommended rewrite of #929), not in packaged Skill prose. This branch has no mergeable target.

Issue #386, opened 2026-08-03
[Design]: Preserve meaningful mathematical intermediates without materializing every result


morluto:
## Scope

Define the remaining gaps in Jacobian's hybrid mathematical-intermediate
retention model now that the inline/durable split, `artifact.put` policy, and
the bounded reasoning trajectory are implemented. The **central residual gap**
is that typed producer-consumer artifact compatibility is only half-wired:
descriptors declare `accepted_artifact_types` but no symmetric
produced-artifact-type contract, so a materialized durable envelope is not
discoverably consumable by URI. This issue tracks that gap and the remaining
materialization/`artifact.put` work. It is a design follow-up to
[#2](https://github.com/morluto/jacobian/issues/2) and
[#3](https://github.com/morluto/jacobian/issues/3) and a narrowing of the
original direction recorded in
[`docs/explanation/persistence-policy-direction.md`](https://github.com/morluto/jacobian/blob/main/docs/explanation/persistence-policy-direction.md).

## What is already implemented (do not re-litigate)

The superseded claims from the original body are removed. The following are
part of the supported surface on `main` (commit `7f5a035`) and must be
preserved, not redesigned:

- **Inline vs. durable is split by operation type, not by a runtime flag.**
  `ComputedOperationAdapter` returns a small typed result inline and writes
  **no** input/result/episode artifact
  ([`operation_installation.py:191-276`](https://github.com/morluto/jacobian/blob/main/src/jacobian/operation_installation.py#L191-L276)).
  `MaterializedOperationAdapter` is the durable path: it writes one input
  artifact, one result artifact with parent lineage, an optional bounded
  `preview`, a typed relationship, and `artifact_uris`
  ([`operation_installation.py:279-393`](https://github.com/morluto/jacobian/blob/main/src/jacobian/operation_installation.py#L279-L393);
  `artifacts.put` at L335 and L341, preview L348-L354, relationships L381-L387).
  The operation contracts carry this intent structurally
  ([`operations.py:57-74`](https://github.com/morluto/jacobian/blob/main/src/jacobian/operations.py#L57-L74)
  for `ComputedOperation`,
  [`operations.py:77-102`](https://github.com/morluto/jacobian/blob/main/src/jacobian/operations.py#L77-L102)
  for `MaterializedOperation` with `preview_model`/`preview`/`preview_complete`).
  Materialization is real and widely used (matrix lattice, projective
  geometry, combinatorics, number theory, posets, certified SNF, graph
[...]

PR #1234, opened 2026-08-11
fix: correct inline verifier lineage guidance

yuelgrace1810-ops:
## Summary

- correct model-facing verifier descriptions that incorrectly called inline computed results “stored”
- correct reference documentation for inline producer → independent checker handoffs across integer and graph operations
- retain artifact-lineage language only for true materializers and bounded searches
- add a registry-wide regression preventing computed-operation verifier cards from claiming stored lineage

## Why

Ordinary `ComputedOperation` producers return typed values inline. Their independent verifiers accept the exact producer `input` and `candidate`, and the resulting verification record binds canonical value digests and checker identity. It does not prove that the candidate came from a particular producer invocation.

Several public capability descriptions and reference pages instead told agents to expect a stored result URI or stored lineage. That field cannot exist for these operations and can cause agents to wait for, invent, or misrepresent provenance.

Explicit materializers and bounded searches still use retained artifact URIs. This change preserves that distinction and does not alter mathematical behavior, checker authority, schemas, or assurance semantics.

## Scope

Updated inline guidance covers:

- integer prime factorization
- integer powerful-number decisions
- graph Hamiltonian path
- graph minimum spanning tree
- graph diameter and radius
- graph distance matrices
- graph maximum matching
- the general domain-operation reference and polynomial GCD tutorial

The static regression checks every built-in computed producer with a checker declaration, so future model-facing verifier descriptions cannot reintroduce “stored” lineage language.

## Validation

- focused exact-domain checker and documentation tests: passed
- unit: 891 passed
- component: 951 passed
- domain: 459 passed
- complexity baseline: unchanged
- strict mypy: 470 source files passed
- documentation command validation: passed
- package build: passed
- `git diff --check`: passed
- local Markdown link validation unavailable because `npx` is not installed; hosted CI remains authoritative

## Overlap

This expands the existing draft rather than opening a duplicate. Issue #386 documents the intended computed-operation architecture and is consistent with the correction. No overlapping open PR was found for these stale verifier descriptions.


morluto (commented):

**Disposition: retain the correction, rewrite on current `main`.**

The central distinction is still important: an inline computed value is not a stored result and does not prove producer-invocation lineage. A verification record should bind the exact operation/checker/conclusion and canonical subject/candidate values; durable artifacts are separate carriers.

Rebase after #1256, update only surviving checker declarations/reference docs, and express the invariant using current `OperationSpec`, `CapabilityResult`, artifact URI, and verification-record terminology. Remove references to generic assurance/completeness and to registries deleted by the cutover. The static regression should inspect current operation-owned descriptions and ensure inline checkers never promise a `result_uri` or stored provenance.

morluto:
Post-#1265 review: **keep, rebase, and update to current docs/tests.** This corrects a real representation/provenance bug and is aligned with the simplified architecture.

Current main still has checker descriptions such as the prime-factorization verifier saying it verifies a “stored” result even though these ordinary computed-operation checkers receive the exact input plus submitted candidate inline. That distinction matters more now, not less: #1256/#1265 made small typed values inline by default and reserved durable artifacts/materializers for a separate publication concern.

The production wording changes are therefore still valid. Rebase onto current main, keep artifact-lineage wording only for genuine materializers/bounded searches, and adapt the registry-wide regression to the current explicit portfolio/bundle construction rather than any removed relationship/result machinery.

I would prioritize this ahead of broader guidance changes because it fixes a concrete false provenance claim in installed checker cards.

morluto:
### Current-main port required; defect remains live

Current `main` still contains computed-operation verifier descriptions using “stored” lineage language, so this PR is not superseded. I am keeping it open.

A mechanical rebase is not appropriate because #1256/#1265 moved and simplified checker declarations, result binding, and several reference pages. The current port should preserve this PR's semantic correction while updating ownership:

- edit the current domain-owned declarations in `domains/number_theory/checkers.py` and `domains/graph_optimization/checkers.py`;
- state that inline verifiers accept the exact producer input plus the complete typed candidate value;
- reserve `result_uri`/stored-lineage wording for real materializers and bounded searches only;
- update the surviving current docs, deleting references to retired generic assurance/completeness language;
- retain a registry-wide invariant, but key it to current `DomainBundle` checker declarations and typed result publication rather than the old computed-operation abstraction.

This is a model-facing correctness fix, not cosmetic documentation. Do not close it until the current descriptors and docs no longer imply nonexistent provenance.

**Disposition: retain and deliberately port; medium-high priority.**

PR #1238, opened 2026-08-11
Fix certified Smith input schema bound

yuelgrace1810-ops:
## What changed

- publish the certified Smith producer's enforced 16×16 input limit in its JSON schema
- state the same bound directly in the capability description
- add a regression test that checks both row and column limits in the generated schema

## Root cause

`CertifiedIntegerMatrix` is shared by producer input and certificate output. Its fields allow dimensions up to 32 because certificate matrices may use that broader output envelope. `CertifiedSmithNormalFormRequest` enforced the producer's stricter 16×16 limit only in a model-level validator, so generated JSON Schema incorrectly advertised 32×32 inputs.

A proof-critical 19×19 integer Gram matrix from Speyer's 2026 minimal counterexample to James's conjecture reproduced the mismatch: the CONTRACT view advertised 32, while execution failed with the correct 16×16 diagnostic. The compatible non-transformation Smith producer and independent verifier both succeeded on the matrix, so this PR changes presentation only.

## Impact

This does not expand or reduce accepted inputs, alter computation, or change assurance. It gives callers the actual producer bound before invocation and preserves the broader certificate-output model.

## Validation

- focused certified-Smith contract/domain tests: 11 passed
- `make lint typecheck`
- `make test-unit`
- direct current-main replay: the 19×19 certified request still fails closed; `matrix.normal_form.smith.compute` returns invariant factors `1` × 12 and `14`, and `matrix.normal_form.smith.verify` returns VERIFIED

### DIFF
--- a/src/jacobian/contracts/certified_snf.py
+++ b/src/jacobian/contracts/certified_snf.py
@@ -48,8 +48,15 @@ class CertifiedIntegerMatrix(ContractModel):
         return self
 
 
+class CertifiedSmithInputMatrix(CertifiedIntegerMatrix):
+    """An integer matrix within the certified Smith producer's input budget."""
+
+    row_count: StrictInt = Field(ge=1, le=MAX_CERTIFIED_SNF_INPUT_DIMENSION)
+    column_count: StrictInt = Field(ge=1, le=MAX_CERTIFIED_SNF_INPUT_DIMENSION)
+
+
 class CertifiedSmithNormalFormRequest(ContractModel):
-    matrix: CertifiedIntegerMatrix
+    matrix: CertifiedSmithInputMatrix
 
     @model_validator(mode="after")
     def require_nonempty_bounded_input(self) -> Self:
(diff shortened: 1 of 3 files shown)


morluto (commented):

**Disposition: retain and revise; the schema mismatch is real.**

Publishing 32×32 while enforcing 16×16 violates the current exact-inspection contract. The strongest fix is not merely extra JSON-schema annotations: give the producer request its own 16×16 input type and keep the broader certificate/output matrix type separate, so unsupported producer states are structurally unrepresentable.

Rebase onto current `main`, derive the generated schema from that narrow request type, and keep the description/regression. Remove generic assurance assertions. If a temporary schema-only constraint is necessary, treat it as a bridge and add a follow-up to split the shared request/output model.

morluto:
Post-simplification review: **keep and rebase.** Tightening the certified-SNF request contract so impossible oversized states fail at the authoritative Pydantic boundary is directly aligned with the current “parse once before preflight/provider work” architecture.

The useful change is the type/schema bound itself and its contract regression. Keep that local to the SNF request model; do not add a second JSON-schema validator or generic schema-reconciliation layer. Exact inspection should simply inherit the corrected generated schema.

Recommendation: rebase onto current certified-SNF operation declarations, verify invalid requests fail before worker/checker execution, and merge.

morluto:
### Superseded by current-main port #1271

I rebuilt this fix directly on current `main` rather than resolving the stale branch against moved contract/operation code.

#1271 preserves the exact three-file scope:

- a request-only `CertifiedSmithInputMatrix` publishes the enforced 16×16 schema bound;
- the broader `CertifiedIntegerMatrix` remains available for certificate outputs and zero-dimensional chain boundaries;
- the operation card states the limit;
- the schema regression locks both dimensions.

No mathematical, backend, publication, or checker behavior changes. Closing this branch to avoid duplicate implementations.

Issue #1060, opened 2026-08-10
[Superseded]: Do not add a mathematical applicability ontology/engine; keep discovery lexical and contracts concrete


morluto:
## Disposition

Closed as not planned. Superseded by #1466, #1477, and concrete operation support checks.

This issue correctly observed that lexical similarity does not prove mathematical applicability. A `QQ` polynomial operation may be wrong for `GF(2)`, a three-variable operation may be wrong for four variables, and a bounded search does not prove a global claim.

Its proposed replacement—`CapabilityApplicabilitySignature`, subject/outcome/domain/dimension/claim/exactness/coverage ontologies, typed requirement queries, incompatibility reason taxonomies, routing-state unions, support-predicate registries, contract cross-validation, and prepared-call binding—would create a second mathematical type system and routing engine beside the concrete request models.

### Discovery

Keep deterministic lexical search over compact package-owned fields:

```text
ID
title
description
tags
one example
```

Add a small ordinary static tag/filter only when a repeated current search need is demonstrated, such as `graph`, `matrix`, `polynomial`, `exact`, or `checker`. Do not make tags authoritative mathematical compatibility.

### Exact inspection

Return the selected operation’s concrete request model/schema, description, bounds, and example. The agent/caller decides whether it matches the task.

### Execution

The concrete request parser and one operation-local support/preflight function reject unsupported domain, dimension, representation, or resource cases precisely before expensive work.

A few operations may expose a concise static support note or local reason code. Do not require every operation to implement the same signature/predicate interface.

### Portfolio gaps

Discovery may honestly return no strong match. It does not need to prove a formal `PORTFOLIO_GAP` over an ontology of all mathematical requests.

## Retained work

- improve titles/descriptions/tags/examples for known agent confusion;
- make concrete request models include field/ring, dimensions, labels/order, and exactness when mathematically necessary;
- use precise operation-local unsupported diagnostics;
- evaluate valid-first-call rate and wrong-operation attempts;
- remove misleading prose/relationships and stale nearby operations.

## Do not introduce

- applicability signature models or registry;
- subject/outcome/domain/claim/exactness/coverage enum ontologies;
- typed task-requirement query language;
- generic incompatibility reason system;
- routing-status algebra for applicable/gap/unknown/unavailable;
- support-predicate IDs/registry;
- schema/signature/protocol cross-validation compiler;
- prepared-call applicability bindings;
- automatic conversion or workflow recommendations.

[...]

PR #1241, opened 2026-08-11
fix(discovery): improve polynomial identity discovery

yuelgrace1810-ops:
## What changed

- add `coefficient-equality` and `sum-of-squares` discovery tags to the existing exact QQ polynomial identity verifier
- add a whole-portfolio regression test for a Jacobian sum-of-squares identity query

## Why

Two independent held-out polynomial workflows surfaced the exact identity verifier below unrelated near-matches. In the Braun–Gwoździewicz–Fernandes–Oréfice-Okamoto degree-seven real-Jacobian witness, the applicable verifier ranked fifth and weak, behind constant-Jacobian, inverse-map, collision, and Jacobian-syzygy operations. The same lower-ranking pattern appeared in the earlier del Valle linear-free-divisor trajectory.

This is a lexical navigation fix only. It does not change schemas, runtime validation, mathematics, checker authority, assurance, or the typed-applicability architecture discussed in #1060.

After the change, the source-bound query ranks `polynomial.identity.verify` second as a strong candidate, immediately after the Jacobian producer.

## Validation

- focused whole-portfolio and polynomial-identity tests: 9 passed
- Ruff check and format check passed
- `git diff --check` passed
- direct mypy passed

The repository `make typecheck` wrapper could not finish dependency synchronization because the sandbox could not download the already-declared `coverage==7.15.2`; no type diagnostic was produced.


morluto (commented):

`coefficient equality` and `sum of squares` are accurate alternate descriptions of the existing exact polynomial-identity checker, so this is an appropriate operation-owned lexical fix.

Rebase onto current `main`, keep only the checker metadata and direct natural-query regression, and remove references to generic assurance and old portfolio ranking infrastructure. Prefer an applicability assertion over freezing an exact ordinal rank, since neighboring operation additions should not break a correct discoverability test.

morluto:
Post-simplification review: **keep and rebase.** Adding `coefficient-equality` / `sum-of-squares` vocabulary to the existing polynomial identity checker is a factual lexical-discovery improvement, not a workflow edge.

That matches the final `math.find` contract: agents search desired mathematical outcomes; exact inspection then supplies the authoritative schema. No new result fields, relationship registry, or alias is needed.

Recommendation: rebase onto current main and keep the whole-portfolio discovery regression. This should remain a very small metadata PR.

morluto (commented):
### Refresh complete

I rebuilt this branch as a true one-commit child of the exact current `main` (`b069742`). Current head: `63091d9`.

The resulting diff is minimal and verified:

- 2 files changed, +17/−0;
- two factual discovery tags on `polynomial.identity.verify`;
- one whole-portfolio sum-of-squares discovery regression;
- no relationship graph, workflow guidance, generic result wrapper, or checker-authority change restored.

CI and Benchmarks both completed successfully. The PR is mergeable and remains draft.

PR #1242, opened 2026-08-11
fix(mcp): prevent capability diagnostics from echoing rejected inputs

yuelgrace1810-ops:
## Summary

Ports the bounded validation-diagnostic fix onto the post-#1256/#1265 ownership model.

- add one bounded public projection for Pydantic validation failures;
- remove rejected input values from typed-expression and explicit-graph diagnostics;
- cap public validation-error collections at eight entries while preserving total and omitted counts;
- retain safe field paths, rule messages, expected contracts, and recovery hints;
- publish the typed-expression 128-node and depth-16 limits directly in its operation description;
- add regressions for rejected-input non-echo, response-size bounds, and multi-error truncation.

## Architecture boundary

The refresh deliberately preserves the current graph semantic path (`jacobian.math.graphs` → `OperationSpec` → narrow graph publisher). It does not restore the pre-cutover graph adapter implementation, generic result metadata, relationship machinery, or checker-service request wrappers. The earlier `exact_domain_checkers.py` edits were dropped from this narrow PR rather than mixing a separate checker-boundary migration into a public-diagnostics fix.

## Why

A valid rejection of a 137-node polynomial AST echoed the complete rejected input through raw Pydantic error projections, producing a response much larger than the actionable diagnostic. The explicit-graph boundary had the same demonstrated issue. Both sites now fail closed without retaining rejected values.

## Validation

The branch is one commit on current `main` at `2b1c961ff4bbc27cd28401e19fcd7b8fbb8173d6`. The final diff changes exactly the two demonstrated public sites, one shared projection module, and their focused tests. No mathematical limits, graph semantics, publication, checker authority, or assurance behavior changes.

### DIFF
--- a/src/jacobian/capability_errors.py
+++ b/src/jacobian/capability_errors.py
@@ -7,6 +7,7 @@ from typing import Any
 from pydantic import ValidationError
 
 from jacobian.contracts.capabilities import CapabilityDiagnostic
+from jacobian.validation_diagnostics import validation_error_message
 
 
 class CapabilityError(RuntimeError):
@@ -56,13 +57,10 @@ def enriched_invalid_request(
     first = errors[0]
     loc = first.get("loc", ())
     path = "/".join(str(part) for part in loc) if loc else None
-    message = str(first.get("msg", ""))
-    if message.startswith("Value error, "):
-        message = message[len("Value error, ") :]
     return base.model_copy(
         update={
             "path": path,
-            "hint": message if message else base.hint,
+            "hint": validation_error_message(exc),
         }
     )
 
--- a/src/jacobian/contracts/polynomial_expressions.py
+++ b/src/jacobian/contracts/polynomial_expressions.py
@@ -30,6 +30,17 @@ MAX_EXPRESSION_TERMS = 1024
 MAX_EXPRESSION_EXPONENT = 127
 MAX_EXPRESSION_INTEGER_DIGITS = 256
 MAX_EXPRESSION_COEFFICIENT_DIGIT_BUDGET = 4096
+POLYNOMIAL_EXPRESSION_PUBLIC_VALIDATION_MESSAGES = frozenset(
+    {
+        "polynomial expression variables must be unique",
+        "polynomial expressions are limited to 128 AST nodes",
+        "polynomial expressions are limited to depth 16",
+        "polynomial expression exponent exceeds 127 for a declared variable",
+        "polynomial expression exceeds the exact coefficient digit budget",
+        "polynomial rational literals are limited to 256 decimal digits",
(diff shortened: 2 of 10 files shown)


morluto (commented):

**Disposition: retain, rebase, and make this the shared diagnostic owner.**

This is aligned with the simplified architecture: one bounded public projection at untrusted Pydantic boundaries, no rejected-value echo, explicit field paths/counts, and no change to mathematical acceptance. It should absorb the surviving regression from stale #923 so checker candidate/resolution failures cannot overflow into a second generic error.

Before merge, rebase onto #1256, update all assertions to the current `CapabilityResult` fields, and remove stale `ERROR`/`HEURISTIC`/assurance wording. Keep the helper transport-agnostic and dependency-light; domain code should raise typed validation failures, while the boundary owns redaction/truncation exactly once.

morluto:
Post-#1265 review: **keep the security/usability fix, but rebase and simplify the ownership.**

The defect class remains present: current code still has multiple `ValidationError.errors()` consumers, and rejected Pydantic inputs must not be echoed into public diagnostics. Using `include_input=False`, bounding the number of projected errors, and retaining only path/rule/recovery information is the right direction.

The branch now crosses paths that #1265 changed (especially graph construction/checker ownership), so do not mechanically rebase the six-file patch. Re-audit current main for every **model-visible** validation projection and fix the surviving public sites only. If two or more sites still need the exact same projection, keep one narrow helper; otherwise prefer the existing capability-error owner instead of creating another general diagnostics framework.

Also preserve the architecture rule that request parsing happens once before preflight/provider work. This should sanitize the error projection, not add a second validation pass or schema abstraction.

Recommendation: revise/rebase, then merge as a bounded diagnostics/privacy fix.

morluto:
### Current-main port required; keep open as a high-priority safety fix

I tested this branch against a base ref identical to current `main`; GitHub reports real conflicts, so I did not force-reparent the old tree.

The conflicts are architectural rather than evidence against the fix:

- `exact_domain_checkers.py` was simplified during #1265 and needs the bounded-message helper inserted into the current verification boundary;
- explicit graph construction now participates in the typed graph/value split, so its validation projection must be added without restoring old result machinery;
- `sympy_polynomial_normalization.py` now returns the simplified current envelope, so only the bounded diagnostics and operation-owned limit wording should be ported;
- tests must assert current execution/result fields rather than the removed generic HEURISTIC/assurance model.

The core design remains correct: one shared `validation_diagnostics` projection, `include_input=False`, bounded collections/messages, and no rejected-value echo. Do not resolve this by accepting the old files wholesale; port the six focused changes onto current owners.

**Disposition: retain, revise on current main, high priority.**

PR #1243, opened 2026-08-11
fix(discovery): restore reciprocal SAT verifier navigation

yuelgrace1810-ops:
## Summary
- expose reciprocal navigation from SAT assignment and DRAT verifiers back to their producers and source CNF
- expose reciprocal navigation from the SMT proof verifier to the proof producer
- add MCP boundary and projection regressions, including optional-provider filtering

## Root cause
The static MCP relationship map only described producer-to-verifier paths. Inspecting a verifier therefore omitted the producer handoff even when both were installed.

## Impact
Agents can inspect either endpoint and recover the intended producer/verifier workflow without another broad catalog search. Optional providers remain filtered from cards when absent.

## Validation
- `pytest -q tests/boundary/mcp/test_mcp_catalog_and_policy.py::test_mcp_compact_capability_index_is_searchable_and_paginated` — 1 passed
- `pytest -q tests/component/adapters/mcp/test_mcp_projection_compaction.py` — 4 passed
- `ruff check src/jacobian/adapters/mcp/projections.py tests/boundary/mcp/test_mcp_catalog_and_policy.py tests/component/adapters/mcp/test_mcp_projection_compaction.py` — passed
- `git diff --check` — passed

### DIFF
--- a/src/jacobian/adapters/mcp/projections.py
+++ b/src/jacobian/adapters/mcp/projections.py
@@ -53,9 +53,17 @@ _RELATED_CAPABILITIES: dict[str, tuple[tuple[str, str], ...]] = {
         ("sat.cnf.materialize", "materialize the exact input CNF"),
         ("sat.model.verify", "independently verify the named assignment"),
     ),
+    "sat.model.verify": (
+        ("sat.cnf.materialize", "materialize the exact input CNF"),
+        ("sat.model.find", "produce a candidate named assignment"),
+    ),
     "sat.unsat_proof.find": (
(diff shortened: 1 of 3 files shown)


morluto (commented):

This PR edits the old MCP relationship projection and reciprocal static map, both removed by #1256. Current exact inspection already returns the complete operation descriptor, and optional-provider availability is represented directly rather than by filtering related-operation cards.

Do not recreate reciprocal SAT/SMT workflow navigation. If a verifier is hard to invoke, put the concrete request example on that verifier’s own `OperationSpec` and improve its lexical metadata. The branch has no surviving code path.

PR #1244, opened 2026-08-11
fix(discovery): improve plain-language polynomial resultant discovery

yuelgrace1810-ops:
## Summary

- add factual `univariate` and `rational` discovery tags to `polynomial.compute.resultant`
- add a full-portfolio regression for natural compute and compute-plus-verify queries
- leave schemas, bounds, execution, assurance, and checker authority unchanged

## Motivation

A proof-critical resultant replay from Chinyere's 2026 Noferini–Williams proof exposed a deterministic routing gap. The exact producer and reciprocal verifier both work, but `math.find` omitted the producer for ordinary queries such as “compute the exact resultant of two univariate rational polynomials.” The terse catalog phrase `polynomial resultant` worked.

The descriptor already states the QQ/univariate semantics in its contract. This patch exposes those factual terms to deterministic lexical retrieval.

## Validation

- `uv run pytest -q tests/composition/runtime/test_portfolio_capability_discovery.py tests/domain/polynomial/test_polynomial_operation_bundle.py tests/domain/polynomial/test_polynomial_verification.py` — 13 passed
- Ruff check and format — passed
- mypy — passed (470 source files)
- unit lane — 891 passed

The broad change-aware fallback was also attempted. Unrelated bounded subprocess tests timed out and several parallel workers were killed under host-wide resource pressure; the focused affected suites above remained green. The composition fallback was stopped instead of spending further resources reproducing unrelated load failures.

## Scope

This is a navigation-only correction. It does not prescribe a workflow, weaken result binding, or change the producer/verifier trust boundary.

### DIFF
--- a/src/jacobian/domains/polynomial/invariants.py
+++ b/src/jacobian/domains/polynomial/invariants.py
@@ -86,6 +86,8 @@ POLYNOMIAL_INVARIANT_CAPABILITIES = (
         "polynomial",
         "resultant",
         "elimination",
+        "univariate",
+        "rational",
         invocation_examples=(
             example(
                 "resultant_x2_minus_one_x_minus_two",
--- a/tests/composition/runtime/test_portfolio_capability_discovery.py
+++ b/tests/composition/runtime/test_portfolio_capability_discovery.py
@@ -106,6 +106,30 @@ def test_discovery_ranks_lexical_matches_and_returns_no_synthetic_fit_labels(
     assert absent.matches == ()
 
 
+def test_discovery_finds_resultant_producer_from_plain_language(
+    attached_complete_runtime_read_only: JacobianRuntime,
+) -> None:
+    capabilities = attached_complete_runtime_read_only.core.capabilities
+
+    for query in (
+        "compute the exact resultant of two univariate rational polynomials",
+        "compute and independently verify the exact resultant of two univariate rational polynomials",
+    ):
+        discovered = capabilities.discover(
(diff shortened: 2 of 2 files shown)


morluto (commented):

**Disposition: retain and rebase; likely merge early.**

Adding `univariate` and `rational` to the resultant operation’s own tags is factual, low-risk, and aligned with the current lexical discovery model. No workflow or relationship machinery is involved.

Rebase onto #1256, update the focused test paths to the current operation declaration/catalog, and remove stale assurance/reciprocal-verifier language from the PR body. The useful invariant is that natural resultant queries surface the producer; checker discovery should be tested independently through its own metadata.

morluto:
Post-simplification review: **keep and rebase.** The `univariate` / `rational` tags make the existing resultant operation more accurately retrievable under the current bounded lexical `math.find` design.

This is exactly the preferred kind of ergonomics improvement now: improve one operation's factual metadata instead of adding projection modes, relationship graphs, or planner hints.

Recommendation: rebase, adapt the discovery regression to current main, and merge as a metadata-only change.

morluto (commented):

This branch was refreshed onto the post-#1265 mainline and has landed as `6e87091`.

The landed patch keeps only the factual `univariate`/`rational` resultant vocabulary and the whole-portfolio discovery regression. It does not restore reciprocal relationship graphs, workflow navigation, alternate result envelopes, or any change to the producer/checker trust boundary.

PR #1247, opened 2026-08-11
fix(discovery): improve finite expectation discovery

yuelgrace1810-ops:
## Summary

- describe the order-one finite raw moment as an exact expectation / expected value
- add expectation, expected-value, and discrete-random-variable discovery tags
- add a model-facing `math.find` regression for the natural query `expectation of a discrete random variable`

## Root cause

`probability.finite_distribution.raw_moment.compute` already computes exact finite expectations when `order=1`, but its public metadata used only the specialist phrase `raw moment`. As a result, a natural expectation query omitted the capability even though a more contract-shaped finite-distribution query found it.

This changes discovery metadata only. It does not change computation, schemas, resource bounds, checker authority, or assurance.

## Evaluation evidence

A deterministic MCP reproduction on current main showed:

- `compute the exact expected value of a finite distribution` ranked the raw-moment operation first;
- `expectation of a discrete random variable` omitted it and returned only convolution;
- after this patch, the natural query includes `probability.finite_distribution.raw_moment.compute` within the bounded five-result probability search.

## Validation

- focused probability and MCP discovery tests passed
- unit: 891 passed
- MCP: 49 passed
- lint, formatting, complexity, mypy (470 files), and package build passed
- broad component: 943 passed, 3 platform skips; four unrelated parallel workers crashed and their exact tests were rerun serially successfully
- broad domain: 454 passed; unrelated worker-startup timeouts remained in integer divisors, discrete logarithm, and Gröbner paths. Serial rerun passed the reported component cases and all domain cases except the two existing advertised-example startup timeouts (`integer.compute.divisors` and `modular.compute.discrete_logarithm`).

No benchmark mathematics or evaluation reward behavior is changed.

### DIFF
--- a/src/jacobian/domains/probability/operations.py
+++ b/src/jacobian/domains/probability/operations.py
@@ -446,12 +446,22 @@ FINITE_PROBABILITY_CAPABILITIES = (
             title="Exact finite-distribution raw moment",
             description=(
                 "Compute one bounded raw moment of a normalized finite exact "
-                "rational distribution, preserving every atom contribution."
+                "rational distribution, preserving every atom contribution. "
+                "Order one is the distribution's exact expectation or expected value."
             ),
             request_type=FiniteRawMomentRequest,
             result_type=FiniteRawMomentResult,
             execute=_raw_moment,
-            tags=("probability", "moment", "finite", "exact", "python-flint"),
+            tags=(
+                "probability",
+                "moment",
+                "expectation",
+                "expected-value",
+                "discrete-random-variable",
+                "finite",
+                "exact",
+                "python-flint",
+            ),
             invocation_examples=(
                 example(
                     "fair_bit_second_moment",
--- a/tests/boundary/mcp/test_mcp_discovery_projection.py
+++ b/tests/boundary/mcp/test_mcp_discovery_projection.py
@@ -10,6 +10,33 @@ from jacobian.adapters.mcp.server import create_server
 from jacobian.contracts.capabilities import CapabilityDescriptor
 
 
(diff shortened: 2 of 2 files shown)


morluto (commented):

The order-one/raw-moment terminology gap is real, and adding `expectation`, `expected value`, and `discrete random variable` to the operation-owned metadata is a simple, composable fix. It does not require another alias operation or routing layer.

Rebase onto current `main`, keep the combinatorics/probability owner unchanged, and update the discovery test to the current lexical matcher. Remove generic assurance wording from the PR and avoid asserting an exact rank if that would make the test brittle; require the operation to be present within the bounded relevant results.

morluto:
Post-simplification review: **keep and rebase.** This is exactly the kind of low-cost discovery improvement that still fits the current architecture.

The operation remains one atomic typed computation; this PR only makes the raw-moment descriptor answer common agent language (`expectation`, `expected value`, `discrete random variable`) and clarifies that order one is expectation. That improves lexical `math.find` without adding recommendations, aliases, relationship graphs, or a second operation.

Recommendation: rebase onto current main, keep the whole-portfolio discovery regression using the current nested `math.find` request shape, and merge. No architectural rewrite is needed.

morluto (commented):

This branch was refreshed onto the post-#1265 mainline and has landed as `2b13f16`.

The landed diff remains metadata-only: it exposes accurate expectation/expected-value vocabulary for the existing order-one raw-moment operation and locks the natural-language discovery route. It changes no mathematical kernel, request/result schema, resource limit, publication behavior, or checker authority.

PR #1250, opened 2026-08-11
fix(discovery): surface polynomial bound checks

yuelgrace1810-ops:
## Summary

- explain on the interval-positivity capability cards that a difference polynomial can encode one expression exceeding a bound
- add inequality and bound-oriented discovery terms to the producer and independent verifier
- add a regression test for the natural query `verify a rational derivative exceeds a proposed bound`

This changes discovery metadata only. It does not change schemas, mathematical execution, assurance, or checker authority.

## Evidence

The issue was reproduced while auditing Mao Shi's 2026 rational Bézier derivative-bound counterexample. Current main could perform the exact interval-positivity check, but the natural task query did not surface it in the first ten polynomial results. A generic positivity query did.

## Validation

- `tests/composition/runtime/test_portfolio_capability_discovery.py`: 6 passed
- lint: passed
- formatting: passed
- complexity check: passed
- mypy: passed (470 source files)
- change-aware test plan reviewed; it selected conservative lane fallbacks because this capability file has no exact ownership mapping

## Overlap

Draft #1191 adds reciprocal relationship navigation for interval operations. This PR addresses the independent semantic-query gap and is compatible with it.

No paid model evaluation was run.


morluto (commented):

This is the correct post-cutover discoverability pattern: expose the actual mathematical interpretation—checking `f - bound > 0` on an interval—through the operation’s own description/tags, then test the natural query directly.

Remove the stated compatibility with #1191 because reciprocal relationship navigation no longer exists. Rebase onto current `main`, keep the change metadata-only, and make the regression assert lexical fit/current operation IDs rather than old portfolio projections or assurance fields.

morluto:
Post-simplification review: **keep and rebase.** The capability already computes/verifies strict positivity; this PR makes the valid inequality/exceeds-a-bound use case discoverable without inventing a second “inequality” operation.

That is aligned with the current portfolio philosophy: one canonical semantic operation, richer factual tags/examples, separate checker ID. Keep the producer/checker independence language, but do not add reciprocal catalog relationships when rebasing.

Recommendation: rebase onto current main, keep a portfolio-level search regression for inequality/bound wording, and merge.

morluto (commented):
### Refresh complete

I rebuilt this branch as a single commit directly on the exact current `main` (`b069742`). Current head: `eaff7d5`.

The refreshed diff is exactly the intended two-file metadata correction:

- explain that a difference polynomial can encode an expression exceeding a bound;
- add inequality/bound discovery vocabulary to the existing producer and checker cards;
- retain one natural-language discovery regression;
- make no mathematical, schema, publication, trust-policy, relationship-graph, or MCP-surface change.

CI and Benchmarks both completed successfully. The PR is mergeable and remains draft.

PR #1253, opened 2026-08-11
fix(runtime): reject volatile shared-memory install roots

N0zoM1z0:
## Summary

- reject /dev/shm install roots both directly and after ancestor resolution
- document durable installation-root constraints for VPS migrations
- preserve deterministic coverage for safe and unsafe resolved ancestors without treating shared memory as durable

## Validation

- make lint typecheck
- make test-process (264 passed)
- make deploy-check (37 passed)
- make docs-linkcheck
- git diff --check

Follow-up for the final review thread on #1246.


N0zoM1z0:
@codex review

PR #1254, opened 2026-08-11
fix(runtime): normalize immutable runtime permissions

N0zoM1z0:
## Summary

- normalize read and traversal/execution permissions on immutable release and managed Python trees after root ownership
- perform normalization before the jacobian service-user runtime validation
- replace the redundant Lean-subtree-only normalization with the complete runtime boundary

## Evidence

A real --with-lean deployment launched under umask 0077 failed closed before activation because the generated .venv entrypoint was not traversable by the jacobian service user. The old production release remained active and the incomplete candidate was removed.

## Validation

- make lint typecheck
- make deploy-check (37 passed)
- bash -n deploy/install.sh
- git diff --check

After merge, deployment will be retried under umask 0077 and must pass the service-user and remote Lean/Mathlib smokes.


N0zoM1z0:
@codex review

PR #1261, opened 2026-08-12
docs(skills): add recent-conjecture reliability evaluation skill

yuelgrace1810-ops:
## Summary

- add a repository-local `recent-conjecture-evaluations` Codex skill
- encode source gating, clean-room oracle construction, current-main contract replay, optional paired evaluation, observable scoring, and failure attribution
- add reusable overlap-search support and link the workflow from contributor guidance
- keep Harbor task authoring and verifier design in the existing specialist skills

## Why

Recent resolved-conjecture probes have exposed useful Jacobian reliability defects, but the workflow was spread across long prompts and evaluator-specific notes. This packages the general method as repository-owned guidance while preserving Jacobian's product boundary: conjectures are probes, deterministic current-main evidence comes first, model calls require information value, and one benchmark family cannot justify a benchmark-specific capability.

## Validation

- skill-creator `quick_validate.py`
- Ruff on `search_inventory.py`
- `tools/check_doc_commands.py`
- repository Markdown link check
- `git diff --check`
- change-aware planner inspected; it conservatively selects the full suite because the new skill path is not yet classified exactly

The full local suite was not rerun because this documentation/skill-only branch was validated in an isolated clone whose dependency sync could not reach PyPI. Hosted CI remains authoritative for the planner-selected lanes.

### DIFF
--- a/.agents/skills/recent-conjecture-evaluations/SKILL.md
+++ b/.agents/skills/recent-conjecture-evaluations/SKILL.md
@@ -1,0 +1,149 @@
+---
+name: recent-conjecture-evaluations
+description: Run source-grounded Jacobian reliability evaluations using recently resolved conjectures as held-out probes. Use for source selection and deduplication, exact input-bound oracles, current-main math.find/math.run contract audits, frozen control/treatment comparisons, observable trajectory scoring, failure attribution, and independent review of another evaluation. Do not use merely to solve a conjecture or create benchmark-specific mathematical helpers.
+---
+
+# Recent Conjecture Evaluations
+
+Treat conjectures as probes. The objective is to improve Jacobian from discovery
+through final-answer use, not to accumulate solved examples.
+
+This skill owns source selection, deterministic capability audits, optional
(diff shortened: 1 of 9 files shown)


morluto:
great that u are adding skills, will review shortly

morluto:
Post-#1265 review: **keep, but rebase/update the skill against the final product model.**

The repository-local evaluation skill is cleanly separated from runtime/product code, which is the correct boundary. The main thing to guard on rebase is that its action policy does not encode the old workflow architecture: `math.find` is bounded search/exact inspection (not the complete inventory), `capability://catalog` is full membership, discovery must not recommend a sequence, and verification remains a separate checker operation with a top-level bound record URI.

Also preserve the current `AGENTS.md` ownership rules rather than replacing them with skill-specific strategy. The skill may orchestrate an evaluation; it should not redefine how the product requires agents to solve arbitrary math tasks.

Recommendation: rebase, update any stale result/tool examples, rerun the skill/script checks, then merge.

PR #1271, opened 2026-08-12
fix(matrix): publish the certified Smith input bound

morluto:
## Summary

Ports the surviving fix from #1238 directly onto current `main`.

- introduce `CertifiedSmithInputMatrix`, a request-only subtype whose generated schema publishes the producer's enforced 16×16 limit;
- keep `CertifiedIntegerMatrix` at the broader 32×32 envelope required by certificate outputs and zero-dimensional chain boundaries;
- use the request-only type in `CertifiedSmithNormalFormRequest`;
- state the 16×16 limit on the operation card;
- lock both row and column maxima in a schema regression.

## Why

The request validator already rejects dimensions above 16, but the generated JSON Schema advertised the shared output model's 32×32 limit. Agents could therefore construct requests that inspection said were valid and execution correctly rejected.

## Architecture boundary

This is a type-owned presentation fix. It does not change accepted inputs, computation, output/certificate bounds, backend selection, publication, checker authority, or verification semantics. The broader matrix value remains the authoritative certificate/output type; the narrower subtype belongs only to the producer request.

## Validation

The branch starts from current `main` at `4c68e4849cbb2ba52d414282763150881996f849`. The focused contract regression asserts the emitted JSON Schema maxima and minima for both dimensions. Existing scalar-bound, canonical-diagonal, rectangle, and zero-dimensional-output tests remain unchanged.

Supersedes #1238.


morluto (commented):
### Current-tree validation complete

This is the clean current-main replacement for #1238: three files, request-only input subtype, unchanged output/certificate envelope, and a schema regression for both dimensions.

CI and Container Image pass on head `e6be616`. The benchmark workflow was cancelled rather than failed; this type-owned schema correction does not alter benchmark content or execution.

**Disposition: review-ready; retain the narrow request/output type split.**

PR #993, opened 2026-08-10
Add independently verified real-quadratic order

kaoru0822-kitauji:
## Summary

- add `arithmetic.real_quadratic.order.compute` for two bounded canonical values `a+b*sqrt(d)` with one shared positive square-free radicand
- retain the exact difference, order, sign basis, and squared-magnitude certificate in a typed result
- add operator-authorized `arithmetic.real_quadratic.order.verify`, replayed in a clean process with standard-library `Fraction` and integer squares and no SymPy or producer import
- keep the scope deliberately below general algebraic-field arithmetic and matrix spectral computation

## Grace-style reproduction

Primary source: Zhekai Pang, *Counterexamples to a multivariable matrix Young conjecture*, arXiv:2607.11866v1 (2026), Theorem 1, `m=4`: https://arxiv.org/abs/2607.11866

Frozen proof-critical scalar subproblem:

`L = 3*sqrt(3)/8`, `R = 1/2 + sqrt(3)/20`, so
`L-R = -1/2 + 13*sqrt(3)/40 > 0`, certified by
`1/4 < 507/1600`.

The authenticated Codex CLI arms used `gpt-5.4-mini`, low reasoning, one run per arm, the same task/schema/oracle, and isolated non-editable installs.

| Arm | Result |
| --- | --- |
| installed `main` at `6df30e45` | mathematics passed, evidence failed; 7 `math.find` calls, 0 `math.run` calls, and all producer/verifier fields were absent |
| this branch | mathematics and evidence passed; direct discovery of the new operation, one producer run (`COMPUTED`), one independent verifier run (`VERIFIED`), record `artifact://sha256/dd0a28b19009e1bc4608c588ecc2797e8d4b7f0ffb19232ef7c645ea070530a4` |

Frozen input hashes:

- task: `9a7931639f7d30c29b14498c9cc47f1a413e21465fdc386ce0035aea7063d6e5`
- output schema: `5f3de7ef08dfc9620b99c3d8deabb33d558980529a004e460116ace5b135e9d4`
- independent oracle: `8ac04cd7a184a362cf183da3d8ed02122d136bd3ae5ebd888d16ea8fe0fcfb29`
- policy: `COMPUTE_VERIFY_NO_RETRIEVAL`, digest `sha256:0749b72e4263a380432feb9973b0fda9d5ed2b04d29417a1262bfc45134c71c6`
- baseline catalog: `sha256:98488060424b44551f7e0d920f6c59a2e76890315a7c34ef5076f1644fd6c23f`
[...]

PR #1287, opened 2026-08-12
feat(arithmetic): add verified real-quadratic order

morluto:
## Summary

Supersedes #993 with a clean post-#1276 semantic port.

- add provider-independent exact values `a+b*sqrt(d)` under `jacobian.math.real_quadratic`;
- compare two values in one shared square-free real quadratic field using exact rational arithmetic and squared magnitudes;
- return the exact difference, order, sign case, and inspectable squared-magnitude certificate;
- install a separate operator-authorized standard-library checker that independently replays the comparison.

## Ownership

The public math layer owns the value and order semantics; it has no SymPy dependency. The arithmetic domain publishes one thin operation and checker declaration. A dedicated clean-process checker uses only `fractions.Fraction` and integer square-free tests.

No general algebraic-field framework, matrix spectral helper, reciprocal relationship graph, generic assurance metadata, or workflow policy is introduced.

## Validation

The branch is based directly on `195c4c1ddcb836206fcc759d602d3cf77eb2d7cf`. Its port workflow compiled every changed source/test file and removed itself before committing. Tests reproduce the `3*sqrt(3)/8 > 1/2+sqrt(3)/20` comparison and reject a forged order.

## Credit

This is a clean post-#1276 port of the original implementation in #993 by @kaoru0822-kitauji. Their original investigation, reproduction, and implementation work are the basis for this replacement.


morluto:
## Attribution

The independently verified real-quadratic order capability originated in #993, authored by @kaoru0822-kitauji. This PR ports that mathematical and verification work onto the current ownership model; the original credit remains theirs.

PR #986, opened 2026-08-10
Bind graph distance-matrix entries to vertex labels

kaoru0822-kitauji:
## Summary

- replace the detached positional `vertices` / `distances` result of `graph.distance_matrix.compute` with v3 source-labelled rows and `distances_by_target` mappings;
- bind every entry directly to its source and target vertex labels while retaining canonical lexicographic serialization;
- advance the independent all-sources BFS witness format to v3 and fail closed on missing, reordered, mislabeled, extra, malformed, or mathematically false rows; and
- document the pre-stable contract change and add focused producer, checker, verification, numeric-label, and adversarial regressions.

This does not add a new capability or workflow policy. The existing atomic distance-matrix producer remains `COMPUTED`, and only the operator-authorized stdlib BFS replay can return `VERIFIED`.

Closes #930.

## Why

The public reproduction is the finite graph used in Zhi-Wei Sun's conjecture inside Tang's 2026 proof of the Sárközy conjecture ([arXiv:2603.29992](https://arxiv.org/abs/2603.29992)). For `p = 17`, the vertices are numeric-looking string labels `"0"` through `"16"`, while Jacobian v1 serialized the matrix in lexicographic order (`"0", "1", "10", ...`).

The v1 artifact declared that order and was mathematically correct, but its row and column labels were detached from the matrix entries. In the frozen weak-model reproduction, the agent declared natural numeric output order and copied the lexicographic matrix as if it already used that order. Jacobian's independent checker correctly verified the stored artifact; the clean-room task verifier correctly rejected the relabeled final answer. This is a representation/handoff defect rather than an error in NetworkX shortest paths or the independent BFS replay.

## Contract and checker boundary

`graph.distance_matrix.compute` version 3 now returns:

- `target_vertices` with explicit target ordering metadata;
- one row per canonical source, each carrying `source_vertex`; and
- `distances_by_target`, an exact mapping covering every declared target.

The Pydantic result boundary validates exact source/target coverage, diagonal, positivity, symmetry, component closure, triangle inequality, and connectedness before publication. The separate `jacobian_checkers` replay requires the exact v3 fields and witness format, reconstructs the matrix from both labels, and exhaustively recomputes every source's distances using its own stdlib BFS implementation. Generic verification still binds the exact input, candidate, semantics, scope, checker identity, and verification record.

The old v1 result shape is intentionally not accepted. Jacobian is pre-stable, and the capability version, semantics version, and witness format all move together to v3.

## Frozen public reproduction

[...]

PR #1293, opened 2026-08-12
Bind graph distance-matrix entries to vertex labels

morluto:
## Summary

Supersedes #986 with a clean current-ownership repair of the graph distance-matrix value.

- replace the positional `vertices` plus nested-list representation with source-labelled rows and target-labelled mappings;
- publish `unweighted-shortest-path-distance-matrix.v3` so every distance is structurally bound to both endpoint labels;
- preserve the bounded all-pairs NetworkX producer behind one thin `OperationSpec`;
- install a separate standard-library checker that independently replays BFS from every source and rejects reordered, rebound, malformed, or false rows;
- retain inline typed publication and a separate verification record.

## Ownership

The graph-invariant contract owns the labelled mathematical value. The graph-optimization domain owns the producer and checker declaration. A dedicated `jacobian_checkers.graph_distance_matrix` module owns the clean-process replay and its declaration-owned provider runtime.

This branch deliberately removes the earlier shared `graph_exact_operations.py` edits and unrelated checker-fixture churn. It is stacked on #1299 only for declaration-owned runtime support; no central runtime map, relationship graph, generic assurance field, or workflow helper is introduced.

## Validation

Focused coverage exercises empty, singleton, connected, disconnected, and numeric-looking vertex labels; structural model rejection; independent public producer-to-verifier replay; and adversarial source-label, target-order, boolean-distance, shortest-path, connectedness, and semantics mutations. Hosted CI is authoritative for the refreshed stack.

Supersedes #986.

## Credit

This is a current-ownership port of the original implementation in #986 by @kaoru0822-kitauji. Their original investigation, reproduction, and implementation work are the basis for this replacement.


morluto:
## Attribution

The source/target-labelled distance-matrix contract and its public reproduction originated in #986, authored by @kaoru0822-kitauji. This PR rebuilds that contribution around the current domain-owned checker runtime seam; it does not supersede the original authorship.

morluto:
#1369 has now merged the complete surviving distance-matrix scope onto current `main`: explicit source/target label binding, exact all-pairs coverage, JSON-null unreachable pairs, independent standard-library BFS replay, and the dedicated polynomial-time 64-vertex request boundary.

That merged implementation is strictly preferable to this stack: retaining #1293 would duplicate the same semantic owner and reintroduce its older 32-vertex `GraphInvariantRequest` boundary. #1369 was authored by @kaoru0822-kitauji, the original author of #986, so the original investigation and implementation provenance are preserved directly in the merged replacement.

Closing this PR as superseded by #1369, not rejected. The declaration-owned runtime work remains independently owned by #1299 and should not be coupled to a duplicate graph operation.

PR #910, opened 2026-08-09
Add verified finite abelian group factorizations

yuelgrace1810-ops:
Closes #908.

## What changed

Adds a bounded exact number-theory producer and independent verifier:

- `finite_abelian_group.exact_factorization.compute`
- `finite_abelian_group.exact_factorization.verify`

The producer works in a declared product of cyclic groups. It:

- normalizes arbitrary bounded integer-vector factor elements;
- exhaustively evaluates every left-plus-right pair;
- returns the complete representation-count histogram;
- decides whether every group element has exactly one representation;
- returns the first missing element and first duplicate representation when false.

A complete coset transversal is represented compositionally as an exact factorization whose right factor is the identity singleton.

The independent checker separately normalizes the payload and replays every sum, binding verification to the exact group presentation, factors, histogram, decision, and witnesses.

## Bounds

- rank at most 6
- group order at most 4,096
- each factor at most 256 elements
- factor Cartesian product at most the group order
- coordinate magnitude at most 1,000,000 before canonical reduction

This does not attempt infinite lattice tiling, orbit-closure reasoning, or arbitrary quotient-presentation canonicalization.

## Why

The held-out audit of Khetan's 2026 Nivat-counterexample cluster could independently verify the Smith-normal-form affine-span argument, but Jacobian could not verify that the eight displayed lattice points form a complete transversal for (2\mathbb Z\times4\mathbb Z). The model had to perform that central exactness claim manually.

The regression uses the paper's eight representatives in (mathbb Z/2\times\mathbb Z/4), including noncanonical coordinates, and obtains a checker-backed exact factorization. A negative regression confirms deterministic missing and duplicate witnesses.

## Validation

- `make check`: passed
  - Ruff and formatting
  - complexity baseline unchanged
  - mypy: 456 source files
  - unit lane: 869 passed
- Full number-theory domain file: 16 passed
- Existing exact-number-theory checker file: 31 passed
- Focused producer/verifier and failure-witness tests: 2 passed
- Documentation command validation: passed

The external Markdown link checker was not run because this environment has no `npx` executable.

PR #1301, opened 2026-08-12
feat(number-theory): add verified finite abelian factorizations

morluto:
## Summary

Supersedes #910 with a current-ownership port of bounded finite abelian group factorization.

- add provider-independent values and exhaustive exact semantics under `jacobian.math.finite_abelian_groups`;
- publish one thin `OperationSpec` for `finite_abelian_group.exact_factorization.compute`;
- normalize bounded integer-vector factors in a declared product of cyclic groups;
- return the complete representation-count histogram, exact-factorization decision, and deterministic first missing/duplicate witnesses;
- represent a coset transversal compositionally by using the identity singleton as the second factor;
- add a separate standard-library checker that independently normalizes and replays every sum.

## Ownership and validation

The result model validates structural invariants only; exhaustive mathematical replay lives in the semantic function and the separate checker, not in a Pydantic validator or capability adapter. The operation returns an inline typed value and creates no artifacts. The checker alone may produce a verification record after operator authorization.

This PR is stacked on #1299 so the number-theory declaration owns its clean-process runtime instead of adding another central checker-runtime mapping.

## Bounds

- rank at most 6;
- group order at most 4,096;
- each factor at most 256 elements;
- factor Cartesian product at most 4,096;
- coordinate magnitude at most 1,000,000 before canonical reduction.

## Validation

Focused coverage uses a noncanonical eight-element transversal of `Z/2 x Z/4`, verifies the complete factorization independently, checks deterministic duplicate witnesses on a negative case, and rejects a forged decision. Hosted CI is authoritative for the full matrix.

## Credit

This is a current-ownership port of the original implementation in #910 by @yuelgrace1810-ops. Their original investigation, reproduction, and implementation work are the basis for this replacement.


morluto:
## Attribution

The bounded exact finite-abelian-group factorization producer and verifier originated in #910, authored by @yuelgrace1810-ops. This PR ports that work into the provider-independent math layer and declaration-owned checker model while preserving the original credit.

morluto:
Reconciled this branch with current main and pushed merge commit 2d0ba831b. The resolution keeps main's checker/runtime installer semantics, then reapplies the finite-Abelian-group declaration and preserves the required optional-provider installation expectation. Focused validation: 61 tests passed.

PR #1302, opened 2026-08-12
feat(matrix): extend exact determinant to order 64

morluto:
## Summary

Supersedes #1020 with a current-tree port of the determinant-owned order-64 contract.

- give `matrix.determinant.compute` an operation-specific exact `QQ` matrix input capped at order 64 while every shared matrix contract remains capped at 32;
- keep execution in `jacobian.math.matrices` using the exact Bareiss backend;
- apply the same order-64 and 256-digit source bounds in the independent Python-FLINT replay, plus the existing 32,768-digit candidate bound;
- bump the producer to version 3 and state the operation-specific scope;
- reproduce the 33×33 truncated Legendre-symbol determinant at `p=71`, obtaining and independently verifying `529`;
- reject order 65 before computation and retain no artifacts.

## Ownership

This is one operation-owned input expansion, not a broad shared-matrix limit change. Rank, RREF, multiplication, characteristic polynomial, and the public raw matrix API remain at order 32. The operation remains a thin `OperationSpec` over the maintained `jacobian.math.matrices` backend, and verification remains a separate checker operation.

## Validation

Focused coverage checks order 33 compute-and-verify, order 65 fail-closed behavior, generated schema bounds, and checker-side source/candidate limits. The branch is one commit directly on current `main`; hosted CI is authoritative for the full matrix.

## Credit

This is a current-tree port of the original implementation in #1020 by @kaoru0822-kitauji. Their original investigation, reproduction, and implementation work are the basis for this replacement.


morluto:
## Attribution

The determinant-owned order-64 contract and order-33 Legendre-symbol reproduction originated in #1020, authored by @kaoru0822-kitauji. This PR is the current-tree port of that contribution; the original authorship remains theirs.

Issue #861, opened 2026-08-08
benchmark: negation/keyword regex accepts false limitation claims (7+ verifiers)


morluto:
## Summary

Limitation validators use `any(term in text for term in (...))` with no negation scoping. A limitation like "This finite counterexample proves a general theorem" passes because it contains "finite" and "general theorem" — the negation context is ignored. Some verifiers have a blocklist (`"formally verified"`, `"proof-assistant verified"`) but it doesn't catch affirmative claims phrased differently.

## Confirmed cases

| Benchmark | Verifier | Issue |
|-----------|----------|-------|
| elementwise-fixed-no-global-invariant | `tests/verifier.py:23-55` | Accepts "This finite counterexample proves a general theorem" (contains "finite" + "general theorem" + "only") |
| newton-polygon-factorization-audit | `tests/verifier.py:29-38` | Accepts "Dumas's theorem is formally verified" if it also contains "newton polygon" and "not formal" elsewhere |
| algebraic-independence-transfer-audit | `tests/verifier.py:23-29` | Accepts contradictory limitation statements |
| marginal-joint-product-audit | `tests/verifier.py:28-38` | Limitation negation not associated with the broad claim |
| fractional-ratio-proof-repair | `tests/verifier.py:23-32` | Evidence terms not associated with asserted relationships |
| local-ring-diagonal-similarity-certificate | `tests/verifier.py:22-37` | Same pattern |
| finite-scheme-rational-points-audit | `tests/verifier.py:25-31` | Same pattern |

## Origin

Unresolved Codex review threads on:
- #519 — "Account for negation in evidence claims" / "Reject affirmative limitation claims"
- #517 — "Reject limitations that affirm formal verification"
- #525 — "Reject contradictory limitation statements"
- #520 — "Associate limitation negation with the broad claim"
- #521 — "Associate evidence terms with their asserted relationships"

## Scope

Substring presence cannot capture semantic negation scope. Require structured limitation fields or use clause-level parsing that binds negation to the claim it modifies. Per-task fix but root cause is shared.

Generated with [Devin](https://devin.ai)

morluto:
Current-main revalidation: keep open. Unordered substring/regex checks cannot establish negation scope or clause association. Convert the affected tasks to structured limitation/claim fields with exact obligation IDs where possible; otherwise treat prose as qualitative output, not deterministic verifier evidence. Do not add a larger keyword/negation regex engine. Track shared protocol migration under #1011.

[...]

PR #1368, opened 2026-08-13
fix(benchmarks): accept natural local-ring evidence phrasing

yuelgrace1810-ops:
## What changed

- accept the natural `PA = BP` spelling in the local-ring evidence explanation
- accept subject-before-verb wording such as “the diagonal entries match”
- add a focused regression using the exact equivalent phrasing observed in the paired benchmark evaluation
- refresh the task verifier checksum

## Why

Round 5 of the controlled Harbor evaluation produced a mathematically correct, fully bound certificate. The verifier recomputed the certificate successfully but assigned zero evidence validity because its phrase list recognized `PA=BP` without spaces and “match the diagonal,” while rejecting the equivalent `PA = BP` and “the diagonal entries match.”

This is a localized alternative-valid-representation rejection. It does not change the mathematical certificate, assurance ceiling, scope, or fail-closed behavior.

## Overlap

PR #1256 changes Jacobian's typed tool interface and does not touch this benchmark verifier. Open issue #861 concerns affirmative false limitation claims being accepted; it does not own this inverse failure mode.

## Validation

- focused task regression: 20 passed
- selected generic verifier attacks: 12 passed
- selected task contract validation: passed
- exact task Oracle: passed
- `make check`: passed
- `git diff --check`: passed


morluto:
wow nice job grace

PR #1504, opened 2026-08-14
Bound integer sequence input digits

### LINKED ISSUE #866: [Bug]: IntegerSequenceRequest has no per-item digit bound

kaoru0822-kitauji:
## Summary

- cap every canonical integer in IntegerSequenceRequest.values at 32,768 decimal digits, excluding the sign
- reuse the repository's canonical scalar ceiling instead of introducing a smaller operation-specific magnitude limit
- test positive and negative values exactly at and one digit above the bound

## Root cause and impact

The request contract bounded the number of sequence items but not each item's digit count, allowing every sequence operation to parse arbitrarily large canonical-integer strings. The new validation runs at the authoritative request boundary before any operation parses the values.

Existing sequence requests within the canonical 32,768-digit ceiling remain valid, including values above CPython's default integer-string conversion limit.

## Validation

- make test-unit TESTS=tests/unit/contracts/test_sequence_contracts.py — 4 passed
- make test-domain TESTS=tests/domain/sequences/test_sequence_operations.py — 3 passed
- make lint typecheck — Ruff, formatting, complexity, and mypy passed
- git diff --check — passed

Fixes #866

### DIFF
--- a/src/jacobian/contracts/sequences.py
+++ b/src/jacobian/contracts/sequences.py
@@ -2,12 +2,19 @@
 
 from __future__ import annotations
 
-from pydantic import Field
+from typing import Self
 
-from jacobian.contracts.exact import CanonicalInteger, CanonicalRational
+from pydantic import Field, model_validator
+
+from jacobian.contracts.exact import (
+    MAX_CANONICAL_RATIONAL_DIGITS,
+    CanonicalInteger,
+    CanonicalRational,
+)
 from jacobian.contracts.results import ContractModel
 
 _MAX_SEQUENCE_LENGTH = 256
+MAX_INTEGER_SEQUENCE_ITEM_DIGITS = MAX_CANONICAL_RATIONAL_DIGITS
 
 
 class IntegerSequenceRequest(ContractModel):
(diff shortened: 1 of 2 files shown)


morluto:
Superseded by #1517, which rebased this exact invariant and its boundary regressions onto current `main` after `ContractModel` moved to `jacobian.contracts.base`.

The replacement explicitly credits @kaoru0822-kitauji in the PR body and squash commit, including a `Co-authored-by` trailer. No mathematical or contract behavior from this PR was dropped.

Issue #1170, opened 2026-08-10
[Testing]: Remove benchmark validation's autouse global import-state cleanup and isolate verifier modules explicitly


morluto:
## Summary

`benchmarks/validation/conftest.py` mutates interpreter-global import state for the entire validation suite and then tries to repair it around every test:

```python
ROOT = Path(__file__).parents[2]
if str(ROOT) not in sys.path:
    sys.path.insert(0, str(ROOT))

@pytest.fixture(autouse=True)
def _isolate_task_verifier_imports():
    original_path = list(sys.path)
    sys.modules.pop("verifier_support", None)
    try:
        yield
    finally:
        sys.modules.pop("verifier_support", None)
        sys.path[:] = original_path
```

This is not hermetic module isolation:

- `sys.path` and `sys.modules` are process-global;
- removing one cache key does not remove modules imported as side effects;
- deleting a `sys.modules` entry does not invalidate references retained by other modules/objects;
- the snapshot already includes the import-time repository-root insertion;
- an autouse fixture hides the dependency from every validation test;
- tests remain order-dependent within each pytest/xdist worker.

The repository already contains both correct primitives needed to replace this pattern:

1. `test_verifier_support_schema.py` loads an exact file under an explicit unique module name with `spec_from_file_location`, `module_from_spec`, and scoped `sys.modules` registration;
2. `_verifier_child.py` executes task verifiers in a fresh interpreter with controlled mounts, path state, module registration, deadlines, and output validation.

The desired architecture is:

> Pure source-module contracts use explicit path-bound module identities; task verifier execution uses a fresh interpreter. No benchmark validation test relies on suite-wide import-cache cleanup.

Static audit baseline: `main` at `2df807e28840c5f73f41686a16176e4e69838232`.

### `conftest.py` changes collection-wide import behavior

[`benchmarks/validation/conftest.py`](https://github.com/morluto/jacobian/blob/2df807e28840c5f73f41686a16176e4e69838232/benchmarks/validation/conftest.py) inserts the repository root into `sys.path` at module import time.

That mutation occurs during pytest configuration/collection, before a test explicitly requests a fixture. The autouse fixture's `original_path` snapshot is captured later and therefore treats the inserted root as baseline state rather than undoing it.

Import resolution for every validation test can consequently differ from:

- the packaged/installed application;
- the fresh verifier child;
- another test runner/import mode;
- a developer importing the module outside pytest.

### Cleanup removes only the exact top-level alias

The fixture pops only:

[...]

PR #1538, opened 2026-08-15
Remove validation import-state fixture

kaoru0822-kitauji:
## What changed

- removed the validation-wide autouse fixture that mutated `sys.path` and the generic `verifier_support` module alias
- extracted the existing path-bound source loader pattern and migrated the seven pure conjecture-probe source tests that still depended on parent-process import state
- made child-verifier tests assert that parent `sys.modules` and `sys.path` state is preserved

Real task verifier execution remains in `run_verifier_in_child`; verifier semantics, reward behavior, and child isolation are unchanged.

## Why

With the fixture disabled, randomized validation exposed one real order dependency: the illumination probe could import the Tutte probe's cached `verifier_support`. Scoping both the exact source identity and its temporary alias removes that dependency without suite-wide cleanup.

Addresses #1170

## Validation

- `make lint typecheck`
- focused pure-loader/child tests in strict forward and reverse file order: 115 passed each
- affected validation families with fixed randomized order: 2,986 passed
- planner-selected `make harbor-host-validation`: four shards, 3,666 passed total
- `git diff --check`

`make harbor-plan BASE=origin/main` selected full host validation and no Oracle or regeneration work.


morluto:
Closing this as superseded: rebasing onto current main exposes tests for retired verifier internals, so a correct rebase would discard most of the commit. I will replace it with a small current-main import-isolation cleanup that preserves the modern validation surface.

PR #1571, opened 2026-08-15
Type boundary sequence certificates

kaoru0822-kitauji:
## Summary
- encode bounded-variation checkpoint rationals as numerator/denominator objects and its sequence/variation formulas as exponent and affine structures
- encode ratio-test boundary terms and identities as rational-function coefficient arrays and all checkpoint bounds as rational objects
- parse typed semantics before replaying convergence/divergence predicates and scalar reward
- preserve acceptance of equivalent unreduced rational representations

Updates include contracts, generated schemas, gold submissions, instructions, verifier checksums, and focused tests.

## Validation
- selected-task Harbor contract check for both tasks
- focused host validation (15 passed)
- focused Ruff check
- Harbor planner selects exactly two host tests and two Oracles

### DIFF
--- a/benchmarks/datasets/mathematical-benchmarks-v1/bounded-variation-uniform-limit/environment/submission_schema.json
+++ b/benchmarks/datasets/mathematical-benchmarks-v1/bounded-variation-uniform-limit/environment/submission_schema.json
@@ -1,4 +1,23 @@
 {
+  "$defs": {
+    "rational": {
+      "additionalProperties": false,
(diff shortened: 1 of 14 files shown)


morluto:
Superseded by #1575 because this contributor-fork branch is no longer updateable from the maintainer checkout. #1575 preserves and credits @kaoru0822-kitauji’s original implementation, carries it onto current `main`, and fixes the stale verifier-checksum CI failure.

PR #1721, opened 2026-08-16
feat(math): add bounded exact code covering radius (#1675)

### LINKED ISSUE #1675: [Feature]: Add bounded exact code theory operations (minimum distance, covering radius, weight distribution)
## Summary

Jacobian has `finite_field.*` operations and `modular.*` operations but no operations for coding theory invariants: minimum distance of a linear code, covering radius, weight distribution, or duality. The conjecture sources (finite geometry, coding theory conjectures, difference set problems) need these invariants as composable building blocks.

## Current behavior

- `finite_field.linear_map.rank.compute` computes rank over a finite field.
- `finite_field.polynomial_map.*` computes polynomial map properties.
- There is no operation that computes the minimum distance of a linear code, its weight distribution, or its covering radius.
- The conjecture sources need coding theory invariants for finite geometry and design theory arguments.

## Requested capability

Add bounded exact coding theory operations:

- `code.minimum_distance.compute`: given a linear code (generator matrix over a finite field), compute the minimum Hamming distance.
- `code.weight_distribution.compute`: compute the weight distribution (number of codewords of each Hamming weight).
- `code.covering_radius.compute`: compute the covering radius of a linear code.

Each operation:
- accepts a bounded generator or parity-check matrix over a declared finite field;
- is backed by exact enumeration for small codes, or a maintained library (e.g., GAP's coding theory package, or SageMath) — not hand-rolled search for large codes;
- fails closed on unsupported field sizes, oversized codes, or budget exhaustion;
- returns `COMPUTED` evidence.

## Why this is in scope

Jacobian's AGENTS.md states that operations may use GAP, SymPy, or another maintained library as a private computational engine. Computing a coding theory invariant is an atomic operation that returns a concrete mathematical value, not a workflow or durable state. These compose with the existing `finite_field.*` operations.

## Acceptance criteria

- [ ] `code.minimum_distance.compute` is supported for small codes.
- [ ] Tests cover a repetition code, a Hamming code, and a budget exhaustion.
- [ ] The operation fails closed on oversized codes and unsupported fields.
- [ ] The operation is discoverable via `math.find` and executable via `math.run`.

## Related

- Existing: `finite_field.linear_map.rank.compute`, `finite_field.polynomial_map.*`
- Conjecture sources: finite geometry, coding theory conjectures, difference set problems

morluto:
## Summary

Closes #1675.

Adds the remaining direct coding-theory operation:

- `code.covering_radius.compute`

The branch has been rebuilt on current `main` as a focused six-file patch. It no longer contains unrelated graph-spectral changes.

## Mathematical kernel

For a linear code `C = row_span(G)` over a declared prime field, the kernel:

1. row-reduces `G` exactly modulo `q`;
2. constructs a basis `H` of the right nullspace, so `C = ker(H)`;
3. performs breadth-first search from the zero syndrome;
4. uses every distinct nonzero scalar multiple of a parity-check column as one move.

One move changes one error coordinate. Thus the shortest path to a syndrome is exactly its coset-leader Hamming weight, and the maximum BFS distance is the covering radius.

## Bounded contract

The request computes `rank(G)` over the prime field before execution and bounds the actual syndrome graph:

- states: `q ** (n - rank(G)) <= 65,536`;
- transitions: `states * distinct_move_upper_bound <= 2,000,000`.

This fixes the prior branch's dependent-row hole: `q ** (n - number_of_rows)` can exponentially *underestimate* the coset count when generator rows are dependent.

The result returns the exact radius with method `SYNDROME_BFS`.

## Tests

Coverage includes:

- binary repetition codes of lengths three and four;
- the binary `[7,4,3]` Hamming code;
- a ternary repetition code;
- dependent generator rows;
- full-space and zero codes;
- non-prime-field rejection;
- regression for the dependent-row state-space hole;
- a separate case whose state count fits but transition count exceeds the work bound.

### DIFF
--- a/src/jacobian/contracts/code_theory.py
+++ b/src/jacobian/contracts/code_theory.py
@@ -8,6 +8,72 @@ from pydantic import Field, model_validator
 
 from jacobian.contracts.base import ContractModel
 
+MAX_EXACT_CODEWORDS = 65_536
+MAX_COVERING_RADIUS_STATES = 65_536
+MAX_COVERING_RADIUS_TRANSITIONS = 2_000_000
+
+
+def _validate_prime_field_matrix(
+    field_order: int,
+    generator_matrix: tuple[tuple[int, ...], ...],
+) -> int:
+    from sympy import isprime
+
+    if not isprime(field_order):
+        raise ValueError("field_order must be prime for this prime-field operation")
+    width = len(generator_matrix[0])
+    if width == 0 or width > 256:
+        raise ValueError("generator rows must have between one and 256 entries")
+    if any(len(row) != width for row in generator_matrix):
+        raise ValueError("generator matrix rows must have equal length")
+    if any(
+        not 0 <= entry < field_order
+        for row in generator_matrix
+        for entry in row
+    ):
+        raise ValueError("generator entries must be canonical field residues")
+    return width
+
+
+def _matrix_rank_mod_prime(
+    matrix: tuple[tuple[int, ...], ...],
+    field_order: int,
+) -> int:
+    rows = [list(row) for row in matrix]
+    row_count = len(rows)
+    column_count = len(rows[0])
+    pivot_row = 0
+    for column in range(column_count):
+        pivot = next(
+            (
+                index
+                for index in range(pivot_row, row_count)
+                if rows[index][column] % field_order != 0
+            ),
+            None,
+        )
+        if pivot is None:
+            continue
+        rows[pivot_row], rows[pivot] = rows[pivot], rows[pivot_row]
+        inverse = pow(rows[pivot_row][column] % field_order, -1, field_order)
+        rows[pivot_row] = [
+            value * inverse % field_order for value in rows[pivot_row]
+        ]
+        for index, row in enumerate(rows):
+            if index == pivot_row:
+                continue
+            factor = row[column] % field_order
+            if factor == 0:
+                continue
+            rows[index] = [
+                (left - factor * right) % field_order
+                for left, right in zip(row, rows[pivot_row], strict=True)
+            ]
+        pivot_row += 1
+        if pivot_row == row_count:
+            break
+    return pivot_row
+
 
 class LinearCodeRequest(ContractModel):
     """A linear code given by its generator matrix over one bounded prime field."""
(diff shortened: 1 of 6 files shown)


morluto:
Audit disposition: **keep, but rewrite before merging**.

There are two independent blockers.

### Contract/algorithm correctness

The request computes the coset bound as `q ** (n - number_of_generator_rows)` and says this safely over-bounds dependent rows. The inequality is reversed: the true count is `q ** (n - rank(G))`; dependent rows make `rank(G) < row_count`, so the current validator can understate the state space exponentially. For example, repeated rows can pass the preflight while the kernel constructs an enormous parity-check space.

Even for full-row-rank input, bounding the number of cosets does not bound the current increasing-weight vector scan. Before all syndromes are reached it may enumerate `sum_w C(n,w)(q-1)^w`, far more than `q ** (n-k)`.

A bounded exact implementation should:

- compute rank over the declared prime field at the request boundary;
- bound `q ** (n-rank)` and the actual transition/work count;
- compute coset-leader distances by bounded BFS in syndrome space (moves are nonzero scalar multiples of parity-check columns), or otherwise prove an equivalent work bound;
- add dependent-row and worst-case boundary tests.

### Branch hygiene

This PR is stacked with unrelated graph-spectral files. It must be rebuilt on current `main` with only the code-theory contract/kernel/declaration/tests.

I am converting it to draft rather than closing it because `code.covering_radius.compute` is still a useful atomic operation. A conflict-only rebase would preserve the exponential preflight hole, so I am not doing that.

morluto:
Final audit validation:

- rebuilt as a focused six-file patch on current `main`;
- removed the unrelated graph-spectral ancestry;
- replaced ambient-vector enumeration with rank-aware syndrome-space BFS;
- bounded both syndrome states and actual BFS transition work;
- added dependent-row, zero/full-space, Hamming, repetition, and work-bound regressions;
- independently compared the kernel with brute-force covering radius on 2,100 random small codes over GF(2), GF(3), and GF(5), with no mismatches;
- required CI, static/type/architecture checks, Python 3.12/3.13 wheels, coverage, benchmarks, and container-image checks all pass.

The branch is now suitable to merge.

PR #1770, opened 2026-08-16
Add floor-sum and congruence-constrained lattice-point counting operations

### LINKED ISSUE #1706: [Arithmetic counting] Add exact floor-sum and congruence-constrained lattice-point operations
## Summary

The resource/conjecture audit found a reusable exact arithmetic-counting kernel in the public `AxiomMath/parity-differential` project:

- https://github.com/AxiomMath/parity-differential
- https://github.com/AxiomMath/parity-differential/blob/main/task.md
- https://github.com/AxiomMath/parity-differential/blob/main/examples.py

For odd `k >= 3`, with `m=(k-1)/2`, the task defines

```text
N_k(n) = # {
  (b1,b2) : 1 <= b1,b2 <= m,
            b1+b2 >= m+1,
            b2 ≡ n*b1 (mod k)
}
```

and

```text
F_k(a) = sum_{i=1}^m floor((a*i+m)/k),
```

then proves the exact finite identity

```text
N_k(n) = F_k(n+1)-F_k(n).
```

The theorem-specific parity statement is not an operation candidate. It exposes two broadly reusable finite functions:

```text
(n,m,a,b) -> sum_{i=0}^{n-1} floor((a*i+b)/m)

bounded integer box / diagonal slice + one linear congruence
    -> exact lattice-point count and residue-class ledger
```

Current Jacobian has exact integer/modular/CRT arithmetic and open issues for general rational-polytope lattice enumeration/Ehrhart theory, but no public operation declaration for floor sums or congruence-constrained box/triangle counts. Repository search on current main finds no `floor_sum`, modular-lattice-count, or congruence-count operation.

An agent can hand-loop over indices or points, but then mathematical floor versus truncation, signed normalization, congruence solvability, interval endpoint semantics, completeness, and the relationship between residue classes and counts are all model-authored.

This issue proposes direct bounded arithmetic/lattice leaves. It does **not** propose a parity-of-differentials wrapper, a Presburger solver, a general polyhedral counting language, or a symbolic summation framework.

Audit baseline: current `main` at `f50dbf6277341f8837ce006d8b10eda8edb6a4b6`.

### 1. Closed integer interval

Conceptually:

```text
lower: bounded integer
upper: bounded integer
```

Use one explicit convention:

```text
{z in Z : lower <= z <= upper}
```

An interval with `lower > upper` may either be a valid empty interval or an invalid request. The preferred counting semantics accepts it as empty and returns count zero, because diagonal slicing naturally produces empty per-row intervals. Make this public and consistent.

### 2. Linear congruence in two integer variables

```text
u*x + v*y ≡ c (mod modulus)
modulus > 0
```

Coefficients and constant are arbitrary bounded integers and normalize modulo `modulus` for congruence arithmetic. The original integers remain available in diagnostics; normalized residues are the computation values.

### 3. Diagonal box slice

A bounded region of one of the two explicit forms:

```text
x in [x_lo,x_hi]
y in [y_lo,y_hi]
x+y >= threshold
```

or

```text
x in [x_lo,x_hi]
y in [y_lo,y_hi]
x+y <= threshold.
```

This is the exact simple geometry used by the source and many modular triangle counts. Do not generalize the first contract to arbitrary caller-supplied inequalities or H-polytopes.

Rectangle counting remains a separate simpler operation with no nullable diagonal fields.

### `integer.floor_sum.compute`

Input:

```text
n >= 0
modulus m > 0
bounded signed integers a,b
```

Compute

```text
S(n,m,a,b) = sum_{i=0}^{n-1} floor((a*i+b)/m)
```

using mathematical floor for signed integers.

Output:

```text
exact integer sum
normalized residues a_mod,b_mod in [0,m)
quotient contributions from Euclidean normalization
bounded Euclidean-recursion ledger:
  current n,m,a,b
  extracted affine contribution
  reduced/swapped state
  cumulative total
term_count = n
```

A standard AtCoder-style Euclidean floor-sum algorithm is appropriate after signed normalization. The ledger is one deterministic replay of the same sum, not a claim that callers must use that algorithm.

Before return, replay every recurrence identity and, within a small secondary check bound, cross-check against direct summation. The production result must not rely on floating division or language-level truncation toward zero.

This operation is deterministic and complete for accepted input. It has no `UNKNOWN` branch.

### `integer.floor_sum.range.compute`

A small optional convenience leaf may compute

```text
sum_{i=L}^{U} floor((a*i+b)/m)
```

through two prefix floor sums, retaining signed `L,U` semantics.

Add it only if direct benchmark/resource demand shows repeated error-prone index shifting. Otherwise callers can transform the affine constant and invoke the prefix operation. Do not add a generic summation-expression API.

### `lattice.linear_congruence.rectangle.count`

Input:

```text
x interval [x_lo,x_hi]
y interval [y_lo,y_hi]
linear congruence u*x+v*y ≡ c (mod m)
```

Output:

```text
exact total point count
one deterministic row per x in the interval:
  x
  effective y interval
  rhs = c-u*x
  g = gcd(v,m)
  is_solvable = (g divides rhs)
  if solvable:
    reduced modulus step = m/g
    one canonical y residue modulo step
    first and last solutions in the y interval
    row_count
sum of row counts
```

For each fixed `x`, solving the one-variable congruence and counting an arithmetic progression in an interval gives a complete exact result. No point-by-point enumeration is required.

If the x interval exceeds the public ledger bound, reject before iteration. A later count-only operation may use periodicity/floor sums to support wider ranges, but it should be a separate concrete leaf with its own completeness proof.

### `lattice.linear_congruence.diagonal_box.count`

Input:

```text
x interval
base y interval
AT_LEAST or AT_MOST diagonal condition on x+y
the same exact linear congruence
```

For each `x`, derive the exact effective y interval:

```text
AT_LEAST: [max(y_lo, threshold-x), y_hi]
AT_MOST:  [y_lo, min(y_hi, threshold-x)]
```

then use the same congruence-progression count.

Output mirrors the rectangle operation and additionally includes:

```text
diagonal condition
base and effective y intervals per x
empty-slice rows
exact total count
```

Use a closed `AT_LEAST | AT_MOST` discriminator. Do not expose independently nullable lower/upper diagonal fields.

### `lattice.linear_congruence.diagonal_box.enumerate`

A separately stricter leaf may return the complete canonical point set for small regions:

```text
all and only (x,y) satisfying the intervals, diagonal condition, and congruence
exact count
per-x residue ledger
```

This is useful for concrete certificates and benchmark conformance. It must have a smaller area/output bound than the count operation. Do not add an `include_points` flag that changes result identity and feasible envelope inside one operation.

### Possible later batch leaf

A batch operation could count several congruences or several diagonal thresholds over the same region while reusing arithmetic. Add it only after concrete repeated demand. Do not front-load a generic system of modular inequalities.

## Why generic polytope/Ehrhart operations are not enough

#1192/#1669 correctly own exact lattice enumeration and Ehrhart counting for bounded rational polytopes. This issue is a different lower-level arithmetic family:

- the region is an axis-aligned box or one diagonal slice;
- a linear modular congruence is interpretation-critical;
- the result has a one-dimensional residue-class decomposition;
- exact floor sums can evaluate large arithmetic-progressive counts without enumerating a polytope; and
- the source identities compare floor differences with congruence counts directly.

A future general polytope operation may privately reuse these leaves for simple two-dimensional cases. Do not make this issue depend on a general H/V-polytope value or Barvinok backend.

Likewise, CRT can help solve congruences but does not own interval counts, diagonal slicing, or complete per-row progressions.

## Public bounds

Enforce before recursion or row-ledger construction:

### Floor sum

- `n` magnitude/digits;
- modulus magnitude/digits;
- `a,b` digits;
- predicted/intermediate product and result digits;
- Euclidean recursion depth;
- ledger row count; and
- aggregate output bytes.

The logarithmic algorithm may support very large `n`; the digit and intermediate bounds remain finite and public.

### Congruence counts

- coefficient/constant/modulus digits;
- x/y interval endpoint digits;
- x-row count;
- effective interval/intermediate digits;
- modular inverse/gcd arithmetic size;
- total result digits;
- explicit point count for enumeration; and
- aggregate result bytes.

The count operation need not bound the entire rectangle area when it uses one row per x, but it must bound the x ledger and exact result/intermediate size. The enumeration leaf must bound total points/area separately.

Out-of-envelope input is invalid. It is not a zero count or an unsatisfiable congruence result.

### Floor sum

- `n=0` yields zero.
- Floor means greatest integer `<=` the rational value, including for negative numerators.
- Replacing `b` by `b+q*m` changes the sum by `q*n`.
- Replacing `a` by `a+q*m` changes the sum by `q*n*(n-1)/2`.
- Equivalent Euclidean-normalized inputs yield exactly related totals.
- No overflow or floating rounding is permitted.

### Congruence counting

- Coefficients/constants reduced modulo `m` define the same congruence.
- Translating x/y intervals and changing `c` coherently preserves the count.
- Multiplying the entire congruence by a unit modulo `m` preserves the solution set.
- When `gcd(v,m)` does not divide `c-u*x`, that x row has no solutions.
- Empty intervals and empty diagonal slices contribute zero under the public convention.
- Rectangle/point row order is not mathematical; `(x,y)` identity is.
- Boundary points satisfying equality in `x+y >= threshold` or `<= threshold` are included.
- Count and enumeration operations must agree exactly when both accept the same request.

## Implementation direction

Use small exact integer kernels:

- Euclidean floor division/modulo with explicit mathematical semantics;
- the logarithmic floor-sum recurrence;
- gcd/extended-gcd/modular inverse;
- arithmetic-progression first/last/count formulas; and
- direct enumeration only in the stricter point-output leaf.

Return only Jacobian-owned immutable integer intervals, congruence rows, point sets, and exact integers.

Before return, replay:

- every floor-sum recurrence step;
- every fixed-x congruence solution class;
- first/last solution membership;
- row arithmetic-progression count;
- diagonal interval intersection; and
- total aggregation.

Do not expose:

- Python `range`/generator state;
- symbolic summation expressions;
- Z3/Presburger formulas;
- an arbitrary inequality AST;
- a polyhedral model/session;
- floating approximations;
- a parity-differential theorem wrapper; or
- a workflow/planner.

### Floor sum

1. `n=0` and `n=1`;
2. small positive parameters cross-checked by direct summation;
3. negative `a`, negative `b`, and both negative under mathematical floor;
4. `m=1`;
5. `a=0` and constant floors;
6. metamorphic `b -> b+q*m`;
7. metamorphic `a -> a+q*m`;
8. a large-`n` case impractical to enumerate but checked by independent recurrence identities;
9. one corrupted recursion-ledger step rejected by result postconditions;
10. exact behavior at and immediately above digit/recursion/output bounds.

### Congruence rectangle/diagonal counts

11. one uniquely solvable residue class per x when `gcd(v,m)=1`;
12. a nontrivial gcd with solvable and unsolvable x rows;
13. modulus one, where every point satisfies the congruence;
14. negative coefficients and interval endpoints;
15. empty x/y intervals and empty diagonal slices;
16. boundary equality on the diagonal;
17. `AT_LEAST` and `AT_MOST` on the same box;
18. coherent coefficient/constant normalization modulo `m`;
19. unit scaling of the congruence;
20. count/enumeration agreement on a small region;
21. a point-set mutation or omitted point rejected by the enumeration postcondition;
22. requests exactly at and immediately above x-row, coefficient, endpoint, point, and output bounds;
23. the exact `N_k(n)` region from the public parity-differential task; and
24. independent composition with `integer.floor_sum.compute` establishing the finite fixture equality `N_k(n)=F_k(n+1)-F_k(n)` without adding a theorem-specific operation.

## Acceptance criteria

- [ ] `integer.floor_sum.compute` supports bounded signed exact inputs and mathematical floor semantics.
- [ ] Its result includes a replayable logarithmic Euclidean recurrence and exact total.
- [ ] Rectangle congruence counting returns one complete exact solution-class/count row per x.
- [ ] Diagonal-box counting derives effective intervals structurally and includes boundary points correctly.
- [ ] A separate bounded enumeration leaf, if implemented, returns all and only satisfying points and agrees with the count leaf.
- [ ] Congruence normalization, interval translation, unit scaling, and floor-sum metamorphic identities are tested exactly.
- [ ] Empty/unsolvable cases are mathematical zero counts, while resource overflow remains invalid input.
- [ ] Operations compose through direct integer/interval/point values and may support the parity-differential fixture without knowing its theorem.
- [ ] General rational-polytope/Ehrhart work remains owned by #1192/#1669.
- [ ] No parity-differential wrapper, Presburger solver, arbitrary inequality/summation DSL, polytope framework, artifact layer, or workflow engine is introduced.

## Related

- Existing modular, gcd, extended-gcd, inverse, and CRT operations should be reused privately where conventions align.
- #1192/#1669 own general bounded rational-polytope lattice enumeration and Ehrhart polynomials; this issue owns exact floor sums and one-congruence box/diagonal arithmetic.
- #1682 owns additive representation counts in integer and finite-Abelian sets, not geometric interval/congruence counts.
- The public `AxiomMath/parity-differential` project is a conformance source, not a requested special-case tool.

morluto:
Closes #1706

Implements two atomic composable math operations:
- `arithmetic.floor_sum.compute` - sum_{i=0}^{n-1} floor((a*i+b)/m)
- `arithmetic.congruence_constrained.count` - count lattice points (b1,b2) with lower<=b1<=upper, 1<=b2<=m, b1+b2>=m+1, b2≡n*b1 mod k

Built on exact integer arithmetic.

### DIFF
--- a/src/jacobian/builtin_operation_modules.py
+++ b/src/jacobian/builtin_operation_modules.py
@@ -53,6 +53,7 @@ BUILTIN_OPERATION_MODULES: tuple[BuiltinOperationModule, ...] = (
         "rational_linear_operations",
     ),
     ("jacobian.domains.lattices", "lattice_operations"),
+    ("jacobian.domains.formal_power_series", "formal_power_series_operations"),
     ("jacobian.domains.polynomial", "polynomial_operations"),
     ("jacobian.domains.analysis", "real_analysis_operations"),
     (
--- a/src/jacobian/contracts/arithmetic_counting.py
+++ b/src/jacobian/contracts/arithmetic_counting.py
@@ -1,0 +1,47 @@
+"""Typed contracts for arithmetic counting operations."""
+
+from __future__ import annotations
+
+from typing import Self
+
+from pydantic import Field, StrictInt, model_validator
+
+from jacobian.contracts.base import ContractModel
+
+
+class FloorSumRequest(ContractModel):
+    """Request to compute sum_{i=0}^{n-1} floor((a*i+b)/m)."""
+
+    n: StrictInt = Field(ge=0, le=100000)
+    m: StrictInt = Field(gt=0, le=100000)
+    a: StrictInt = Field(ge=0, le=100000)
+    b: StrictInt = Field(ge=0, le=100000)
+
+
+class FloorSumResult(ContractModel):
+    """Result of a floor sum computation."""
+
+    value: StrictInt = Field(ge=0)
+
+
+class CongruenceConstrainedCountRequest(ContractModel):
+    """Count lattice points (x,y) with lower <= x <= upper, 1 <= y <= m, and y ≡ c*x (mod k)."""
+
+    k: StrictInt = Field(gt=0, le=10000, description="Modulus k > 0.")
+    m: StrictInt = Field(gt=0, le=10000, description="Upper bound on both coordinates.")
+    n: StrictInt = Field(ge=0, le=10000, description="Multiplier for congruence y ≡ n*x mod k.")
+    lower: StrictInt = Field(ge=0, le=10000, description="Lower bound on b1 (inclusive).")
+    upper: StrictInt = Field(ge=0, le=10000, description="Upper bound on b1 (inclusive).")
+
+    @model_validator(mode="after")
+    def validate_bounds(self) -> Self:
+        if self.lower > self.upper:
+            raise ValueError("lower must be <= upper")
+        return self
+
+
+class CongruenceConstrainedCountResult(ContractModel):
+    """Result of a congruence-constrained lattice point count."""
+
+    count: StrictInt = Field(ge=0)
+    detail: str = Field(min_length=1, max_length=1024)
--- a/src/jacobian/contracts/graph_coloring_ops.py
+++ b/src/jacobian/contracts/graph_coloring_ops.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from typing import Self
+from typing import Literal, Self
 
 from pydantic import Field, model_validator
 
@@ -53,3 +53,29 @@ class MaximumIndependentSetRequest(ContractModel):
 class MaximumIndependentSetResult(ContractModel):
     independent_set: tuple[int, ...]
     cardinality: int = Field(ge=0, le=20)
+
+
+class MaximalIndependentSetRequest(ContractModel):
+    """Decide whether a given vertex set is a maximal independent set."""
+
+    graph: GraphEdgeList
+    candidate_set: tuple[int, ...] = Field(max_length=20)
+
+    @model_validator(mode="after")
+    def require_valid_candidate(self) -> Self:
+        seen: set[int] = set()
+        for vertex in self.candidate_set:
+            if not (0 <= vertex < self.graph.vertex_count):
+                raise ValueError("candidate vertices must lie in 0..vertex_count-1")
+            if vertex in seen:
+                raise ValueError("candidate set must contain no duplicate vertices")
+            seen.add(vertex)
+        return self
+
+
+class MaximalIndependentSetResult(ContractModel):
+    """Decision outcome for a candidate vertex set."""
+
+    decision: Literal["MAXIMAL", "NOT_INDEPENDENT", "INDEPENDENT_NOT_MAXIMAL"]
+    candidate_set: tuple[int, ...]
+    vertex_count: int = Field(ge=1, le=20)
--- a/src/jacobian/domains/arithmetic_counting/__init__.py
+++ b/src/jacobian/domains/arithmetic_counting/__init__.py
@@ -1,0 +1,16 @@
+"""Arithmetic counting operations."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from jacobian.math_tools import MathTools
+
+__all__ = ["arithmetic_counting_operations"]
+
+
+def arithmetic_counting_operations() -> MathTools:
+    from jacobian.domains.arithmetic_counting.math_tools import ARITHMETIC_COUNTING_OPERATIONS
+
+    return ARITHMETIC_COUNTING_OPERATIONS
--- a/src/jacobian/domains/arithmetic_counting/math_tools.py
+++ b/src/jacobian/domains/arithmetic_counting/math_tools.py
@@ -1,0 +1,74 @@
+"""MathTool declarations for arithmetic counting operations."""
+
+from __future__ import annotations
+
+from jacobian.contracts.arithmetic_counting import (
+    CongruenceConstrainedCountRequest,
+    CongruenceConstrainedCountResult,
+    FloorSumRequest,
+    FloorSumResult,
+)
+from jacobian.domains._examples import example
+from jacobian.domains.arithmetic_counting.operations import (
+    compute_congruence_constrained_count,
+    compute_floor_sum,
+)
+from jacobian.math_tools import MathTool
+
+
+ARITHMETIC_COUNTING_OPERATIONS: tuple[MathTool, ...] = (
+    MathTool(
+        operation_id="arithmetic.floor_sum.compute",
+        version="1",
+        title="Compute sum_{i=0}^{n-1} floor((a*i+b)/m)",
+        description=(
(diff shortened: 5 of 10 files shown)


morluto:
Closing as superseded by #1817. The floor-sum and congruence-constrained counting operations landed there as the bounded arithmetic-counting domain and were reviewed together with the repository-wide exact-contract hardening.

PR #1974, opened 2026-08-18
Add first-order term rewriting operations (#1921)

morluto:
## Summary

Implements core first-order term rewriting operations for the Jacobian MCP server, addressing issue #1921.

## Design

First-order terms are represented as trees with integer-symbolled function applications and variables:

- `term_rewriting.substitution.compute` — apply a variable-to-term substitution
- `term_rewriting.matching.compute` — one-way matching (pattern vs. ground subject)
- `term_rewriting.unification.compute` — most general unifier with occurs check
- `term_rewriting.rewrite_step.compute` — one leftmost-outermost rewrite step
- `term_rewriting.normal_form.compute` — iterated rewriting with step bound and convergence flag

## Library choices

Direct recursive tree algorithms — no external rewriting library needed. Terms are immutable Pydantic models with recursive `children` tuples, enabling structural equality and tree-shakeable composition.

## Tests

18 tests covering:
- Substitution (variables, children, no-op)
- Matching (variables, functions, symbol/arity mismatch)
- Unification (same symbol, variables, failure, occurs check)
- Rewrite step (root, child, no-match)
- Normal form (convergent, divergent with step limit)
- Validation: variable with children, LHS must be function

[Continue this on Linzumi](https://serve.linzumi.com/w/williamlin1327-workspace/thread/general/22f3d42f-c745-43f4-af74-fe0d58ea1090)

Issue #1921, opened 2026-08-17
[First-order term rewriting] Add exact substitutions, matching, unification, rewrite steps, critical pairs, normal forms, and confluence profiles


morluto:
## Summary

The continued research/resource, benchmark, source-tree, and live issue-ownership audit found a missing first-order rewriting layer between Jacobian's finite signatures/terms, finite universal algebra, context-free grammars, free associative algebras, formal proof tooling, and finite automata:

```text
finite first-order signature + finite terms with variables
    -> exact positions, subterms, replacement, substitution, and composition
    -> one-way matching and first-order unification

finite term-rewrite system
    -> all one-step redexes/reducts
    -> supplied rewrite-sequence replay
    -> deterministic reduction strategies
    -> complete normal-form family under structural termination

rewrite rules
    -> complete overlap and critical-pair family
    -> left-linearity / orthogonality profiles
    -> joinability and local/global confluence under explicit hypotheses
    -> one bounded completion/interreduction step under a supplied reduction order
```

#1912 owns finite universal-algebra signatures, terms, term evaluation, induced term-operation tables, subalgebras, congruences, and homomorphisms. It explicitly excludes term rewriting and Knuth–Bendix. #1889 owns free associative/noncommutative polynomial words and Gröbner–Shirshov-style reductions; a word is not a branching first-order term with variables, substitutions, arbitrary function-symbol arities, and nested overlap positions. #1894 owns context-free productions and parse trees, not general term equations or first-order rewriting.

Current source search at audit baseline found no rewrite-rule, redex, critical-pair, unification, term-order, or first-order TRS operation. Live issue search found no owner.

Primary references and formal precedent:

- J. W. Klop, *Term rewriting systems from Church–Rosser to Knuth–Bendix and beyond*: https://ir.cwi.nl/pub/2667
- J. W. Klop, *Term Rewriting Systems*, Handbook of Logic in Computer Science: https://doi.org/10.1093/oso/9780198537618.003.0001
- A. L. Galdino and M. Ayala-Rincón, formalization of the Knuth–Bendix–Huet critical-pair theorem: https://doi.org/10.1007/s10817-010-9165-2
- N. Dershowitz, L. Marcus, and A. Tarlecki, *Existence, Uniqueness, and Construction of Rewrite Systems*: https://doi.org/10.1137/0217039

[...]

PR #1976, opened 2026-08-18
refactor(catalog): curate public operation basis

kaoru0822-kitauji:
## Summary

- curate the built-in operation inventory from 360 candidates to a 200-operation public MCP basis
- add one explicit, exhaustive, fail-closed admission decision for every candidate
- retain 56 useful projections and classical helpers under supported `jacobian.math` APIs while dropping 104 catalog-only or redundant helpers
- repair exact Nash-equilibrium and stationary-distribution contracts
- publish the frozen audit of all open mathematical PRs and the admission policy

## Classification

| Decision | Count | Effect |
| --- | ---: | --- |
| `KEEP` | 200 | Published through `math.find` / `math.run` |
| `NATIVE_ONLY` | 56 | Supported through explicit `jacobian.math.<domain>` APIs |
| `DROP` | 104 | Removed from supported discovery |
| `SPLIT` | 0 | No unresolved aggregate enters the catalog |
| `CONTRACT_FIX` | 0 | No unresolved contract defect enters the catalog |

Catalog construction now compares the complete 360-operation candidate tuple with the exhaustive ledger and fails closed on missing, stale, or duplicate decisions. Only `KEEP` rows are published.

## Notable curation decisions

- Remove scalar arithmetic, finite-set and finite-sequence convenience operations, redundant invariant projections, and duplicate Boolean/graph results from MCP discovery.
- Move graph radius, biconnected components, SCCs, DFA complement, convergents, classical combinatorial numbers, formal-series projections, finite-metric balls, Young-diagram helpers, matrix projections, and finite-field projections to supported native APIs.
- Retain three coherent complete profiles after a second accidental-loss review: pairwise-distance multiplicities, finite-metric profiles, and numerical-semigroup gap profiles.
- Update public examples, CI wheel smoke, documentation, and benchmark selection expectations away from removed scalar GCD and other excluded IDs.

## Contract repairs

- Replace heuristic zero-sum-game support/grid search with exact primal-dual linear programming. Shifted payoffs correctly support negative game values; returned strategies and minimax inequalities are replay-checked exactly.
- Replace the arbitrary first stationary eigenvector with the canonical extreme stationary distribution for every closed communicating class, explicitly reporting uniqueness and the complete stationary simplex basis.
- Correct the finite-metric profile method claim from `FLOYD_WARSHALL` to `DIRECT_DISTANCE_MATRIX_SCAN`.

## Migration implications

- Callers using removed IDs should compose retained typed results or use the documented native Python helper.
[...]

PR #1979, opened 2026-08-18
Make term rewrite choices explicit and complete (#1921)

kaoru0822-kitauji:
Focused replacement for #1974, cleanly restacked on current `main` without its unrelated poset commit/snapshot entries.

## Rewrite-step semantics

The old operation silently chose the first matching rule at the leftmost-outermost position. The request now has two explicit scopes:

- omit `selection` to return `ALL_APPLICABLE_STEPS`;
- provide `selection.position` and `selection.rule_index` to request exactly that derivation.

Every application carries the exact position, rule index, matching substitution, and resulting term. Distinct derivations remain distinct even if they happen to produce equal terms. Result validation replays the declared scope and rejects missing, extra, or policy-selected steps.

An adversarial term with root/nested overlap and two matching rules returns all four `(position, rule)` pairs. Explicit selection of nested rule 1 returns only that result. Invalid positions/rule indices fail request validation; a valid but inapplicable selected rule returns an exact empty application family.

## Related correctness repairs

- bounded normalization now requires the strategy `LEFTMOST_OUTERMOST_RULE_ORDER` explicitly;
- `NORMAL_FORM` is returned only after proving no next rewrite exists, including when normal form is reached exactly on the step boundary;
- `STEP_LIMIT` carries the exact next-step witness and is not called divergence/nonconvergence;
- rewrite rules reject RHS variables absent from the LHS;
- unification now returns an idempotent MGU: chained bindings are composed so replaying the substitution actually makes both inputs equal;
- the rule family is bounded by the declared 64-rule contract.

## Validation

- `make test-math TESTS=tests/math/term_rewriting` — 25 passed
- `make test-catalog TESTS=tests/catalog/test_snapshot.py` — 4 passed
- `make check` — lint, formatting, complexity, mypy, and 1,002 tests passed
- `git diff --check` — passed

`make test-plan BASE=origin/main` is unavailable on this base (`No rule to make target test-plan`).

No routing, MCP naming, or additional catalog-admission changes are included; the replacement preserves the original term-rewriting operation declarations only.

### DIFF
--- a/src/jacobian/catalog/builtins.py
+++ b/src/jacobian/catalog/builtins.py
@@ -32,6 +32,9 @@ from jacobian.math.finite_game_theory._tools import TOOLS as FINITE_GAME_THEORY_
 from jacobian.math.finite_metric_spaces._tools import (
     TOOLS as FINITE_METRIC_SPACES_TOOLS,
 )
+from jacobian.math.term_rewriting._tools import (
+    TOOLS as TERM_REWRITING_TOOLS,
+)
 from jacobian.math.finite_sets._tools import TOOLS as FINITE_SETS_TOOLS
 from jacobian.math.formal_power_series._tools import TOOLS as FORMAL_POWER_SERIES_TOOLS
 from jacobian.math.geometry._tools import TOOLS as GEOMETRY_TOOLS
@@ -145,6 +148,7 @@ BUILTIN_TOOLS: MathTools = (
     *ALGEBRAIC_COMBINATORICS_TOOLS,
     *REAL_ALGEBRA_TOOLS,
     *FINITE_METRIC_SPACES_TOOLS,
+    *TERM_REWRITING_TOOLS,
 )
 
 __all__ = ["BUILTIN_TOOLS"]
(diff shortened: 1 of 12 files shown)


kaoru0822-kitauji:
@morluto @N0zoM1z0 — correctness work is complete and CI is green. Please review this after #1976 lands: explicit all-step/selected rewrite semantics, replayable substitutions, and fail-closed normalization are ready, but the branch should then rebase onto the curated admission ledger and apply the final public-vs-native operation decisions before merge.

kaoru0822-kitauji:
Boundary correction pushed in `344b98e66` (tree `253f1ccf8c3b20b264df6d29a579fe85c51c0614`).

The catalog now exposes only the audited matching, unification, and explicit-choice rewrite-step outcomes. Substitution and bounded normalization remain supported in `jacobian.math.term_rewriting`; native selected-step and normalization calls now fail explicitly for an invalid rule index or negative bound. No #1976 policy infrastructure is duplicated.

Validation on the final tree:
- focused capability/catalog/native/import suite: 42 passed
- `make check`: 1,005 passed, lint/format/complexity/typecheck clean
- `make test-plan BASE=61589543b`: unavailable on current `main` (`No rule to make target test-plan`)
- catalog: 363 operations, digest `sha256:84dda3e3b37f439b1ed2634cba60442f2bc9d9c7ac3a4e49ecaef919950f7696`
- policy digest: unavailable until #1976 lands; rebase remains intentionally deferred

morluto:
Superseded by #1997: this draft's capability set was cherry-picked, conflict-resolved, and admitted to the public operation ledger on `agent/integration-capabilities`. Closing as merged-by-integration.

PR #1971, opened 2026-08-18
Add finite topology operations (#1920)

morluto:
## Summary

Implements core finite topological space operations for the Jacobian MCP server, addressing issue #1920.

## Design

Every finite topology is Alexandrov and equivalent to a specialization preorder. Operations work directly on open sets and their derived structure:

- `topology.specialization_preorder.compute` — specialization preorder matrix
- `topology.closure.compute` — smallest closed set containing a subset
- `topology.interior.compute` — largest open set contained in a subset
- `topology.connected_components.compute` — partition via preorder graph
- `topology.is_continuous.compute` — preimage-of-open-set check for point maps
- `topology.beat_points.compute` — up/down beat point detection

## Library choices

Direct set operations and BFS — no external topology library needed. The implementation follows the existing `StrictModel` Pydantic pattern and reuses set/frozenset for open/closed set computation.

## Tests

21 tests covering:
- Discrete, trivial, and Sierpinski topologies
- Specialization preorder for each topology type
- Closure and interior with known-answer verification
- Connected components in connected and disconnected spaces
- Continuity (identity, constant, non-continuous maps)
- Beat point detection for Sierpinski and trivial topologies
- Validation: missing empty/full sets, out-of-range points, duplicate open sets

[Continue this on Linzumi](https://serve.linzumi.com/w/williamlin1327-workspace/thread/general/22f3d42f-c745-43f4-af74-fe0d58ea1090)

PR #1987, opened 2026-08-18
Add exact finite topology capabilities

kaoru0822-kitauji:
Replacement for #1971 after the maintainer audit classified that PR as SPLIT.

This PR rebuilds finite topology from current `main` and exposes only the four audited outcomes:

- `topology.specialization_preorder.compute`
- `topology.connected_components.compute`
- `topology.is_continuous.compute`
- `topology.beat_points.compute`

Closure and interior remain supported native functions in `jacobian.math.finite_topology`; they are not separate public catalog operations.

Correctness changes relative to #1971:

- validates the full finite-topology axioms, including canonical open sets and binary union/intersection closure, before any computation;
- gives the specialization matrix an explicit orientation: relation[x,y] means x is in the closure of singleton y;
- binds point-map carrier sizes to the exact domain and codomain topologies;
- returns the first codomain open set and its non-open preimage when continuity fails;
- restricts beat-point computation to T0 finite spaces so strict-order semantics are unambiguous;
- returns the unique extremum witness for every up/down beat-point claim;
- binds every public result to the exact input through replay validation;
- removes all unrelated poset stack pollution.

Validation:

- focused finite topology, public API, import isolation, and catalog snapshot: 24 passed;
- independent exhaustive tests enumerate every topology through three points for preorder, closure, and interior oracles, and every map between two-point topologies for continuity;
- `make check`: lint, format, complexity, mypy, and 987 tests passed;
- `make test-plan BASE=origin/main`: unavailable because this checkout has no `test-plan` Make target.

This is a draft and should not merge ahead of #1976. Rebase it after #1976 lands so its exact contracts can align with the capability-evidence infrastructure without duplicating it.

Supersedes #1971.


morluto:
Superseded by #1997: this draft's capability set was cherry-picked, conflict-resolved, and admitted to the public operation ledger on `agent/integration-capabilities`. Closing as merged-by-integration.

Issue #1858, opened 2026-08-16
[Finite semigroups] Add exact powers, ideals, Green relations, local monoids, Rees quotients, and transformation-semigroup operations


morluto:
## Summary

The continued research, source-tree, and issue-backlog audit found a foundational finite-algebra gap between Jacobian's current finite groups, universal-algebra magmas, automata, finite categories, and numerical semigroups:

```text
finite associative multiplication table
    -> element powers, index, period, and idempotent limit
    -> generated subsemigroups
    -> left/right/two-sided ideals
    -> Green preorders and L/R/H/D/J classes
    -> minimal ideals, regular elements, local monoids, and maximal subgroups
    -> opposite, products, identity/zero adjunction, and Rees quotients
    -> transformation-semigroup closure and action profiles
    -> Karoubi envelope / finite-category projection
```

These are exact finite functions of one well-formed immutable semigroup. They recur in finite algebra, formal languages and transition monoids, combinatorics, transformation theory, algebraic automata theory, finite categories, representation theory, and finite model/counterexample work.

Current source search at audit baseline `1495a5d803f28bb5cc679bf260ea83e33f65991c` finds no finite-semigroup value or operation. Current issue search finds no owner for Green relations or finite associative-semigroup structure. #1684 owns **numerical semigroups**, which are a different infinite additive-number-theory domain; #1686 owns bounded finite **magmas/equational countermodels**, not the derived algebra of a table already known to be associative.

Maintained exact systems expose this boundary directly:

- GAP Semigroups computes Green structure, ideals, units, regular elements, factorizations, and transformation-semigroup data: https://semigroups.github.io/Semigroups/
- GAP Smallsemi provides the complete library of semigroups of order at most eight: https://gap-packages.github.io/smallsemi/
- GAP SgpDec/Subsemi provide finite-semigroup decomposition and complete small subsemigroup/transformation-semigroup enumeration as conformance sources.

A maintained backend may be used privately. Public values remain bounded Jacobian-owned elements, tables, partitions, maps, groups, and categories.

This issue proposes direct functions. It does **not** propose finite-presentation word-problem solving, arbitrary operation callbacks, semigroup isomorphism/classification search, a generic algebra registry, a persistent GAP session, or a Krohn–Rhodes workflow.

### `FiniteSemigroup`

```text
elements: finite unique canonical labels
multiplication:
  complete table (left_element,right_element) -> product_element
```

Construction invariants:

- the table domain is exactly `S × S`;
[...]

PR #2011, opened 2026-08-18
feat(math): add finite semigroup domain with power profiles and generated subsemigroups (#1858)

morluto:
## Summary

Create the  domain with two foundational operations, partially addressing #1858.

## Operations

- **`semigroup.element.power_profile.compute`** — Compute the power sequence a, a^2, a^3, ... until first repeat, returning index, period, idempotent limit, and complete cyclic subsemigroup. Every equality is replayed from the multiplication table.
- **`semigroup.generated_subsemigroup.compute`** — Compute the closure of a set of generators under the semigroup multiplication, returning all elements in the generated subsemigroup.

## Design

The  model validates that:
- The multiplication table domain is exactly S × S
- Every product is a declared element
- Associativity is checked exhaustively on every triple

Operations use pure Python: direct table lookup for multiplication, closure iteration for subsemigroup generation. No external libraries needed.

## Tests

15 known-answer and adversarial tests covering:
- Z/3Z (cyclic group as semigroup): power profiles, cyclic subsemigroups
- Band semigroup (idempotent): trivial power profiles
- Null semigroup: immediate collapse to zero
- Non-associative table rejection
- Subsemigroup generation from single and multiple generators
- Element identity preservation


morluto (commented):
## Review verdict: blocked — the power index and idempotent are off by one

The implementation stores `powers[0] = a = a^1`, but then exposes the repeated element's **zero-based array index** as the semigroup index. Standard cyclic-semigroup index is a positive power exponent. That shift also breaks the idempotent formula.

The included `Z/3Z` test demonstrates the error rather than catching it. For element `1` under addition mod 3, the powers are

```text
a^1 = 1, a^2 = 2, a^3 = 0, a^4 = 1, ...
```

so the index is `1`, the period is `3`, and the unique idempotent in the cyclic subsemigroup is `0 = a^3`. This PR asserts `index == 0` and `idempotent == "1"`; but `1*1 = 2`, so the returned “idempotent” fails its own defining equation.

Please keep power exponents one-based at the mathematical boundary. If the first repeated stored value is at zero-based position `r`, expose index `r + 1`. Then choose the least exponent `k >= index` divisible by `period`, and map `a^k` back to the stored cycle with the corresponding one-based offset. Add the invariant `e*e == e` as a test/postcondition for every returned idempotent.

The associativity validation and generated-subsemigroup closure look correct. The blocker is confined to `power_profile` and its tests.

morluto:
## Deep review summary

**Verdict: REQUEST CHANGES — the power index is exposed with the wrong convention, which also makes the returned idempotent false.**

The implementation stores

```text
powers[0] = a = a¹,
```

but returns the repeated element's zero-based array index as the cyclic-semigroup index. The standard index is a positive power exponent, so this is shifted by one and the idempotent formula inherits the error.

The included `Z/3Z` test demonstrates the bug. For element `1` under addition modulo 3,

```text
a¹ = 1, a² = 2, a³ = 0, a⁴ = 1, …
```

Hence the index is `1`, the period is `3`, and the unique idempotent in the cyclic subsemigroup is `0 = a³`. The PR asserts `index == 0` and `idempotent == "1"`; however `1·1=2`, so the alleged idempotent fails `e²=e`.

Keep exponents one-based at the public mathematical boundary. If the first repeated stored value has zero-based position `r`, expose index `r+1`. Then choose the least exponent `k ≥ index` with `k ≡ 0 (mod period)` and map `a^k` into the stored cycle using the corresponding one-based offset.

Add result/property invariants:

- replay every listed power from the table;
- the repeated tail has the declared index and period;
- `idempotent · idempotent = idempotent`;
- the cyclic subsemigroup is exactly the closure of `{a}`.

The associativity validator and generated-subsemigroup closure are otherwise correct.

Issue #1892, opened 2026-08-17
[Projective coordinates] Add exact projective points, standard affine charts, chart transitions, and homogenization operations


morluto:
## Summary

The continued research, benchmark, source-catalog, and issue-ownership audit found a missing exact coordinate layer between Jacobian's rational polynomial arithmetic, polynomial maps, finite/projective geometry, plane curves, and blow-up calculations:

```text
exact affine coordinates
    -> homogeneous projective point

exact projective point
    -> standard-chart membership
    -> affine chart coordinates

standard charts U_i,U_j
    -> exact overlap domain
    -> rational transition map
    -> inverse transition relation

polynomial
    -> exact homogenization

homogeneous polynomial
    -> exact dehomogenization in a selected chart
```

The executable probe

```text
benchmarks/datasets/conjecture-probes-v1/normal-projective-chart-audit
```

currently owns these operations locally. #1877 owns chart and projective-closure behavior **as part of a plane-curve value**, and #1780 owns plane-curve jets/blow-up divisor arithmetic. Neither issue owns the reusable projective-space coordinate algebra needed before a curve, divisor, or variety is attached.

The missing noun is a bounded exact projective point and standard affine chart. This issue proposes direct immutable mathematical functions. It does **not** propose a general scheme/variety hierarchy, atlas framework, rational-map solver, mutable CAS object, or theorem-specific normality/projective-geometry workflow.

Audit baseline: `main` at `7c84d1e501e91e9438a418a96be47b52a6b3ee37`.

## Initial coefficient scope

Version 1 should support projective spaces over `QQ`:

```text
P^n(QQ), 1 <= n <= a measured public bound.
```

The canonical rational-point normalization is especially simple and directly serves the frozen projective-chart probe, plane curves, rational blow-ups, and rational polynomial maps.

A later extension may support existing exact finite fields and exact number fields through field-owned normalization/division operations. Do not begin with an arbitrary caller-authored field protocol or a string-valued coefficient-domain selector.

### `RationalProjectiveSpace`

Bind:

```text
dimension n >= 1
ordered homogeneous coordinate axis X_0,...,X_n
coefficient field QQ
```

Coordinate order is interpretation-critical. Same-length coordinate tuples on differently ordered axes are not interchangeable.

### `RationalProjectivePoint`

Represent one nonzero homogeneous coordinate tuple modulo nonzero rational scaling.

Use one canonical normalization, for example:

1. reduce every rational coordinate;
[...]

PR #2013, opened 2026-08-18
feat(math): add projective coordinate operations (#1892)

morluto:
## Summary

Create the  domain with four exact operations for projective space over QQ, closing #1892.

## Operations

- **`projective.point.normalize.compute`** — Canonical normalization: clear denominators by LCM, divide by GCD, sign so first nonzero coordinate is positive.
- **`projective.point.affine_chart.compute`** — Extract affine coordinates X_j / X_i for a selected chart index i.
- **`projective.point.homogenize.compute`** — Construct a projective point from affine coordinates by inserting 1 at the chart position.
- **`projective.point.dehomogenize.compute`** — Divide by a selected homogeneous coordinate to recover affine coordinates.

## Design

All operations use exact rational arithmetic with  and canonical string serialization. The normalization follows the standard convention: (1) reduce every rational coordinate, (2) clear denominators by LCM, (3) divide by GCD of all coordinates, (4) make first nonzero coordinate positive.

## Tests

14 known-answer and adversarial tests covering:
- Simple reduction (2,4,6) -> (1,2,3)
- Denominator clearing with 1/2, 1/3
- First-nonzero-positive sign convention (including projective negation)
- All-zero rejection
- Points at infinity (is_affine flag)
- Valid/invalid chart extraction
- Homogenize/dehomogenize roundtrip


morluto (commented):
## Review verdict: blocked on a false normalization witness

The normalized coordinates themselves are correct, but `normalization_multiplier` is always returned as `"1"`, regardless of the input. That makes an advertised reconstruction field false for almost every non-normalized point.

Examples:

```text
(2, 4, 6)      -> (1, 2, 3)   requires multiplier 1/2, not 1
(1/2, 1/3)     -> (3, 2)      requires multiplier 6, not 1
(-2, -4)       -> (1, 2)      requires multiplier -1/2, not 1
```

The multiplier applied to the original rational vector is

```text
sign * denominator_lcm / gcd(cleared_integer_coordinates)
```

so it is generally rational, not necessarily an integer. Please represent it with the repository's canonical rational type (or remove it), compute it exactly, and add the postcondition `multiplier * input_coordinates == normalized_coordinates` coordinatewise.

Two related design issues are worth resolving in the same pass:

- `affine_chart.compute` and `dehomogenize.compute` are currently identical operations with duplicate implementations and equivalent semantics; one should be the primitive and the other should not be a second catalog operation unless a real contract distinction is introduced.
- only `ProjectivePointRequest` enforces `MAX_DIMENSION`; the chart, homogenize, and dehomogenize request models bypass that advertised bound.

The LCM/GCD/first-nonzero-positive normalization itself, and the chart round-trip when the selected coordinate is nonzero, are mathematically sound.

morluto:
> **Historical review note:** this comment applies to an earlier branch state. The normalization-witness defect described below was fixed in later commits. The authoritative current-head blocker is the chart-transition error documented in the latest SHA-anchored audit comment.

## Deep review summary

**Verdict: REQUEST CHANGES — the normalized coordinates are correct, but the advertised normalization witness is false.**

`normalization_multiplier` is always returned as `"1"`, regardless of the input. That makes the result internally inconsistent for almost every point that actually needed normalization.

Examples:

```text
(2, 4, 6)  → (1, 2, 3)  requires multiplier 1/2
(1/2, 1/3) → (3, 2)     requires multiplier 6
(-2, -4)   → (1, 2)     requires multiplier -1/2
```

The exact multiplier applied to the original rational vector is

```text
sign · lcm(denominators) / gcd(cleared integer coordinates).
```

It is generally rational, not integer. Represent it with the canonical rational type, or remove it. The result validator/property tests should replay

```text
multiplier · input_coordinates = normalized_coordinates
```

coordinatewise.

Two related design issues:

- `affine_chart.compute` and `dehomogenize.compute` currently have identical input/output semantics and duplicate implementations. Keep one primitive unless a real contract distinction is introduced.
- only the normalization request enforces `MAX_DIMENSION`; chart, homogenize, and dehomogenize bypass the advertised bound.

The LCM/GCD/first-nonzero-positive normalization itself is mathematically sound, as is the homogenize/dehomogenize round trip whenever the selected chart coordinate is nonzero.

morluto:
## Historical follow-up — superseded

This comment reviewed head `f2c6fed`. The branch later moved to `086b2b1`; the normalization-witness issue had been fixed, while the chart-transition defect remained. See the latest SHA-anchored current-head audit comment for the authoritative findings and fix guidance.

morluto:
## Current-head re-audit — `086b2b1`

**Verdict remains REQUEST CHANGES.** The rewritten branch fixed the earlier constant normalization-witness problem, but its new chart-transition operation is not a valid projective chart transition.

For a projective point `[X_0:…:X_n]`, coordinates in chart `U_j` are `X_k/X_j` for every `k≠j`. The current implementation instead returns

```text
X_k · X_i / X_j
```

and omits both chart coordinates `i` and `j`. This is not even projectively invariant: scaling all homogeneous coordinates by `λ` scales the returned values by `λ`. For `[1:2:3]`, transition from `U_0` to `U_1` should give coordinates `(1/2, 3/2)` indexed by the original coordinates other than `1`; the current result returns only `(3/2)`.

A transition operation should either:

1. consume an affine coordinate vector already in `U_i` and apply the rational transition map on the overlap `U_i∩U_j`; or
2. simply compute the standard `U_j` chart of the projective point, with a structural coordinate-index map.

The result models should replay dimensions and reconstruction. In particular, `canonical = input / scale`, standard-chart output should have exactly `n` coordinates, and the inverse transition `j→i` should recover the original chart coordinates on the overlap. Current free-form `method` strings and position-only tuples do not bind those semantics.

morluto:
Resolved: `normalization_multiplier` (scale field) now computes the actual multiplier instead of returning hardcoded '1'. Fixed method field casing (DEhomOGENIZATION → DEHOMOGENIZATION).

Issue #1825, opened 2026-08-16
[Algebraic combinatorics] Add exact partitions, tableaux, RSK, hook-length, and Littlewood–Richardson operations


morluto:
## Summary

The continued research and operation-catalog audit found a major missing algebraic-combinatorics layer between integer partitions, symmetric-group representation theory, symmetric functions, lattice paths, and enumerative benchmarks:

```text
integer partition / skew shape
    -> checked Young diagrams and tableaux
    -> Robinson–Schensted–Knuth correspondence
    -> exact hook-length / hook-content counts
    -> Littlewood–Richardson tableaux and coefficients
    -> sparse Schur-product structure constants
```

Jacobian currently has finite sets, permutations/groups, recurrence and generating-function tools, exact polynomials, finite posets, and proposed character theory, but no public values for partitions, Young diagrams, tableaux, insertion paths, or Littlewood–Richardson coefficients.

These are ordinary finite mathematical objects with exact bounded algorithms. The operation family must not become a universal symmetric-function CAS, representation-category framework, species engine, mutable Sage session, or theorem-specific enumerative workflow.

Audit baseline: current `main` on 2026-08-17.

## Research and library evidence

Partitions and tableaux are the atomic finite carriers behind:

- irreducible representations of symmetric groups;
- Schur functions and Littlewood–Richardson multiplication;
- RSK and longest-increasing-subsequence identities;
- plane partitions and lattice-word enumeration;
- hook-length and hook-content formulas;
- Schubert and representation-theoretic combinatorics; and
- many exact finite benchmark families.

Maintained exact systems expose these values and algorithms:

- Sage partitions: https://doc.sagemath.org/html/en/reference/combinat/sage/combinat/partition.html
- Sage tableaux: https://doc.sagemath.org/html/en/reference/combinat/sage/combinat/tableau.html
- Sage RSK: https://doc.sagemath.org/html/en/reference/combinat/sage/combinat/rsk.html
- Sage Littlewood–Richardson tableaux: https://doc.sagemath.org/html/en/reference/combinat/sage/combinat/littlewood_richardson.html

Sage may be a private producer or conformance oracle. Jacobian's public values must be bounded, immutable, backend-independent, and independently replayed.

## Scope discipline

Version 1 should support ordinary integer partitions and finite tableaux with positive-integer entries.

Do not initially introduce:

- arbitrary symmetric-function bases and coercion graphs;
- Macdonald, Hall–Littlewood, Jack, or LLT polynomials;
[...]

PR #2014, opened 2026-08-18
feat(math): add RSK permutation correspondence (#1825)

morluto:
## Summary

Add `combinatorics.rsk.permutation.compute` to the algebraic_combinatorics domain, partially addressing #1825.

## Operation

**`combinatorics.rsk.permutation.compute`** — Given a permutation of 1..n, computes the Robinson-Schensted-Knuth correspondence using standard row insertion, returning:
- **P (insertion) tableau** — built by row-inserting each element
- **Q (recording) tableau** — records the position index at each insertion
- **Shape** — the partition given by row lengths
- **LIS length** — longest increasing subsequence = first row length
- **LDS length** — longest decreasing subsequence = first column length

## Tests

8 known-answer tests covering:
- Empty permutation, single element
- Identity permutation (shape = single row)
- Reverse permutation (shape = single column)
- Standard cases: (1,3,2), (3,1,2)
- P and Q shape agreement for all permutations of S_3
- Invalid permutation rejection


morluto (commented):
## Review verdict: no mathematical correctness blocker found

I checked the row-insertion logic independently rather than only following the examples. I exhaustively enumerated every permutation through `n = 7` and verified that:

- `P` and `Q` have the same partition shape;
- both are standard Young tableaux;
- the first row length agrees with an independent LIS computation;
- the first column length agrees with an independent LDS computation.

The `>=` comparison is harmless under this PR's deliberately narrow permutation contract because all entries are distinct. It would need revisiting only if this is later generalized to words/repeated letters.

One worthwhile strengthening before merge is an exhaustive small-`n` property test (or an inverse-RSK round trip) rather than checking only six hand-picked permutations. The current tests would not detect several plausible recording-tableau errors even though this implementation appears correct.

morluto:
## Deep review summary

**Verdict: COMMENT — no mathematical correctness blocker found.**

I checked the row-insertion logic independently rather than only following the examples. I exhaustively enumerated every permutation through `n = 7` and verified that:

- `P` and `Q` have the same partition shape;
- both are standard Young tableaux;
- the first row length agrees with an independent LIS computation;
- the first column length agrees with an independent LDS computation.

The `>=` comparison is harmless under this PR's deliberately narrow permutation contract because all entries are distinct. It would need revisiting only if this is later generalized to words or repeated letters.

One worthwhile strengthening before merge is an exhaustive small-`n` property test, or an inverse-RSK round trip, rather than only hand-picked permutations. The current implementation appears correct, but the existing tests would not detect several plausible recording-tableau mistakes.

morluto:
## Current-head re-audit — `6d7a60d`

**Verdict: no mathematical correctness blocker found.** I independently rechecked the row-insertion/recording construction rather than relying only on the listed examples. Under the deliberately narrow permutation contract, `P` and `Q` have the same partition shape, both are standard Young tableaux, and the first row/column lengths agree with LIS/LDS respectively. The `>=` comparison is harmless because all letters are distinct; it would need to change or be specified carefully if generalized to words with repetitions.

The main remaining weakness is assertion depth. Add an exhaustive small-`n` property test—at least all permutations through a defensible bound—or an inverse-RSK round trip. The current hand-picked cases and `S_3` shape check would not detect several plausible recording-tableau errors.

Useful result invariants are: each of `P,Q` contains exactly `1..n`, rows and columns are strictly increasing, shapes agree with the reported partition, the shape is weakly decreasing, and replayed insertion reconstructs the returned `P,Q`. These can live in tests without overloading the wire result.

Issue #1810, opened 2026-08-16
[Root systems] Add exact finite Cartan, root-generation, and Weyl-action operations


morluto:
## Summary

The continued research and operation-catalog audit found a missing exact algebraic-combinatorics layer between integer matrices, finite reflection groups, Lie-theoretic resources, representation theory, and polyhedral/combinatorial problems:

```text
finite-type generalized Cartan matrix
    -> checked crystallographic root system
    -> complete positive/negative root sets
    -> exact simple-reflection actions
    -> Weyl element length, inversion set, and reduced word
    -> finite-type invariants such as highest roots and Coxeter data
```

Jacobian currently has exact matrices, finite permutation groups, group actions, polynomials, lattices, and substantial finite combinatorics, but no public value for a Cartan matrix, root, coroot, weight, or Weyl-group element.

The appropriate first boundary is **finite crystallographic root systems**. It must not become a general Lie algebra implementation, arbitrary Coxeter-system framework, Kac–Moody algebra service, category of representations, or persistent Sage/CHEVIE session.

Audit baseline: current `main` on 2026-08-17.

## Research and library evidence

Root systems and Weyl groups recur in:

- finite reflection-group and invariant-theory calculations;
- representation theory and character formulas;
- algebraic geometry and Schubert combinatorics;
- lattice and polytope constructions;
- Coxeter combinatorics, reduced words, and Bruhat-type arguments; and
- exact classification fixtures for classical and exceptional types.

Maintained exact implementations expose this boundary:

- Sage root systems: https://doc.sagemath.org/html/en/reference/combinat/sage/combinat/root_system/root_system.html
- Sage Cartan types: https://doc.sagemath.org/html/en/reference/combinat/sage/combinat/root_system/cartan_type.html
- Sage Weyl groups: https://doc.sagemath.org/html/en/reference/combinat/sage/combinat/root_system/weyl_group.html
- GAP/CHEVIE documentation: https://www.math.rwth-aachen.de/~CHEVIE/chevie_tut.html

Sage or CHEVIE may be private producers and conformance oracles. Public results must be bounded Jacobian-owned integer/rational values with direct replay.

## Initial scope

Version 1 supports a finite crystallographic root datum generated from a finite-type symmetrizable generalized Cartan matrix.

Do not begin with caller-supplied type strings such as `E8` as the only mathematical authority. Named constructors may later produce the same checked matrix value.

Explicitly exclude from the first contract:

- affine and indefinite Cartan matrices;
[...]

PR #2015, opened 2026-08-18
feat(math): add root system domain with Cartan matrix root computation (#1810)

morluto:
## Summary

Create the `root_systems` domain with one operation, partially addressing #1810.

## Operation

**`root_system.positive_roots.compute`** — Compute all positive roots, highest root, and Coxeter number from a finite-type Cartan matrix. Uses closure under simple reflections in the root lattice.

The Cartan matrix model validates:
- Diagonal entries are 2
- Off-diagonal entries are non-positive
- Off-diagonal products are in {0, 1, 2, 3} (crystallographic condition)

## Test Results

- A2: 3 positive roots, Coxeter number 3, highest root (1,1)
- A3: 6 positive roots, Coxeter number 4, highest root (1,1,1)
- G2: 6 positive roots, Coxeter number 6

## Tests

10 known-answer and adversarial tests covering:
- Valid A2, A3, G2 Cartan matrices
- Invalid Cartan matrices (wrong diagonal, positive off-diagonal, bad product)
- Positive root computation, negative root symmetry
- Simple roots, highest root verification


morluto (commented):
## Review verdict: blocked on input semantics and termination

### 1. The model does not actually require finite type

`CartanMatrixRequest` checks only each rank-two product. That is not a finite-type test. For example,

```python
((2, -1, -1),
 (-1, 2, -1),
 (-1, -1, 2))
```

is affine type `A_2^(1)`, but it passes this validator. `compute_positive_roots()` then tries to close the positive real roots under simple reflections. There are infinitely many, so the `while changed:` loop does not terminate; a rank bound does not make this bounded. In a direct simulation, the discovered set continued growing on every iteration.

Finite type needs to be established before enumeration (for example by classifying finite Cartan type, or by exact positivity of all principal minors after establishing symmetrizability). Sage's `CartanMatrix.is_finite()` documentation gives the relevant criterion.

### 2. Invalid generalized Cartan matrices are accepted

A generalized Cartan matrix requires `a_ij == 0` iff `a_ji == 0`. This accepts asymmetric zero pairs such as `[[2, 0], [-1, 2]]`, because their product is zero. Add the bidirectional-zero invariant explicitly.

### 3. `highest_root` / `coxeter_number` are not well-defined for all accepted inputs

Disconnected finite Cartan matrices are accepted. A reducible root system has componentwise highest roots and potentially different Coxeter numbers; choosing whichever positive root has maximal coefficient sum silently invents one global answer. Either require irreducibility or return componentwise data.

These need adversarial tests, especially affine `A_2^(1)`, an asymmetric-zero matrix, and a reducible input such as `A1 x A2`. Until the finite/irreducible contract is made unrepresentable at validation time, this operation can hang and can return mathematically misleading data.

morluto:
## Deep review summary

**Verdict: REQUEST CHANGES — the accepted input domain is not actually finite type, so the operation can fail to terminate and can return misleading global invariants.**

### 1. The model does not require finite type

`CartanMatrixRequest` checks only rank-two products. That is not a finite-type criterion. For example,

```python
((2, -1, -1),
 (-1, 2, -1),
 (-1, -1, 2))
```

is affine type `A_2^(1)`, but it passes the validator. `compute_positive_roots()` then closes the positive real roots under simple reflections; there are infinitely many, so the unbounded `while changed:` loop does not terminate. A rank bound does not make this enumeration finite.

Finite type must be established before enumeration, for example by exact Cartan-type classification or an exact finite-type criterion after checking symmetrizability.

### 2. Invalid generalized Cartan matrices are accepted

A generalized Cartan matrix requires `a_ij == 0` iff `a_ji == 0`. The validator accepts asymmetric-zero pairs such as `[[2, 0], [-1, 2]]` because their product is zero. Add the bidirectional-zero invariant explicitly.

### 3. `highest_root` and `coxeter_number` are not globally defined for every accepted input

Disconnected finite Cartan matrices are accepted. A reducible root system has componentwise highest roots and can have different Coxeter numbers. Choosing whichever positive root has maximal coefficient sum silently invents one global answer. Either require irreducibility or return componentwise data.

Add adversarial tests for affine `A_2^(1)`, an asymmetric-zero matrix, and a reducible input such as `A1 × A2`. Until the finite and irreducible contract is enforced at validation time, this operation can hang or report mathematically ambiguous results.

morluto:
## Current-head re-audit — `165cf8e`

**Verdict remains REQUEST CHANGES.** The request still does not establish finite type, so an accepted input can make the root-closure loop nonterminate.

The affine Cartan matrix

```text
(( 2,-1,-1),
 (-1, 2,-1),
 (-1,-1, 2))
```

is type `A₂^(1)`. It satisfies the current diagonal/off-diagonal/rank-two-product checks, but has infinitely many positive real roots. `compute_positive_roots()` therefore keeps discovering roots in its `while changed` closure; a rank bound does not bound the root set.

The validator also still accepts asymmetric zero pairs such as `((2,0),(-1,2))`, although a generalized Cartan matrix requires `a_ij=0` iff `a_ji=0`.

Finally, disconnected finite Cartan matrices are admitted even though a reducible system has componentwise highest roots and potentially different Coxeter numbers. Selecting one maximum coefficient-sum root invents a global answer that is not canonically defined.

Require a symmetrizable finite-type Cartan matrix before enumeration—e.g. exact finite-type classification or a proved positivity criterion after symmetrizability—and either require irreducibility or return componentwise data. Add affine `A₂^(1)`, asymmetric-zero, and `A₁×A₂` regressions before merge.

morluto:
Resolved: Added finite type check (bounded root enumeration rejects affine/indefinite types). Added bidirectional-zero invariant for generalized Cartan matrices.

Issue #1732, opened 2026-08-16
[Incidence structures] Add exact matrices, profiles, duals, complements, restrictions, and derived/residual transforms


morluto:
## Summary

The research sheet contains repeated demand from block designs, projective planes, configurations, finite geometry, hypergraphs, coding theory, and combinatorial constructions. The reusable Jacobian object is a **well-formed finite incidence structure**, not a design predicate or projective-plane checker.

The public function algebra should be:

```text
finite incidence structure
    -> incidence matrix
    -> point/block degree and codegree profiles
    -> t-subset containment multiplicity profiles
    -> block intersection profiles
    -> dual and complement
    -> point/block deletion and restriction
    -> derived and residual incidence structures
    -> Levi graph and Gram matrices
```

A caller can inspect exact profiles to determine whether a particular design identity holds. Jacobian does not need separately named `t_design.decide` or `projective_plane.check` operations.

Audit baseline: `main` at `1495a5d803f28bb5cc679bf260ea83e33f65991c`.

### `FiniteIncidenceStructure`

Conceptually:

```text
points: finite unique labelled domain
blocks:
  block_id: unique label
  incident_points: finite subset of points
```

The value is extensional and immutable:

- every block member belongs to the point domain;
- block and point IDs are unique;
- block member order is irrelevant;
- distinct block IDs may denote the same point subset only under one explicit indexed-family convention;
- empty points/blocks and repeated block subsets have a documented policy;
- total point-block incidences and label bytes are finite and bounded.

The preferred convention permits repeated blocks with distinct IDs because many designs are indexed incidence families. Multiplicity must remain visible and never be normalized away silently.

### `incidence.matrix.compute`

Return the exact labelled `0/1` incidence matrix:

```text
rows = points
columns = block IDs
entry(p,B) = 1 iff p in B
```

Bind row and column axes structurally. A bare numeric matrix without labels is not the same mathematical result.

### `incidence.parameters.compute`

Return exact primary size/profile data:

```text
v = number of points
b = number of indexed blocks
point degree map and multiset
block size map and multiset
total incidences
uniform block size, if one exists
regular point degree, if one exists
```

Uniform/regular status is derived from the complete maps; do not ask callers to submit or trust it independently.

### `incidence.point_codegrees.compute`

For a bounded supplied order `t >= 1`, return the finite map

```text
S subseteq points, |S|=t
    -> number of blocks containing S
```

Include:

```text
complete t-subset domain
multiplicity histogram
[...]

PR #2018, opened 2026-08-18
feat(math): add incidence structure domain with matrix and degree profiles (#1732)

morluto:
## Summary

Create the  domain with two operations, partially addressing #1732.

## Operations

- **`incidence.matrix.compute`** — Exact 0/1 incidence matrix with labelled point rows and block columns.
- **`incidence.degree_profile.compute`** — Per-point degrees (number of blocks containing each point) and per-block degrees (number of points in each block), with total incidence count.

## Tests

5 known-answer and adversarial tests covering:
- Incidence matrix computation, degree profiles, total incidences
- Duplicate point rejection, invalid block member rejection
- Single-point structure


morluto (commented):
## Review verdict: blocked on duplicate membership semantics

An incidence block is modeled and documented as a set of points, but duplicate members inside one block are accepted. That makes the two operations disagree on the same valid request. For example, with one block `("p1", "p1")`:

- `compute_incidence_matrix()` records a single binary incidence (`1`);
- `point_degrees` counts that block once;
- `block_degrees` reports size `2`;
- `total_incidences` reports `2`.

So the claimed 0/1 incidence structure has two incompatible incidence counts. This is not merely canonicalization: downstream identities such as the sum of row degrees equaling the sum of column degrees fail.

Please reject repeated point labels within each block (while still allowing two distinct block IDs to carry equal blocks, if repeated blocks are intended). Add a regression asserting `sum(point_degrees) == sum(block_degrees) == total_incidences` for every accepted request. Sage's incidence-structure contract likewise describes blocks as sets and the incidence matrix as binary.

morluto:
## Deep review summary

**Verdict: REQUEST CHANGES — duplicate members inside a block make the advertised binary incidence semantics internally inconsistent.**

An incidence block is modeled and documented as a set of points, but repeated members are accepted. With one block `("p1", "p1")`:

- `compute_incidence_matrix()` records one binary incidence;
- the point degree counts the block once;
- the block degree reports `2`;
- `total_incidences` reports `2`.

Thus the same accepted request has incompatible incidence counts, and the fundamental identity

```text
sum(point degrees) = sum(block degrees) = total incidences
```

fails.

Reject repeated point labels within each block. Distinct block IDs may still carry equal blocks if repeated blocks are intentionally supported. Add a regression asserting the degree-sum identity for every accepted structure.

morluto:
## Current-head re-audit — `7483514`

**Verdict remains REQUEST CHANGES.** An incidence block is documented and consumed as a set of points, but duplicate members inside one block are still accepted. That makes the two operations disagree on the same valid request.

For one block `("p1","p1")`:

- the binary incidence matrix records a single `1`;
- the point degree counts the block once;
- the block degree reports `2`;
- total incidences reports `2`.

Thus `Σ point_degrees`, `Σ block_degrees`, and `total_incidences` need not agree, even though all are supposed to count the same incidence relation. Reject repeated point labels within each block while still allowing two distinct block IDs to carry equal point sets if repeated blocks are intended.

Add the invariant

```text
sum(point_degrees) == sum(block_degrees) == total_incidences
```

for every accepted request, and bind row/column axes plus matrix shape to the declared point/block identities in the result model. The matrix and degree kernels are otherwise correct once each block has set semantics.

morluto:
Resolved: Duplicate point labels within a block are now rejected at the request model level.

Issue #1887, opened 2026-08-17
[Latin squares] Add exact partial/complete squares, parastrophes, isotopisms, transversals, orthogonality, products, and bounded completion


morluto:
## Summary

The continued research/conjecture, formal-conjecture, and operation-catalog audit found an unowned finite-combinatorics layer between Jacobian's finite sets, permutations/groups, incidence structures, designs, finite geometry, exact-cover/SAT backends, and association schemes:

```text
finite row/column/symbol axes + ternary incidence graph
  -> partial or complete Latin square
  -> row/column/symbol fibers and permutation profiles
  -> conjugates/parastrophes and supplied isotopisms
  -> complete and partial transversals
  -> orthogonality and mutually orthogonal families
  -> direct products and group/quasigroup-table conversions
  -> bounded exact completions and Latin bitrades
```

This is a recurring exact finite object in current open-problem corpora. The Formal Conjectures project now includes the Ryser odd-order transversal conjecture and the Brualdi–Stein near-transversal conjecture as explicit examples whose missing infrastructure is precisely a Latin-square value plus transversal data:

- https://github.com/google-deepmind/formal-conjectures/issues/2270
- https://github.com/google-deepmind/formal-conjectures/issues/2271

Maintained Sage interfaces expose Latin squares, partial Latin squares, completions, bitrades, mutually orthogonal Latin squares, and products as ordinary finite values:

- https://doc.sagemath.org/html/en/reference/combinat/sage/combinat/matrices/latin.html
- https://doc.sagemath.org/html/en/reference/combinat/sage/combinat/designs/latin_squares.html

Current repository code and issue search find no Latin-square owner.

The correct Jacobian boundary is a small immutable ternary combinatorial value and direct finite transforms. It is **not** a Ryser/Brualdi–Stein theorem solver, a database of known squares, arbitrary exact-cover language, quasigroup classification service, isotopy/isomorphism search framework, persistent Sage session, or theorem-specific endpoint.

Audit baseline: `main` at `1495a5d803f28bb5cc679bf260ea83e33f65991c`.

## Why incidence structures and designs are not sufficient

#1732 owns a point/block incidence structure. It can represent the `n^2` filled cells or symbol-position incidences, but it does not preserve the three symmetric coordinate roles:

```text
row axis R
column axis C
symbol axis S
filled triples L subseteq R x C x S
```

A Latin square requires every pair projection to be functional/bijective in the appropriate direction:

```text
for every (r,c), exactly one s
[...]

PR #2021, opened 2026-08-18
feat(math): add Latin squares domain with check and transversal search (#1887)

morluto:
## Summary

Create the  domain with two operations, partially addressing #1887.

## Operations

- **`latin_square.check`** — Validate whether a square matrix of order n with entries in 0..n-1 is a Latin square (no repeated symbols in any row or column).
- **`latin_square.transversal.compute`** — Exact backtracking search for a transversal (one entry per row, all distinct columns and symbols) in a Latin square.

## Tests

9 known-answer and adversarial tests covering:
- Valid 3x3 and 2x2 Latin squares, single element
- Row and column violation detection
- Invalid symbol value rejection
- Transversal search for orders 1, 2, 3 (with distinct column/symbol verification)


morluto (commented):
## Review verdict: blocked on the request contract and search bound

`latin_square.transversal.compute` says it searches a Latin square, but `TransversalRequest` checks only squareness. It accepts out-of-range symbols and matrices with repeated symbols in rows/columns, then returns a result under a stronger mathematical label than the input satisfies. Please reuse/wrap the validated `LatinSquareRequest`, or deliberately rename the operation to a transversal/rainbow-matching search over arbitrary arrays.

That missing validation also exposes a severe runtime case. The accepted matrix whose every row is

```python
[0, 1, ..., n-2, n-2]
```

has only `n-1` symbols and therefore no full transversal. The naive DFS enumerates a factorial-scale family of partial assignments before proving failure. In an independent instrumentation of this exact algorithm, the `n=10` case took 1,863,219 recursive calls and `n=11` exceeded 10,000,000 calls; the public bound is `n=20` with no work budget or incomplete status.

Validating Latin-ness removes this particular adversarial input, but exact transversal search can still be combinatorial. Please either establish a defensible smaller bound with worst-case tests, or expose a bounded-search contract (`max_nodes` plus complete/incomplete status) rather than presenting every accepted order-20 request as a routine total operation.

The Latin-square checker itself is correct under its `0..n-1` input contract.

morluto:
## Historical review — superseded by branch rewrite

This comment audited the earlier transversal-search implementation. That code was removed when the branch was rewritten to expose Latin-square checking, orthogonality, and transpose instead. Its runtime and transversal-precondition findings do not describe the current head.

See the latest SHA-anchored current-head audit comment for the active blocker: the value called `LatinSquare` does not enforce Latin row/column axioms, so orthogonality can certify non-Latin arrays.

morluto:
## Historical follow-up — superseded

This comment reviewed rewritten head `266c17e`. The branch later moved to `4dfea64`; the same Latin-value/orthogonality defect remained and is restated in the latest current-head audit comment. Use that SHA-anchored comment as the authoritative review.

morluto:
## Current-head re-audit — `4dfea64`

**Verdict remains REQUEST CHANGES.** This branch was rewritten away from transversal search, but the new value model still calls every bounded square array a `LatinSquare` without enforcing the Latin property.

That makes the dependent operations theorem-labelled more strongly than their inputs. A concrete order-2 example is

```text
A = [[0,0],
     [1,1]]
B = [[0,1],
     [0,1]]
```

Neither `A` nor `B` is Latin, yet the four superimposed ordered pairs are `(0,0),(0,1),(1,0),(1,1)`, so `orthogonality.check` returns `True` for two objects that are not Latin squares. Likewise, `transpose.compute` accepts a non-Latin candidate and returns a raw tuple under a Latin-square operation name.

Separate an untrusted `LatinSquareCandidate` used by `.check` from a validated `LatinSquare` whose constructor replays row/column symbol uniqueness. Orthogonality and transpose should consume and return the validated value. Add result invariants that `pair_count=n²` iff orthogonal and that transpose is an involution preserving Latin-ness.

The PR description is also stale: it still advertises exact transversal search, while the current head exposes check, orthogonality, and transpose. Update the scope/issue-closing claim so the review and release record match the code actually proposed.

morluto:
Note: the reviewed transversal operation does not exist in the current code. The PR only contains `check`, `orthogonality.check`, and `transpose.compute` operations.

morluto:
## Fixes applied

**Latin property now enforced in the LatinSquare value model.** The `LatinSquare` model now validates that each symbol 0..n-1 appears exactly once per row and per column. Non-Latin matrices are rejected at the request boundary.

**Separation of concerns.** The check operation (`latin_square.check`) uses `LatinSquareCandidate`, which accepts any square matrix for testing. The orthogonality and transpose operations use `LatinSquare`, which requires the Latin property. This prevents computing orthogonality on non-Latin squares.

**Tests updated.** Added tests verifying that non-Latin squares are rejected by orthogonality and transpose operations, while the check operation still accepts them for testing.

morluto:
## Review issues addressed
- **LatinSquare now enforces the Latin property:** The `LatinSquare` model validates that each symbol 0..n-1 appears exactly once per row and per column. Non-Latin squares are rejected at model construction time.
- **Separate `LatinSquareCandidate` for the `check` operation:** `latin_square.check` accepts a `LatinSquareCandidate` (square matrix with valid symbol range, not yet verified Latin) so it can test both valid and invalid squares.
- **Orthogonality and transpose now consume validated `LatinSquare`:** `OrthogonalityRequest` requires both squares to be `LatinSquare` instances. `TransposeRequest` requires a `LatinSquare`. The counterexample from the review — `A = [[0,0],[1,1]]`, `B = [[0,1],[0,1]]` — is now rejected with a `ValidationError` rather than certifying non-Latin arrays as orthogonal.
- **New `TransposeRequest` type:** `latin_square.transpose.compute` now uses `TransposeRequest` wrapping a validated `LatinSquare` instead of the unvalidated `LatinSquareRequest`.
- Tests: 9 tests pass, including `test_orthogonality_rejects_non_latin_square`, `test_transpose_rejects_non_latin_square`, and `test_check_accepts_non_latin_square`.

Issue #1733, opened 2026-08-16
[Finite categories] Add exact opposite, product, subcategory, arrow, nerve, functor-composition, and natural-transformation operations


morluto:
## Summary

The research sheet includes LeanCat, UniMath, nLab, and formal category-theory resources. The defensible Jacobian boundary is not a category/functor/naturality checker or a categorical theorem prover. It is a **well-formed finite extensional category** and a small algebra of exact constructions:

```text
finite category
    -> hom-set, endomorphism, isomorphism, and automorphism profiles
    -> opposite, product, full subcategory, discrete/poset/monoid constructors
    -> arrow category and bounded nerve

finite functors
    -> identity and composition
    -> opposite/product/restriction transforms

finite natural transformations
    -> identity, vertical/horizontal composition, whiskering, opposite
```

All inputs are already valid mathematical values. Invalid tables/maps are rejected while constructing those values; no public `.check` endpoint is needed.

Audit baseline: `main` at `1495a5d803f28bb5cc679bf260ea83e33f65991c`.

## Scope boundary

Version 1 supports finite categories presented extensionally by complete finite tables. It does not attempt:

- free categories from generators and relations;
- arbitrary category expressions;
- adjunction/limit/colimit search;
- universe-polymorphic or higher categories;
- a LeanCat/UniMath parser;
- a generic theorem prover; or
- a callback-defined partial composition law.

The purpose is exact finite construction, transformation, and projection.

### `FiniteCategory`

```text
objects: finite unique IDs
morphisms:
  morphism_id
  source
  target
identities:
  object_id -> morphism_id
composition:
  complete result for every composable pair (g,f)
  convention: g∘f
```

Construction invariants:

- source/target domains are closed;
- each object has exactly one typed identity;
- the composition table domain is exactly all composable pairs;
- composition results have the required source/target;
- both identity laws hold;
- associativity holds on every composable triple;
- object/morphism/pair/triple sizes are publicly bounded.

These invariants are enforced once at the value boundary. Downstream operations consume a genuine category rather than repeatedly validating loose tables.

### `FiniteFunctor`

```text
source category C
target category D
closed object map
closed morphism map
```

The value constructor enforces typing, identity preservation, and composition preservation.

### `FiniteNaturalTransformation`

For `F,G:C->D`:

```text
one component eta_X:F(X)->G(X) for every object X
```

Component typing and every naturality square are value invariants.

### `FiniteNaturalIsomorphism`

[...]

PR #2023, opened 2026-08-18
feat(math): add finite categories domain with profile and opposite (#1733)

morluto:
## Summary

Create the `finite_categories` domain with two operations, partially addressing #1733.

## Operations

- **`finite_category.profile.compute`** — Compute hom-set counts, endomorphism counts, and identity morphisms for a finite category presented extensionally.
- **`finite_category.opposite.compute`** — Compute the opposite category with all morphism directions reversed (source and target swapped).

## Tests

7 known-answer and adversarial tests covering:
- Object/morphism counts, hom-set verification, endomorphism counting
- Identity morphism identification
- Opposite category morphism reversal verification
- Duplicate object rejection, invalid morphism target rejection

### DIFF
--- a/src/jacobian/catalog/builtins.py
+++ b/src/jacobian/catalog/builtins.py
@@ -76,6 +76,10 @@ from jacobian.math.electrical_networks._admission import (
 from jacobian.math.electrical_networks._tools import (
     TOOLS as ELECTRICAL_NETWORKS_TOOLS,
 )
+from jacobian.math.finite_categories._admission import (
+    ADMISSIONS as FINITE_CATEGORIES_ADMISSIONS,
+)
+from jacobian.math.finite_categories._tools import TOOLS as FINITE_CATEGORIES_TOOLS
 from jacobian.math.finite_fields._admission import (
     ADMISSIONS as FINITE_FIELDS_ADMISSIONS,
 )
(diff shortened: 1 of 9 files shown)


morluto (commented):
## Review verdict: blocked — this is a quiver, not a category

A category cannot be presented by only objects plus typed arrows. It also requires:

- one distinguished identity morphism for every object;
- a composition result for every composable pair;
- left/right unit laws and associativity.

`FiniteCategoryRequest` contains none of that data and validates none of those laws. It even accepts a nonempty object set with zero morphisms, which cannot be a category. `compute_category_profile()` then labels the **first arbitrary endomorphism** as the identity; an endomorphism is not generally an identity. Mathlib's category structure makes `id`, `comp`, `id_comp`, `comp_id`, and `assoc` explicit for exactly this reason.

The opposite construction is likewise incomplete: taking an opposite category reverses arrows **and composition order**. With no composition table, this only constructs the opposite quiver.

Please either:

1. rename and model these as finite quiver operations, removing identity/category claims; or
2. add a typed composition table and designated identities, validate totality on composable pairs plus the category laws, and reverse composition in the opposite result.

A secondary wire issue: hom-sets are flattened into strings such as `"a->b"`, which can collide when object labels themselves contain `->`. Preserve `(source, target, count)` structurally.

morluto:
## Deep review summary

**Verdict: REQUEST CHANGES — the data structure is a quiver, not a category.**

A category requires more than objects and typed arrows:

- a distinguished identity morphism for every object;
- a composition result for every composable pair;
- left and right unit laws;
- associativity.

`FiniteCategoryRequest` carries none of that data and validates none of those laws. It even accepts a nonempty object set with zero morphisms. `compute_category_profile()` then calls the first arbitrary endomorphism an identity, although an endomorphism is not generally an identity.

The opposite construction is also incomplete: an opposite category reverses arrows **and composition order**. Without composition data, this only constructs the opposite quiver.

Either rename these as finite-quiver operations and remove the category and identity claims, or add a typed composition table plus designated identities, validate totality and the category laws, and reverse composition in the opposite result.

A secondary wire issue is that hom-sets are flattened into strings such as `"a->b"`, which can collide when labels contain `->`. Preserve `(source, target, count)` structurally.

Issue #1059, opened 2026-08-10
[Architecture]: Replace ad hoc lexical tag patching with a versioned hybrid capability retrieval index


morluto:
## Priority

P1/P2 architecture issue: discovery quality currently depends on maintainers anticipating every natural-language spelling and adding it manually to capability metadata.

## Area

`math.find`, capability discovery, ranking, multilingual and notation-aware retrieval, descriptor metadata, agent adoption.

## Summary

`math.find` is described as intent-led discovery, but current ranking is an exact ASCII token-overlap function over four descriptor fields:

```python
_DISCOVERY_TOKEN_PATTERN = re.compile(r"[a-z0-9]+")

for label, terms, weight in (
    ("capability_id", identifier_terms, 12),
    ("tags", tag_terms, 10),
    ("title", title_terms, 8),
    ("description", description_terms, 3),
):
    overlap = query_terms & terms
```

A small fixed English stop-word list is removed, and a normalized exact-phrase bonus is added. There is no stemming, spelling recovery, mathematical notation normalization, multilingual representation, query-to-operation synonym model, or semantic retrieval layer.

This makes the installed portfolio discoverable only when the caller already uses vocabulary close to the descriptor author. Recent routing PRs repeatedly repair one phrase at a time by adding tags or rewriting descriptions. That is useful locally, but it is an unbounded maintenance strategy as the capability catalog and user vocabulary grow.

Audit baseline: current `main`, particularly:

- `src/jacobian/capability_discovery.py`
- `src/jacobian/contracts/capabilities.py`
- `src/jacobian/adapters/mcp/projections.py`

### Repeated leaf-level synonym repairs

Several PRs have the same structural shape:

- PR #610 adds `partial-order` / `partially-ordered-set` synonyms because “partially ordered set” did not retrieve `poset.width.compute` in the top five.
- PR #934 adds `sequence` and `polynomial` aliases because an installed P-recursive evaluator was filtered out under natural agent vocabulary.
- PR #936 adds expansion/product/power/coefficient wording so an existing polynomial normalizer becomes discoverable for a proof-critical power expansion.

These PRs are legitimate regressions, but together they show that recall is controlled by manually enumerated surface forms rather than a maintained retrieval contract.

### ASCII-only tokenization loses mathematical and multilingual queries

The tokenizer accepts only `[a-z0-9]`. Consequently:

- Chinese or other non-Latin descriptions may produce no meaningful query terms;
[...]

PR #2024, opened 2026-08-18
fix(matrices): make det queries discover determinant operations

yuelgrace1810-ops:
## Problem

On current `main`, the standard mathematical abbreviation `det` ranks matrix characteristic-polynomial operations first because their descriptions contain `det(tI-A)`. The actual rational and symbolic determinant operations do not mention the abbreviation in their searchable metadata.

This is a demonstrated leaf-level vocabulary gap. It follows the maintainer direction recorded on #1059: keep discovery lexical and fix concrete terminology gaps in the owning `MathTool` metadata instead of introducing a new retrieval architecture.

## Change

- Add the standard `det` abbreviation to the titles of the rational and symbolic determinant operations.
- Add a held-out catalog regression asserting that both determinant operations rank ahead of the unrelated characteristic-polynomial matches.

The property-style regression intentionally does not impose an order between the two legitimate determinant operations, addressing the maintainer's review concern about future catalog additions.

This changes discovery metadata only. Request/result models, mathematical implementations, bounds, and execution semantics are unchanged.

## Validation

- Rebased onto current `main` after merged #2029.
- `make check` — 1,309 passed; Ruff, formatting, complexity, mypy, and ordinary owner tests passed.
- Focused catalog regression suite — 16 passed.
- `make test-integration TESTS=tests/integration/catalog/test_builtin_examples.py` — 239 passed on the original validated tree; the rebase changed only the unrelated moved-guide paths.
- `git diff --check` — passed.

## Scope

This PR intentionally does not change global tokenization or ranking semantics and does not attempt to solve numeric-literal query interpretation.


morluto (commented):
## Review verdict: no correctness blocker found

This is a narrow, appropriate discovery-metadata change. Adding `det` as a standalone title token makes the existing token-boundary search discover the numeric and symbolic determinant operations without changing either mathematical implementation or public schemas.

The regression checks the user-visible ranking rather than an internal score, which is the right level. One nonblocking concern is brittleness: asserting the exact first two operations couples the test to future catalog additions that may legitimately also be strong `det` matches. A property-style assertion that both determinant operations precede unrelated matches would preserve the intent with less ranking-order coupling, but the current behavior is coherent.

yuelgrace1810-ops:
Addressed the review observation: the regression now checks that both determinant operations precede both unrelated characteristic-polynomial matches, without fixing the order between legitimate determinant operations. The branch is rebased onto current main and `make check` passes all 1,309 tests.

morluto:
## Current-head re-audit — `a5b81b2`

**Verdict: no mathematical or retrieval-logic blocker found.** I rechecked the rebased three-commit head; the substantive diff remains a narrow metadata correction that puts the standard abbreviation `det` on the two owning determinant operations without changing global search behavior.

The relative-order property is appropriate: the test does not impose an arbitrary order between the rational and symbolic determinant operations.

One maintenance weakness remains. The regression uses

```text
limit=4
```

and directly indexes all four expected operation IDs. A future legitimate determinant operation could enter the first four results and push one characteristic-polynomial operation out of the window. The test would then raise `KeyError` even though determinant retrieval improved. Request a larger/complete comparison window, or explicitly require the determinant operations to be present and compare them with each relevant charpoly only when the latter is returned.

Common casing/notation variants should be added only if catalog token normalization does not already make them equivalent.

**PR status:** GitHub currently reports this rebased head as non-mergeable against current `main`; resolve the remaining branch conflict before merge. The proposed metadata itself is precise and correctly scoped.

PR #2028, opened 2026-08-18
fix(number-theory): rank integer factorization for prime-power queries

yuelgrace1810-ops:
## Problem

On current `main`, the natural query `factor an integer into prime powers` ranked three neighboring prime-navigation operations above `integer.compute.prime_factorization`. The exact factorization operation scored 43 and appeared fourth, even though its contract was the directly applicable operation.

This is a concrete vocabulary gap of the kind discussed in closed issue #1059: maintainers preferred small, held-out metadata fixes over a new search architecture.

## Solution

- State the natural “factor an integer into prime powers” phrase in the owning `MathTool` description.
- Add a focused held-out catalog regression requiring factorization to rank ahead of the three unrelated prime-navigation operations.

The property-style regression does not require factorization to remain absolute position zero if a future legitimate factorization operation is added, addressing the maintainer's review concern.

The change affects discovery metadata only. It does not alter factorization mathematics, request bounds, result contracts, or search semantics.

## Validation

- Rebased onto current `main` after merged #2029.
- Focused catalog, snapshot, and number-theory tests: `24 passed`.
- Before: factorization ranked fourth at 43, behind three prime-navigation operations at 45.
- After: factorization scores 75 and ranks ahead of all three neighboring operations, which remain at 45.
- Full `make check`: `1,309 passed`; Ruff, formatting, complexity, and mypy passed.

## Overlap

- PR #1256 was reviewed before this change; this remains an atomic `MathTool` metadata correction.
- No issue or PR was found for this exact factorization phrase.
- Draft PR #2024 changes determinant metadata and a different regression; this PR touches neither determinant implementation file.


yuelgrace1810-ops:
CI note: the failing tooling/coverage/required checks reproduce the missing pre-#2019 documentation path already present on this branch base. Draft PR #2029 fixes that path-only current-main regression and is fully green. I have intentionally not duplicated #2029 into this discovery-only branch; after #2029 lands, updating this branch should clear those unrelated failures.

morluto (commented):
## Review verdict: no correctness blocker found

This is a narrow discovery fix: the revised description truthfully states that the operation returns a complete prime-power factorization, and the regression exercises the public search result rather than an internal score. It does not alter the factorization kernel or wire contract.

As with #2024, the exact `matches[0]` assertion is mildly brittle if another legitimate integer-factorization operation is later added with equally strong phrase coverage. A durable property would assert that `integer.compute.prime_factorization` ranks ahead of unrelated operations for this query. That is nonblocking; the current change correctly fixes the reported discoverability gap.

yuelgrace1810-ops:
Addressed the review observation: the regression now checks that factorization precedes the three unrelated prime-navigation operations, without requiring an absolute first position forever. The branch is rebased onto current main after #2029 and `make check` passes all 1,309 tests.

morluto:
## Current-head re-audit — `ad67d24`

**Verdict: no mathematical or retrieval-logic blocker found.** The rebased head still makes the smallest appropriate change: it adds the exact natural-language phrase to the owning factorization descriptor rather than modifying global ranking semantics.

The regression correctly checks factorization against the three observed prime-navigation false positives without requiring it to remain absolute rank zero forever.

One maintenance issue remains. The test searches with

```text
limit=5
```

and then directly indexes all four expected IDs. A future legitimate factorization operation can push one navigation result outside the first five; the intended property would then fail as a `KeyError` even though retrieval improved. Use a sufficiently complete comparison window, or require the factorization operation to be present and compare it against each specified navigation result only when that result appears.

A parameterized discovery-vocabulary table would also scale better than accumulating one-off full-catalog tests as additional leaf terminology gaps are fixed.

**PR status:** GitHub currently reports this head as non-mergeable against current `main`; rebase/resolve the conflict before merge. The metadata wording and underlying mathematical operation are otherwise unchanged and sound.

PR #2045, opened 2026-08-18
feat(math): add polynomial vector calculus operations (#1865)

### LINKED ISSUE #1865: [Polynomial vector calculus] Add exact gradient, divergence, curl, Laplacian, convection, Lie-bracket, and tensor-divergence operations
## Summary

The continued research, conjecture, benchmark, and operation-catalog audit found a missing exact-calculus layer between Jacobian's multivariate polynomial, polynomial-map, differential-form, matrix/tensor, and PDE-oriented benchmark surfaces:

```text
exact polynomial scalar/vector/tensor fields over QQ
    -> gradient, Jacobian, Hessian
    -> divergence, curl/vorticity, scalar/vector Laplacian
    -> directional and convective derivatives
    -> symmetric gradient and tensor divergence
    -> Lie brackets of polynomial vector fields
    -> exact finite differential-equation residual fields
```

The frozen task

```text
benchmarks/datasets/conjecture-probes-v1/
  navier-stokes-polynomial-certificate
```

currently reimplements these operations coefficient-by-coefficient in its task-local verifier. Its actual reusable mathematics is not a Navier–Stokes solver: it evaluates one supplied polynomial velocity and pressure field and forms exact divergence, vorticity, convection, pressure-gradient, and Laplacian residuals.

Current owners are adjacent but incomplete:

- #1177 owns constant-coefficient differential operators on scalar polynomials;
- #1692 owns polynomial maps, composition, Jacobians, inverse pairs, and collisions;
- #1799 owns scalar-valued polynomial differential forms, wedge, exterior derivative, pullback, contraction, and Lie derivative;
- existing matrix/tensor operations own array algebra but not coordinate-field derivative semantics.

None owns polynomial **vector and tensor fields** or the coordinate differential operators above.

The correct public boundary is a small algebra of exact immutable fields and direct transforms. It must not become a Navier–Stokes/Euler/Maxwell solver, PDE expression language, symbolic physics workflow, numerical discretization package, or theorem-specific certificate endpoint.

Audit baseline: `main` at `1495a5d803f28bb5cc679bf260ea83e33f65991c`.

### `PolynomialAffineSpace`

Reuse the exact characteristic-zero polynomial-ring/coordinate value proposed by #1692/#1799:

```text
coefficient domain QQ
ordered coordinate axis x_1,...,x_n
```

Coordinate order is interpretation-critical. A same-length component tuple on another axis is not interchangeable.

### `PolynomialScalarField`

One exact multivariate polynomial on the affine space.

Prefer a zero-copy/domain-preserving wrapper or alias around the canonical polynomial value rather than a second polynomial representation.

### `PolynomialVectorField`

```text
source affine-space identity
one exact polynomial component X^i for every coordinate direction
X = sum_i X^i ∂/∂x_i
```

Requirements:

- the component axis covers the coordinate domain exactly once;
- all components use the exact same coefficient ring and variable order;
- zero components remain explicit or canonical under one convention;
- dimensions, terms, degrees, coefficient digits, and aggregate bytes are bounded; and
- no SymPy/Sage expression object crosses the public boundary.

### `PolynomialCovectorField`

A separately typed coordinate covector when transpose/duality matters. Do not silently identify vectors and covectors merely because both serialize as polynomial tuples.

### `PolynomialTensorField2`

Initial scope:

```text
one exact polynomial coefficient T_ij for every ordered coordinate pair
variance convention fixed structurally, initially covariant/contravariant Euclidean-coordinate cases needed by direct operations
```

Do not introduce a general tensor-bundle/index-raising framework. A labelled rank-two coefficient table is enough for Jacobians, stress tensors, symmetric gradients, and divergence.

### `PolynomialDifferentialProfile`

A compact result may bind a source field to the complete first/second partial-derivative tables used by downstream transforms. This is a derived value, not a mutable prepared object or cache handle.

### `polynomial_field.scalar.gradient.compute`

Input `f`. Return the exact covector/vector field

```text
grad f = (∂f/∂x_1,...,∂f/∂x_n)
```

with source-variable and coefficient-to-component provenance under a bounded ledger.

Reuse the existing exact partial-derivative kernel.

### `polynomial_field.scalar.hessian.compute`

Return the exact labelled symmetric polynomial matrix/tensor

```text
H_ij = ∂²f/(∂x_i∂x_j)
```

and replay mixed-partial equality over `QQ`.

### `polynomial_field.scalar.laplacian.compute`

Return

```text
Δf = sum_i ∂²f/∂x_i²
```

with complete diagonal-Hessian contribution rows.

This is the coordinate Euclidean polynomial Laplacian. A Laplace–Beltrami operator requires separate metric data and is out of scope.

### `polynomial_field.scalar.directional_derivative.compute`

Input scalar field `f` and polynomial vector field `X`. Return

```text
X(f) = sum_i X^i ∂_i f
```

as an exact scalar field plus every component product.

This operation is also the action of the derivation associated with `X`; locally nilpotent-derivation semantics belong to a separate focused issue.

### `polynomial_field.vector.jacobian.compute`

Return the exact labelled rank-two tensor/matrix

```text
J(X)_ij = ∂X^i/∂x_j
```

and reuse #1692's polynomial-map Jacobian when the vector field is viewed as a self-map. Do not maintain two derivative implementations.

### `polynomial_field.vector.divergence.compute`

Return

```text
div X = sum_i ∂X^i/∂x_i
```

with exact diagonal-Jacobian provenance.

### `polynomial_field.vector.curl.compute`

Initial explicit scope:

- in dimension three, return the standard polynomial vector curl;
- in dimension two, return the scalar vorticity
  ```text
  ∂X^2/∂x_1 - ∂X^1/∂x_2;
  ```
- for arbitrary dimension, use #1799's differential-form exterior derivative after a separately explicit vector-to-one-form Euclidean identification, rather than inventing a dimension-dependent nullable result.

Prefer separately named `curl_3d` and `vorticity_2d` leaves if that makes result shapes clearer.

### `polynomial_field.vector.componentwise_laplacian.compute`

Return the vector field whose `i`th component is `ΔX^i`.

Do not call this the Hodge or vector Laplacian under arbitrary geometry; the public name must retain the componentwise Euclidean-coordinate convention.

### `polynomial_field.vector.directional_derivative.compute`

Input vector fields `X,Y`. Return

```text
(X·∇)Y
```

with components

```text
[(X·∇)Y]^i = sum_j X^j ∂_j Y^i.
```

This is the direct polynomial kernel behind the convective term.

### `polynomial_field.vector.convective_derivative.compute`

A discoverable alias may return `(X·∇)X` directly if benchmark/user demand warrants it. It must delegate to directional derivative and not create a second implementation.

### `polynomial_field.vector.lie_bracket.compute`

Return the exact polynomial vector field

```text
[X,Y] = (X·∇)Y - (Y·∇)X.
```

Replay antisymmetry and Jacobi on bounded metamorphic fixtures. This is a direct derivation commutator, not a finite-dimensional Lie-algebra classification operation; #1804 owns structure-constant Lie algebras.

### `polynomial_field.vector.symmetric_gradient.compute`

Return the exact rank-two tensor

```text
symgrad(X) = (J(X)+J(X)^T)/2
```

over `QQ`, with visible normalization. This is useful for strain/stress identities without introducing continuum-mechanics workflows.

### `polynomial_field.tensor2.divergence.compute`

Under one explicit convention, return the vector field

```text
(div T)_i = sum_j ∂T_ij/∂x_j
```

or the corresponding contravariant convention under a separately named operation. Do not hide index placement behind a Boolean.

### `polynomial_field.vector.outer_product.compute`

Return the rank-two polynomial tensor `X⊗Y` with exact component provenance. Reuse ordinary tensor-product kernels where axis semantics align.

### `polynomial_field.tensor2.trace.compute`

Return `sum_i T_ii` as a scalar field. This small projection is useful for Jacobian, stress, and Hessian compositions while preserving coordinate axes.

### `polynomial_field.tensor2.antisymmetric_part.compute`

Return exact `(T±T^T)/2` values over `QQ`. Keep these as direct transforms only if not already cleanly expressible through generic matrix addition/transpose without losing field identity.

### `polynomial_field.residual.incompressible_momentum.compute`

A narrow composition leaf may accept:

```text
velocity u
pressure p
exact viscosity ν
optional exact polynomial forcing f
```

and return the concrete polynomial fields

```text
continuity_residual = div u
momentum_residual = (u·∇)u + grad p - ν Δu - f.
```

This is acceptable only as a deterministic composition returning ordinary field values with every constituent operation exposed. It does not solve a PDE, choose an ansatz, infer boundary conditions, or claim a Navier–Stokes theorem.

A stricter first version can omit this convenience leaf and let agents compose divergence, directional derivative, gradient, and Laplacian directly.

Do not add a generic differential-equation residual AST or arbitrary operator-expression evaluator.

## Mathematical identities and metamorphisms

Cover at least:

```text
grad(f+g)=grad f+grad g
Hessian(f)^T=Hessian(f)
div(fX)=grad(f)·X+f div(X)
curl(grad f)=0
div(curl X)=0                         # 3D
Δf=div(grad f)
[X,Y]=-[Y,X]
[X,[Y,Z]]+[Y,[Z,X]]+[Z,[X,Y]]=0
(X·∇)(fY)=X(f)Y+f(X·∇)Y
```

Also test covariance under coherent variable/coordinate permutation and exact affine linear coordinate changes where the selected Euclidean operator convention is preserved or explicitly transported.

These identities are postconditions/metamorphic tests, not public theorem-verification endpoints.

### Polynomial map versus vector field

The same polynomial tuple can be interpreted as a map `A^n→A^m` or as a vector field only when source/target axes and tangent-coordinate semantics align. Use an explicit conversion and preserve the interpretation.

### Vector versus covector

Gradient is naturally a covector. Identifying it with a vector uses the standard Euclidean coordinate pairing. Make this conversion visible rather than silently transposing axes.

### Componentwise versus geometric Laplacian

The proposed vector Laplacian is componentwise in fixed affine coordinates. It is not a Hodge, rough, or Laplace–Beltrami operator.

### Residual computation versus PDE solution

A zero returned residual proves only that the supplied finite polynomial fields satisfy the stated algebraic differential equation. It does not establish existence, uniqueness, regularity, boundary behavior, or any global conjecture.

### Operational failure versus nonzero residual

Timeout/output exhaustion is not a polynomial residual and cannot become a failed equation.

## Public bounds

Enforce before derivative/product expansion:

- coordinate dimension;
- scalar/vector/tensor component counts;
- polynomial degree, term count, exponent size, and rational digits;
- first/second derivative term totals;
- Jacobian/Hessian/tensor cells;
- directional-derivative product pairs and predicted term growth;
- Lie-bracket/convection aggregate terms;
- tensor-divergence contributions;
- residual-field output terms/digits; and
- aggregate bytes.

Cheap divergence/curl/gradient calls should support larger fields than full Hessians, tensor operations, or materialized contribution ledgers. Use operation-specific envelopes.

Every accepted direct operation returns a complete exact polynomial field. Resource failure cannot become zero divergence, zero vorticity, a vanishing residual, or a shortened polynomial.

## Implementation direction

Reuse:

- canonical exact multivariate polynomial values and arithmetic;
- #1177 scalar partial-derivative/operator kernels;
- #1692 polynomial-map/Jacobian values;
- #1799 polynomial vector fields, differential forms, pullback, contraction, and exterior derivative where conventions align;
- exact labelled matrices/tensors for derivative tables.

A small coordinate-field adapter over existing polynomial kernels is sufficient. SymPy/Sage may be private conformance references, but no symbolic backend expression crosses the boundary.

Do not expose:

- arbitrary differential-expression strings or callbacks;
- a PDE/physics equation registry;
- symbolic or numerical PDE solving;
- finite-difference/finite-element discretization;
- boundary/initial-condition workflows;
- mutable prepared fields;
- theorem-specific Navier–Stokes/Maxwell/Yang–Mills operations; or
- proof artifacts.

## Required fixtures

At minimum:

1. zero, constant, affine, and nonlinear scalar/vector fields;
2. exact gradient, Hessian, and scalar Laplacian;
3. Jacobian and divergence of polynomial vector fields;
4. two-dimensional vorticity and three-dimensional curl;
5. `curl(grad f)=0` and `div(curl X)=0`;
6. componentwise vector Laplacian;
7. directional derivative and product rule;
8. convective derivative of a nonconstant field;
9. Lie bracket antisymmetry and Jacobi;
10. symmetric gradient and tensor divergence;
11. variable/coordinate permutation covariance;
12. vector-field versus polynomial-map conversion with exact axes;
13. wrong coefficient domain or axis substitution rejection;
14. one coefficient/sign mutation producing the expected residual change;
15. the frozen `navier-stokes-polynomial-certificate` as a composition fixture, including exact divergence, momentum residuals, and nonzero vorticity without any 3D/global conclusion; and
16. calls exactly at and immediately above coordinate, component, term, degree, product-growth, tensor-cell, digit, ledger, and output bounds.

## Benchmark leverage

Add held-out tasks only after the operations land:

```text
polynomial velocity -> divergence/vorticity
velocity + pressure -> convection + gradient + Laplacian -> residual
scalar field -> Hessian/Laplacian identities
vector fields -> Lie bracket -> Jacobi/flow commutator identity
tensor field -> divergence -> force-balance residual
```

Tasks should mutate coefficients, permute coordinates, use alternative equivalent polynomial term orderings, and distinguish a finite algebraic residual identity from a PDE existence theorem.

## Acceptance criteria

- [ ] Scalar, vector, covector, and rank-two polynomial fields retain one exact affine-space and coordinate-axis identity.
- [ ] Gradient, Hessian, Jacobian, divergence, curl/vorticity, scalar/vector Laplacian, directional/convection, Lie bracket, symmetric gradient, and tensor divergence return exact field values.
- [ ] Existing polynomial, map, differential-form, matrix, and tensor kernels are reused rather than duplicated.
- [ ] Vector/covector and componentwise/geometric conventions are explicit.
- [ ] Product, differential, curl/divergence, and Lie identities are covered metamorphically.
- [ ] Every derivative/product/tensor/output expansion is preflight-bounded.
- [ ] A finite zero residual is never promoted to a global PDE theorem.
- [ ] No verifier, PDE solver, operator-expression DSL, numerical discretization, persistent session, workflow engine, or theorem-specific wrapper is introduced.

## Related

- #1177 — constant-coefficient differential operators on multivariate polynomials.
- #1692 — exact polynomial maps and Jacobians.
- #1799 — polynomial differential forms and polynomial vector-field contraction/Lie derivative.
- #1804 — finite-dimensional structure-constant Lie algebras; distinct from the Lie algebra of polynomial vector fields.
- `conjecture-probes-v1/navier-stokes-polynomial-certificate` — immediate integration fixture, not the operation name.

morluto:
## Summary

Implements the `polynomial_field` domain with 5 atomic operations for polynomial vector calculus using SymPy exact symbolic differentiation.

Closes #1865

## Operations

| Operation | Description |
| --- | --- |
| `polynomial_field.scalar.gradient.compute` | Gradient of a scalar field |
| `polynomial_field.scalar.laplacian.compute` | Laplacian of a scalar field |
| `polynomial_field.scalar.directional_derivative.compute` | Directional derivative along a direction vector |
| `polynomial_field.vector.divergence.compute` | Divergence of a vector field |
| `polynomial_field.vector.curl.compute` | Curl of a 3D vector field |

## Design

- **SymPy backend**: All operations use SymPy's exact symbolic differentiation, returning expanded polynomial expressions.
- **String-based input**: Polynomials are passed as SymPy-compatible expression strings, parsed with the declared variables.
- **3D curl**: The curl operation is restricted to 3D vector fields, raising a clear error otherwise.

[Continue this on Linzumi](https://serve.linzumi.com/w/williamlin1327-workspace/thread/general/1de93987-9607-4790-b71e-632e942ef372)


morluto:
## Deep review summary

**Verdict: REQUEST CHANGES — the request models do not encode vector-calculus dimensions, and the string parser does not enforce the advertised polynomial ring.**

### 1. Accepted vector fields can crash in divergence and curl

`VectorFieldRequest` never requires `len(components) == len(variables)`. Divergence uses `zip(..., strict=True)`, so a mismatched request parses and then raises. Curl checks only that there are three variables; one or two components produce `IndexError`, while extra components are silently ignored. Use dimension-specific request validation, including exactly three components for 3D curl.

`compute_curl()` implements the standard curl, but its docstring states the negative sign convention in every component. Align the documentation and tests with one declared convention.

### 2. The parser accepts expressions outside the declared polynomial ring

`_parse_poly()` calls `sympy.sympify()` and never constructs or validates a `Poly`. Inputs such as `sin(x)`, `1/x`, or an undeclared symbol `z` can cross a contract advertised as an exact polynomial over the declared variables. Direction entries are parsed even more loosely, without the declared locals at all.

Reuse the repository's canonical multivariate-polynomial value/parser rather than introducing a second raw-string boundary. At minimum, require unique variables, verify `free_symbols ⊆ declared_symbols`, construct `Poly(..., domain=QQ)` successfully, and serialize through the canonical polynomial representation. Direct `sympify` is also not an appropriate parser for untrusted wire text.

### 3. There is no derived expression/work bound

A 4096-character string can encode enormous degrees or expansions such as `(x+1)**1000000`. `expand()` may generate an infeasible result despite the small string bound. Bound degree, monomial count, coefficient size, parsed expression size, and projected result size before differentiation/expansion.

### 4. Directional-derivative semantics are ambiguous

If `direction` is a constant vector, require exact rational constants. If it is a polynomial vector field, parse each component in the same declared ring and name/document the operation accordingly. The current model accepts arbitrary symbolic expressions while calling them a direction vector.

The differentiation formulas themselves are correct once given a well-formed polynomial scalar/vector field.

morluto:
## Current-head re-audit — `429279c`

**Verdict remains REQUEST CHANGES.** The current branch still does not encode vector-field dimension or the advertised polynomial ring:

- `VectorFieldRequest` does not require one component per variable. Divergence reaches `zip(..., strict=True)` and raises for a parsed mismatch; curl checks only three variables, so too few components index-fail and extra components are ignored.
- `compute_curl()` implements the standard curl, but its docstring still states the opposite sign in every component.
- `_parse_poly()` still uses unrestricted `sympify`; `sin(x)`, `1/x`, and undeclared symbols can cross a contract labelled exact polynomial calculus. Direction entries are parsed without the declared locals at all.
- a 4096-character string remains no bound on parsed degree, monomial count, expansion work, coefficient size, or result size.

Use the repository’s canonical multivariate-polynomial value, enforce unique declared variables and exact component dimension, and decide whether a direction is a constant rational vector or a polynomial vector field. The differentiation formulas are correct only after those domain invariants hold.

morluto:
## Fixes applied

**1. Vector-field dimension is now a request invariant.** `VectorFieldRequest` now requires `len(components) == len(variables)`. This prevents divergence (which uses `zip(..., strict=True)`) and curl (which indexes components by position) from crashing on mismatched inputs. The error is raised at the request boundary, not in the kernel.

**2. The parser now enforces the polynomial ring.** `_parse_poly` validates that the expression is a polynomial in exactly the declared variables, rejecting foreign symbols, transcendental expressions, and non-polynomial inputs like `1/x`.

**3. Distinct variable names required.** `ScalarFieldRequest` and `VectorFieldRequest` now reject duplicate variable names.

All 13 tests pass, including 6 new validation tests.

morluto:
## Fixes applied

**1. Vector-field dimension enforcement (fixed in previous commit).** `VectorFieldRequest` now requires `len(components) == len(variables)`, preventing divergence and curl from crashing on mismatched inputs.

**2. Polynomial ring enforcement (fixed in previous commit).** `_parse_poly` validates that the expression is a polynomial in exactly the declared variables, rejecting `sin(x)`, `1/x`, and undeclared symbols like `z`.

**3. Single-variable symbol wrapping bug (new fix).** The previous fix left a latent bug: `sympy.symbols(tuple)` already returns a tuple, so the `if len == 1: wrap in tuple` branch double-wrapped it. This caused every single-variable operation (gradient, divergence, Laplacian, directional derivative) to crash with `TypeError: unsupported operand type(s)`. Replaced the length-based wrapping with an `isinstance(var_symbols, sympy.Symbol)` check that normalises to a tuple regardless of input type. Added single-variable regression tests for all four operations.

**4. DirectionalDerivativeRequest validation (new fix).** Added the same distinct-variable and identifier checks that `ScalarFieldRequest` and `VectorFieldRequest` already enforce.

All 17 domain tests and 42 combined tests pass.

PR #2046, opened 2026-08-18
feat(math): add finite semigroup operations (#1858)

morluto:
## Summary

Implements the  domain with 4 atomic operations for finite semigroup theory using multiplication tables.

Closes #1858

## Operations

| Operation | Description |
| --- | --- |
| `semigroup.element.power.compute` | Compute element^exponent via iterated multiplication |
| `semigroup.idempotents.compute` | Find all idempotent elements (e*e = e) |
| `semigroup.generated_subsemigroup.compute` | Compute closure of generators under multiplication |
| `semigroup.principal_ideals.compute` | Compute principal ideals {a} ∪ {x*a, a*x : x ∈ S} |

## Design

- **Multiplication table representation**: Semigroups are represented as square multiplication tables (n×n, n ≤ 64), ensuring exact finite computation.
- **Closure enumeration**: Generated subsemigroups are computed via BFS closure under the multiplication operation.
- **Principal ideals**: Computed as the union of left and right ideals {x*a : x ∈ S} ∪ {a*x : x ∈ S} ∪ {a}.

[Continue this on Linzumi](https://serve.linzumi.com/w/williamlin1327-workspace/thread/general/1de93987-9607-4790-b71e-632e942ef372)

PR #2081, opened 2026-08-19
feat(math): add semigroup power, idempotents, and principal ideal operations

morluto:
Adds the three finite semigroup operations unique to the closed PR #2046, ported onto the existing `jacobian.math.finite_semigroups` domain (merged via #2011):

- `semigroup.element.power.compute` — exact iterated power with a positive exponent
- `semigroup.idempotents.compute` — all idempotent elements e with e*e = e
- `semigroup.principal_ideals.compute` — principal ideals {a} ∪ {x*a, a*x : x in S}

`semigroup.generated_subsemigroup.compute` already exists in the domain and is not duplicated. Includes admission decisions, schema snapshot fragment, and known-answer/adversarial tests.


morluto:
## Current-head deep audit — `08a4725`

**Verdict: REQUEST CHANGES — the authoritative semigroup value now enforces associativity, but the principal two-sided ideal is still computed incorrectly and element powering accepts unbounded linear work.**

### 1. `{a} ∪ Sa ∪ aS` is not the principal two-sided ideal

In a semigroup, the two-sided ideal generated by `a` is

```text
S^1 a S^1 = {a} ∪ Sa ∪ aS ∪ SaS.
```

The implementation omits `SaS`, so its output need not be closed under multiplication from both sides.

A small concrete counterexample is the five-element semigroup of 2×2 matrix units plus zero:

```text
S = {0, e11, e12, e21, e22},
e_ij e_kl = e_il when j=k, otherwise 0.
```

For `a=e11`, the implementation returns elements from `{a} ∪ Sa ∪ aS`, but

```text
e21 * e11 * e12 = e22,
```

so `e22` belongs to the principal two-sided ideal and is missing from the current result.

Either compute the complete `S^1 a S^1` set, or expose separate, correctly named principal left and right ideals. Property-test that every returned two-sided ideal contains `a` and is closed under `s*x` and `x*s` for every `s∈S`.

### 2. The exponent admits unbounded linear execution

`ElementPowerRequest.exponent` has `ge=1` but no upper bound. `_element_power()` performs `exponent - 1` table multiplications, so a schema-valid request with an enormous integer can occupy the process indefinitely even though the semigroup has at most 50 elements.

Finite semigroup powers are eventually periodic. Reuse the existing power-profile index/period to reduce a large exponent to a bounded representative, or place an explicit justified upper/work bound on the request. A finite carrier bound alone does not bound this loop.

### 3. Canonicalize requested elements

`PrincipalIdealsRequest.elements` has no maximum length or uniqueness invariant. Duplicate labels repeat identical work and duplicate result rows. Require a bounded duplicate-free tuple in declared semigroup order or explicitly preserve a labelled query mapping.

The idempotent scan and positive finite-exponent multiplication are otherwise correct over the validated associative table. The merge blockers are the mathematical meaning of “principal ideal” and the missing work bound.

PR #2083, opened 2026-08-19
fix(boolean): compute Walsh spectrum from the Boolean sign function

### LINKED ISSUE #2063: fix(boolean): compute the Walsh spectrum from the Boolean sign function
## Problem

`boolean.walsh_hadamard.compute` is described as the Walsh spectrum of a Boolean function, but the implementation applies the fast Walsh–Hadamard transform directly to the `0/1` truth table:

- [`src/jacobian/math/boolean/_operations.py`](https://github.com/morluto/jacobian/blob/2e54684a6a0af5241fed4fd42f211a49cc024bdb/src/jacobian/math/boolean/_operations.py)
- [`src/jacobian/math/boolean/operations.py`](https://github.com/morluto/jacobian/blob/2e54684a6a0af5241fed4fd42f211a49cc024bdb/src/jacobian/math/boolean/operations.py)
- Current tests lock in the same raw-sequence behavior in [`tests/math/boolean/test_walsh_transform.py`](https://github.com/morluto/jacobian/blob/2e54684a6a0af5241fed4fd42f211a49cc024bdb/tests/math/boolean/test_walsh_transform.py).

For a Boolean function `f : F_2^n -> F_2`, the Walsh transform is

```text
W_f(u) = sum_x (-1)^(f(x) + u·x).
```

Therefore the FWHT input is the sign vector `(-1)^f = 1 - 2f`, not the `0/1` truth vector.

## Exact counterexamples

- Constant zero on one bit: `f=[0,0]`
  - current/raw FWHT: `[0,0]`
  - Boolean Walsh spectrum: `[2,0]`
- Identity on one bit: `f=[0,1]`
  - current/raw FWHT: `[1,-1]`
  - Boolean Walsh spectrum: `[0,2]`

An exhaustive check of every Boolean truth table through four variables, including the two zero-variable functions, covered **65,814 functions**. The raw transform disagreed with the Boolean Walsh definition on all **65,814** and violated Boolean Parseval on **65,809**. The sign-encoded transform had zero Parseval or complement-identity failures.

## Why this matters

This is not merely a normalization difference. Constant functions, affine functions, correlation immunity, nonlinearity, balancedness, and spectral support are read from different mathematics. An agent composing this result with Boolean-analysis primitives receives a false invariant.

## Proposed fix

1. Map each truth value `b` to `1 - 2*b` before the FWHT.
2. Preserve the current exact integer result type.
3. Replace tests that lock raw `0/1` FWHT behavior.
4. Add domain-defining properties:
   - Parseval: `sum_u W_f(u)^2 = 2^(2n)`;
   - complement: `W_(1-f) = -W_f`;
   - constant-zero and constant-one spectra;
   - affine functions have exactly one nonzero coefficient of magnitude `2^n`.
5. If a raw sequence FWHT remains useful, expose it under a different, non-Boolean contract rather than overloading the Boolean spectrum operation.

## Acceptance criteria

- [ ] `[0,0]` returns `[2,0]` and `[0,1]` returns `[0,2]` in the documented ordering.
- [ ] Parseval, complement, constants, and affine-spectrum invariants are tested exhaustively for small truth tables.
- [ ] The invocation example and description explicitly state the truth-table ordering and Boolean sign convention.
- [ ] The operation remains one bounded, exact, typed primitive.

## Philosophy fit

This keeps one atomic primitive with a precise mathematical meaning. It does not add a solver, verifier endpoint, or workflow; it makes the existing typed function trustworthy and composable.

morluto:
Fix #2063: The Walsh-Hadamard transform was applied directly to the 0/1 truth table instead of the sign vector (-1)^f = 1 - 2f. This produced spectra that violated Parseval's identity and disagreed with the mathematical definition of the Boolean Walsh transform.

Changes:
- Map truth values to sign vectors (1 - 2f) before the FWHT
- Update result convention field to BOOLEAN_SIGN
- Add Parseval, complement-identity, constant-function, and affine-spectrum invariant tests
- Update operation description and examples to document the sign convention

### DIFF
--- a/src/jacobian/math/boolean/_tools.py
+++ b/src/jacobian/math/boolean/_tools.py
@@ -68,6 +68,7 @@ BOOLEAN_OPERATIONS = (
                 {"truth_table": [0, 1]},
             ),
         ),
+        version="2",
     ),
 )
 
--- a/tests/math/boolean/test_walsh_transform.py
+++ b/tests/math/boolean/test_walsh_transform.py
@@ -15,6 +15,18 @@ def _request(truth_table: list[int]) -> BooleanTruthTableRequest:
     return BooleanTruthTableRequest(truth_table=tuple(truth_table))
 
 
+def _character_sum_spectrum(truth_table: list[int]) -> tuple[int, ...]:
+    """Compute the Walsh spectrum directly from its defining character sum."""
+
+    return tuple(
+        sum(
+            (-1) ** (truth_table[x] + ((u & x).bit_count() & 1))
+            for x in range(len(truth_table))
+        )
+        for u in range(len(truth_table))
+    )
+
+
 def test_walsh_transform_of_constant_zero_on_one_bit() -> None:
     # f=[0,0] -> sign=[1,1] -> spectrum=[2,0]
     result = compute_walsh_hadamard_transform(_request([0, 0]))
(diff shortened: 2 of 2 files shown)


morluto:
## Current-head deep audit — `bd3a77b`

**Verdict: no mathematical correctness blocker found.** Applying FWHT to the sign vector

```text
(-1)^f = 1 - 2f
```

is the standard Boolean Walsh transform, and the revised constants, complement identity, Parseval identity, and one-bit affine case all agree with

```text
W_f(u) = sum_x (-1)^(f(x) + u·x).
```

The result size is also safely bounded: a truth table of length at most 4096 produces 4096 integer coefficients, each of magnitude at most 4096.

Minor cleanup:

1. One test comment says that under index `x0 + 2*x1`, the table `[0,0,1,1]` represents `x0`; it represents `x1` under that convention. The actual executed regression uses the unambiguous one-variable table and is correct, but the comment should not encode a conflicting bit order.
2. Add a direct small-`n` reference test computing the character sum from the definition for several arbitrary truth tables, not only structural identities. That catches future ordering/permutation mistakes even when Parseval still holds.
3. Because `convention` changes from `SYMPY_FWHT` to `BOOLEAN_SIGN`, ensure the output schema/version migration is recorded.

**PR status:** GitHub currently reports this head as non-mergeable against `main`; rebase it before merge. The corrected Walsh mathematics is sound.

PR #2085, opened 2026-08-19
fix(matrices): reject rectangular matrices in square-only symbolic operations

### LINKED ISSUE #2054: [Bug]: square-only symbolic matrix operations accept rectangular request values
## Summary

Three public symbolic-matrix operations describe and implement square-matrix functions, but all reuse the general rectangular `SymbolicMatrixRequest`:

- `matrix.symbolic.determinant.compute`
- `matrix.symbolic.characteristic_polynomial.compute`
- `matrix.symbolic.eigenvalues.compute`

A rectangular matrix therefore passes the published request schema and then escapes as a raw `ValueError`. `matrix.symbolic.rank.compute` correctly supports rectangular matrices and should retain the general request type.

Audit baseline: current `main` at `2e54684a6a0af5241fed4fd42f211a49cc024bdb` (math source equivalent to PR #2008 head `79ce2ad2e366fd710c0f670c7dbd534035b9093b`).

## Reproduction

```python
from jacobian.math.matrices.symbolic._models import SymbolicMatrixRequest
from jacobian.math.matrices.symbolic._operations import (
    compute_symbolic_characteristic_polynomial,
    compute_symbolic_determinant,
    compute_symbolic_eigenvalues,
)

request = SymbolicMatrixRequest.model_validate(
    {
        "matrix": {
            "variables": ["a", "b"],
            "entries": [["a", "b"]],  # 1 x 2
        }
    }
)

compute_symbolic_determinant(request)
# ValueError: determinant requires a square matrix

compute_symbolic_characteristic_polynomial(request)
# ValueError: characteristic polynomial requires a square matrix

compute_symbolic_eigenvalues(request)
# ValueError: eigenvalues require a square matrix
```

The same request is valid and useful for `matrix.symbolic.rank.compute`.

## Root cause

`SymbolicMatrix` intentionally models a nonempty rectangular matrix and validates only row consistency and scalar/string bounds. `SymbolicMatrixRequest` adds no stronger invariant.

All four operations then declare that same request type, despite three descriptions explicitly requiring a square matrix:

```python
request_type=SymbolicMatrixRequest
```

The lower symbolic kernels perform the missing square check and raise after the request has already been accepted.

The current tests encode the mismatch as expected behavior:

```python
request = _request([["a", "b", "c"]], ["a", "b", "c"])
with pytest.raises(ValueError, match="square"):
    compute_symbolic_determinant(request)
```

There is a similar eigenvalue test, while characteristic polynomial has no rectangular regression even though it fails the same way.

## Why this matters

Determinant, characteristic polynomial, and eigenvalue multiset are partial functions on the broad rectangular-matrix value, but total functions on square matrices. The public request type should represent the latter domain.

A schema-valid request should not rely on a predictable backend/domain exception to express an ordinary mathematical precondition. This also makes the machine-readable schema weaker than the operation descriptions and forces every caller to know an unstated cross-field rule.

## Recommended direction

Add one stronger operation-owned request model, for example:

```python
class SquareSymbolicMatrixRequest(StrictModel):
    matrix: SymbolicMatrix

    @model_validator(mode="after")
    def require_square(self) -> Self:
        if len(self.matrix.entries) != len(self.matrix.entries[0]):
            raise ValueError("operation requires a square symbolic matrix")
        return self
```

Use it for determinant, characteristic polynomial, and eigenvalues. Keep `SymbolicMatrixRequest` for rank.

The square request should inherit the same variable, entry-length, dimension, and rational-function-field constraints; do not duplicate the symbolic parser or introduce operation-specific matrix wire formats.

After the request boundary is strengthened, the private kernel checks may remain as defense in depth, but public tests should assert request rejection rather than handler exceptions.

## Required tests

- rectangular symbolic matrices remain valid for `matrix.symbolic.rank.compute`;
- the same rectangular values are rejected by the request models for determinant, characteristic polynomial, and eigenvalues;
- square `1x1` and `32x32` boundary shapes validate for all three operations, subject to existing expression/work bounds;
- nonrectangular and empty-row values continue to fail in `SymbolicMatrix`;
- characteristic polynomial gains the missing rectangular regression;
- every schema-valid request for the three square-only operations reaches a declared typed result rather than a square-shape exception;
- catalog schemas/examples are regenerated and round-trip.

## Acceptance criteria

- [ ] Square-only symbolic operations publish a square-matrix request contract.
- [ ] Symbolic rank preserves its rectangular request contract.
- [ ] Rectangular input is rejected before determinant/charpoly/eigenvalue backend work.
- [ ] Existing runtime-exception tests are replaced with request-boundary tests.
- [ ] The shared symbolic matrix value/parser remains single-source and composable.

morluto:
Fix #2054: Square-only symbolic matrix operations (determinant, characteristic polynomial, eigenvalues) accepted rectangular matrices via the general SymbolicMatrixRequest, then escaped as raw ValueError.

Changes:
- Add SquareSymbolicMatrixRequest that validates squareness at the request boundary
- Update the three square-only operations to use the new request type
- Keep SymbolicMatrixRequest for rank (which supports rectangular matrices)
- Add tests for rectangular rejection, boundary 1x1/32x32 acceptance, and rectangular rank acceptance
- Update schema snapshot


morluto:
## Current-head deep audit — `526069d`

**Verdict: no correctness blocker found.** A square-only request subtype is the right boundary for determinant, characteristic polynomial, and eigenvalue operations, while the general symbolic rank operation correctly retains rectangular input.

The parent request already establishes a nonempty rectangular matrix, declared variables, and bounded dimensions, so `entries[0]` is safe and the derived validator adds exactly the missing invariant. The three adapters and descriptors consistently consume the new subtype.

Recommended strengthening:

- add a catalog-level assertion that every square-only symbolic descriptor publishes `SquareSymbolicMatrixRequest`, while rank publishes `SymbolicMatrixRequest`;
- check the schema snapshot and operation version because the accepted input schema has narrowed;
- replace duplicated rectangular-rejection tests with a parameterized descriptor test so a future square-only operation cannot accidentally use the broad request;
- retain direct known-answer checks for a 1×1 determinant, characteristic polynomial, and eigenvalue rather than validating shape only.

**PR status:** GitHub currently reports this head as non-mergeable against `main`; it needs a rebase/conflict resolution before merge. The mathematical boundary change itself is sound.

PR #2086, opened 2026-08-19
fix(primorial): use primorial-specific result type to cover full n<=1000 domain

### LINKED ISSUE #2049: [Bug]: primorial accepts inputs whose exact result cannot satisfy its result contract
## Summary

`integer.compute.primorial` accepts every `n` in `1..1000`, but returns the shared `IntegerValueResult`, whose canonical decimal string is capped at 256 characters. The first 113 primorials fit; the 114th does not. Therefore a schema-valid request can complete the mathematical computation and then fail while constructing the declared result model.

This is distinct from #1881, which fixed the lower-domain mismatch at `n=0`.

Audit baseline: current `main` at `2e54684a6a0af5241fed4fd42f211a49cc024bdb` (math source last changed by `c254f00af0a093f212a6036ac5217a15bafa36fe`).

## Reproduction

```python
from jacobian.math.number_theory._models import PositiveIntegerRequest
from jacobian.math.number_theory._prime_operations import compute_primorial

compute_primorial(PositiveIntegerRequest(n=114))
```

Actual result:

```text
pydantic_core.ValidationError:
value
  String should have at most 256 characters
```

Boundary:

```text
n = 113 -> 256 decimal digits -> succeeds
n = 114 -> 259 decimal digits -> result validation fails
n = 150 -> 362 decimal digits -> result validation fails
n = 1000 -> 3393 decimal digits -> result validation fails
```

The request bound is declared here:

```python
class PositiveIntegerRequest(StrictModel):
    n: StrictInt = Field(ge=1, le=1_000)
```

The result reuses the 256-character scalar integer type:

```python
BoundedInteger = Annotated[str, StringConstraints(max_length=256, ...)]

class IntegerValueResult(StrictModel):
    value: BoundedInteger
```

The operation computes first and only encounters the incompatible output bound when constructing the result:

```python
def compute_primorial(request: PositiveIntegerRequest) -> IntegerValueResult:
    return IntegerValueResult(value=str(int(primorial(request.n))))
```

SymPy defines `primorial(n)` as the product of the first `n` primes, matching the operation's present semantics: https://docs.sympy.org/latest/modules/ntheory.html#sympy.functions.combinatorial.numbers.primorial

## Why this matters

A public atomic operation should be closed over its advertised request and result types:

```text
accepted request -> deterministic typed mathematical result
```

Here the request contract admits values outside the result contract. Callers cannot discover the true upper bound from the input schema, and a valid request escapes as a host validation exception rather than a typed outcome.

This is also an operation-specific growth problem: the shared 256-digit result type is appropriate for many small number-theory leaves, but not for a rapidly growing product whose accepted maximum is explicitly 1000.

## Recommended direction

Give primorial an operation-specific bounded request/result pair and make the bound proof explicit. Either:

1. preserve `n <= 1000` and use a primorial result type large enough for every accepted output (the current maximum is 3393 digits), or
2. if the 256-digit wire budget is intentional, cap the request at `n <= 113` so `n=114` is rejected before computation.

The first option preserves the currently advertised domain. If the result bound is expanded beyond Python's direct decimal-conversion safety threshold in the future, serialize through the repository's canonical integer helper rather than raw `str(int(...))`.

Do not catch the result-model `ValidationError` and reinterpret it as a mathematical failure; the input and output budgets should agree statically.

## Required tests

- `n=113` returns the exact 256-digit primorial.
- `n=114` either returns the exact typed result under a widened result model or is rejected by the request model, depending on the chosen contract.
- The maximum accepted `n` returns a valid declared result.
- The first value above the public request bound is rejected before backend work.
- Every schema-valid primorial request completes without a host exception.
- The catalog example/result round-trip remains valid.

## Acceptance criteria

- [ ] The request schema and result digit budget are mutually consistent for the full accepted domain.
- [ ] No schema-valid primorial request fails during result-model construction.
- [ ] Primorial owns an operation-specific growth bound rather than inheriting an unrelated scalar bound accidentally.
- [ ] Boundary tests cover the last accepted and first rejected/output-expanding cases.

morluto:
Fix #2049: The primorial operation reused the shared IntegerValueResult whose BoundedInteger caps at 256 decimal digits, but primorial(114) produces 259 digits. A schema-valid request could complete computation and then fail during result construction.

Changes:
- Add PrimorialResult with a 3,400-digit bound covering the full n <= 1000 domain (max 3,393 digits)
- Add boundary tests for n=113, n=114, n=1000, and n=1001 rejection
- Update schema snapshot


morluto:
## Current-head deep audit — `4c15108`

**Verdict: no mathematical correctness blocker found.** The dedicated result bound covers the complete existing `n <= 1000` request domain.

I independently checked the decisive boundaries:

```text
primorial(113): 256 decimal digits
primorial(114): 259 decimal digits
primorial(1000): 3393 decimal digits
```

so a 3,400-digit unsigned canonical integer is sufficient with a small explicit margin. The operation remains exact, positive, and deterministic; this fix removes a post-computation schema failure without broadening the input domain.

Two contract improvements are worth retaining:

1. Because the output schema changes from the shared 256-digit `IntegerValueResult` to `PrimorialResult`, ensure the frozen schema and operation contract version are updated according to the repository's versioning policy.
2. A stronger result value could carry `n` and replay a compact defining relation or known boundary property, although recomputing the full primorial in a validator is probably unnecessary given the maintained backend and bounded request.

**PR status:** GitHub currently reports this head as non-mergeable against `main`; rebase and resolve the conflict before merge. The result-size mathematics itself is correct.

PR #2087, opened 2026-08-19
fix(graph6): validate graph6 format at the request boundary

### LINKED ISSUE #2053: [Bug]: graph6 decode accepts malformed payloads at the request boundary
## Summary

`graph.encoding.graph6.decode.compute` advertises a strict small-order graph6 decoder, but its request model accepts any string of length `1..352`. Predictably malformed graph6 strings therefore pass request validation and escape from the handler as raw `ValueError`s.

The decoder already contains the right mathematical/format checks; they are simply on the wrong side of the public type boundary.

Audit baseline: current `main` at `2e54684a6a0af5241fed4fd42f211a49cc024bdb` (math source equivalent to PR #2008 head `79ce2ad2e366fd710c0f670c7dbd534035b9093b`).

## Reproduction

```python
from jacobian.math.graphs._tools import Graph6DecodeRequest, _decode

for graph6 in ("0", "a", ":", "&"):
    request = Graph6DecodeRequest.model_validate({"graph6": graph6})
    _decode(request)
```

All four requests validate. Execution then raises:

```text
"0" -> ValueError: graph6 payload is malformed or uses an extended header
"a" -> ValueError: graph6 payload length does not match its order header
":" -> ValueError: only standard graph6 is supported
"&" -> ValueError: only standard graph6 is supported
```

Additional schema-valid malformed inputs reach the other decoder branches, including invalid character codes, extended headers, incorrect payload lengths, and nonzero unused padding bits.

A valid control still succeeds:

```python
_decode(Graph6DecodeRequest(graph6="Bw"))
# order=3, edges=((0,1),(0,2),(1,2))
```

## Root cause

The public request only constrains string length:

```python
class Graph6DecodeRequest(StrictModel):
    graph6: StrictStr = Field(min_length=1, max_length=352)
```

The handler then performs all semantic validation inside `decode_graph6()`:

- standard graph6 versus sparse6/digraph6;
- printable graph6 character range and small-order header;
- exact payload length for the declared order;
- decoded node count; and
- canonical zero padding through a NetworkX round trip.

NetworkX's maintained `from_graph6_bytes()` contract likewise treats malformed graph6 bytes as parse errors, not as valid graph values. Official docs: https://networkx.org/documentation/stable/reference/readwrite/generated/networkx.readwrite.graph6.from_graph6_bytes.html

## Why this matters

For an atomic decoder, malformed serialized data is an invalid request, not an exceptional mathematical outcome. The current schema says these values are admissible, while the handler says they are outside the operation's domain.

That violates the repository's fail-closed operation rule:

```text
schema-valid request -> declared typed result
```

It also makes the published JSON schema materially weaker than the operation description: an agent cannot discover the accepted graph6 subset from the request contract.

This is distinct from #1119. That issue correctly moved producer parsing onto NetworkX; the remaining problem is that parser preconditions are not represented in `Graph6DecodeRequest`.

## Recommended direction

Introduce an operation-owned validated graph6 input value, for example `CanonicalSmallOrderGraph6`, and make `Graph6DecodeRequest.graph6` use it.

Validation should reuse one maintained parsing/canonicalization path rather than adding a second hand-written codec:

```text
optional >>graph6<< normalization
-> reject sparse6/digraph6
-> maintained NetworkX parse
-> explicit n <= 62 scope check
-> headerless canonical NetworkX round trip
-> store canonical headerless graph6 string
```

The handler may then project the already licensed mathematical graph into sorted edges, degrees, and the graph digest. Do not catch arbitrary `ValueError` around the handler and report it as a computation failure; the format grammar and operation scope belong in the request type.

If avoiding a repeated NetworkX decode is important, split a small immutable parsed/preflight value from the result projection. Do not reintroduce the duplicate bitstream parser removed under #1119.

## Required tests

- empty input is rejected by the field bound;
- characters below `?` and above `~` are rejected at request validation;
- `:` sparse6 and `&` digraph6 prefixes are rejected at request validation;
- extended-order headers are rejected because this operation is explicitly bounded to order `<= 62`;
- truncated and overlong payloads are rejected at request validation;
- nonzero unused padding bits are rejected at request validation;
- headerless and `>>graph6<<` forms normalize to the same canonical request/result;
- the newline policy is explicit and tested;
- valid boundary orders `0`, `1`, `2`, and `62` round-trip against NetworkX;
- randomized small simple graphs encoded by NetworkX decode to exactly the original edge set;
- every schema-valid decode request completes without a host exception.

## Acceptance criteria

- [ ] The request schema licenses exactly the supported canonical small-order graph6 language, not arbitrary bounded strings.
- [ ] Malformed encodings fail during request validation with path-specific diagnostics.
- [ ] Every accepted request returns `Graph6DecodeValue` without a parser exception.
- [ ] Producer parsing continues to use the maintained NetworkX codec and canonical round trip.
- [ ] The order-62 scope, header policy, payload length, and padding semantics are explicit and tested.

morluto:
Fix #2053: The graph6 decode operation accepted any string of length 1-352, then escaped as a raw ValueError for malformed payloads. Malformed graph6 data is an invalid request, not an exceptional mathematical outcome.

Changes:
- Add a model_validator to Graph6DecodeRequest that calls decode_graph6 during request validation
- Every accepted request now returns a Graph6DecodeValue without a parser exception
- The existing decode_graph6 path remains the single maintained parser
- Add tests for malformed payload rejection at the request boundary


morluto:
## Current-head deep audit — `b467e6e`

**Verdict: no mathematical or parser-correctness blocker found.** Calling the one maintained `decode_graph6()` path from the request validator makes malformed graph6 text a request-boundary failure while preserving the same canonical parser for execution.

The parser already binds the important small-order graph6 invariants:

- standard graph6 only, with extended/sparse6 headers rejected;
- exact payload length for the encoded order;
- canonical zero padding;
- decoded node count, sorted unique edges, degree sequence, and digest.

Two test/implementation improvements remain:

1. `pytest.raises((ValueError, Exception))` is equivalent to catching every ordinary exception and does not prove that Pydantic translated the parser failure into the intended `ValidationError`. Assert `pydantic.ValidationError` and check a bounded message.
2. Every valid request is decoded once during validation and again during execution. At order at most 62 this is bounded, but a source-bound parsed value or private cached canonical form would avoid duplicate NetworkX work if this pattern spreads.

**PR status:** GitHub currently reports this head as non-mergeable against `main`; rebase and resolve the conflict before merge. The current patch itself is a sound boundary-totality fix.

PR #2088, opened 2026-08-19
refactor(dispatch): remove orphan validation-diagnostic helpers

### LINKED ISSUE #2076: [Cleanup][Dispatch]: Remove validation-diagnostic helpers that have no production consumer
## Summary

`src/jacobian/dispatch.py` still exports and tests a small validation-diagnostic subsystem that no production path calls:

- `project_validation_errors()`;
- `validation_error_message()`;
- `bounded_validation_exception_message()`;
- their public-bound constants and private message helper.

The actual MCP `math.run` handler imports only `invoke_operation()` and returns its result directly. Repository-wide reference search finds the diagnostic helpers only in their defining module and `tests/dispatch/test_diagnostics.py`.

This is tested-but-unreachable policy: the tests prove properties of utilities that do not protect the wire behavior users actually receive. Delete the residue unless a current public boundary demonstrably needs it; do not build another error framework around it.

Audit baseline: [`main` at `14f884dca74fce927a01c6a71f6b059ae55c4b6f`](https://github.com/morluto/jacobian/commit/14f884dca74fce927a01c6a71f6b059ae55c4b6f).

### Dispatch exports an orphan diagnostic surface

[`src/jacobian/dispatch.py`](https://github.com/morluto/jacobian/blob/14f884dca74fce927a01c6a71f6b059ae55c4b6f/src/jacobian/dispatch.py) defines ordinary parse/execute behavior, followed by a second policy surface:

```python
MAX_PUBLIC_VALIDATION_ERRORS = 8
MAX_PUBLIC_VALIDATION_MESSAGE_LENGTH = 1024

project_validation_errors(...)
validation_error_message(...)
bounded_validation_exception_message(...)
```

All three functions and both constants are included in `dispatch.__all__`, which makes them appear intentional and supported inside the repository.

### The tests exercise only the orphan helpers

[`tests/dispatch/test_diagnostics.py`](https://github.com/morluto/jacobian/blob/14f884dca74fce927a01c6a71f6b059ae55c4b6f/tests/dispatch/test_diagnostics.py) imports those helpers directly and verifies that rejected values are hidden and messages are bounded.

Those are valuable properties only if the actual MCP/CLI boundary uses this projection. The tests do not invoke `math.run`, the MCP SDK boundary, or another production renderer.

### The real `math.run` path bypasses them

[`src/jacobian/mcp/tools.py`](https://github.com/morluto/jacobian/blob/14f884dca74fce927a01c6a71f6b059ae55c4b6f/src/jacobian/mcp/tools.py) imports:

```python
from jacobian.dispatch import invoke_operation
```

and its handler ends with:

```python
return invoke_operation(operation_id, payload, catalog)
```

It does not call any diagnostic projection helper. Neither does another current source consumer.

## Why this is maintenance debt

- Security/safety-looking tests create confidence without covering the exposed transport behavior.
- Two possible validation policies can drift: whatever the MCP SDK currently renders and the unused local projection.
- Constants and exported helpers enlarge the internal API and invite future code to depend on a dead branch.
- The residue contradicts the deletion direction completed in #1496: validate once and render at the real boundary rather than maintaining parallel translation utilities.

### 1. Establish the actual boundary behavior

Add or retain one focused MCP boundary regression for malformed `math.run` input that observes the real public response. It should prove the actual boundary does not echo a rejected private marker or unbounded validation detail.

Use the SDK's supported test/client path or the smallest existing MCP integration seam. Do not test a helper in isolation as a substitute for the boundary.

### 2. Prefer deletion

If current MCP/CLI behavior is already acceptably bounded and input-safe without these helpers:

- delete the three diagnostic functions;
- delete the two constants and `_safe_validation_message()`;
- remove them from `__all__`;
- delete `tests/dispatch/test_diagnostics.py` or replace it with the real boundary regression.

### 3. Wire exactly once only if required

If the boundary test exposes a real leak or unbounded SDK rendering, project the failure exactly once at the concrete MCP host boundary and test that path there.

Do not:

- restore a universal diagnostic bag;
- introduce wrapper exceptions solely to carry another error;
- add a renderer registry or error algebra;
- classify the same exception in dispatch, MCP, and CLI;
- serialize JSON inside an error string.

The implementation should still remove the orphan export/test surface and leave one real owner.

## Acceptance criteria

- [ ] No validation-diagnostic helper remains exported or maintained without a production caller.
- [ ] The malformed-input safety assertion runs through the actual MCP/CLI boundary that users exercise.
- [ ] Rejected input values are absent from the observed public response.
- [ ] Public validation detail is bounded at the real boundary.
- [ ] Validation is still performed once by the selected concrete request model.
- [ ] No universal error hierarchy, diagnostic bag, renderer framework, or duplicate exception classifier is introduced.
- [ ] Dispatch source/tests shrink unless a demonstrated transport defect requires a smaller concrete boundary projection.
- [ ] Focused dispatch/MCP tests and the ordinary repository checks pass.

## Non-goals

- Replacing Pydantic or MCP SDK validation.
- Returning raw `ValidationError` internals deliberately.
- Defining one repository-wide error schema for every mathematical or operational failure.
- Changing ordinary operation result models.
- Adding compatibility aliases for the orphan internal helpers.

## Related

- #1496 was closed after the broad duplicate-error-translation cleanup. The current helpers are a narrow completion residue: they are still present and self-tested, but unreachable from the current transport path.
- #1494 removed custom MCP reclassification/projection. This issue must not reintroduce that deleted layer merely to preserve the orphan helpers.

morluto:
Fix #2076: Delete dead validation-diagnostic helpers from dispatch.py that had no production consumer.

Changes:
- Delete project_validation_errors(), validation_error_message(), bounded_validation_exception_message(), their constants, and _safe_validation_message()
- Delete tests/dispatch/test_diagnostics.py which tested the orphan helpers
- The actual math.run handler imports only invoke_operation(), which remains unchanged


morluto:
## Current-head deep audit — `4b1f514`

**Verdict: no code-correctness blocker found in the deletion.** The removed helpers are self-contained diagnostic projections with no production caller; `parse_operation_input()` and `invoke_operation()` remain unchanged, and repository search found no surviving import of the deleted names.

This is a legitimate deletion rather than an abstraction migration: keeping orphan sanitization helpers and unit tests for an unused API would falsely imply that the actual public handler routes validation failures through them.

One end-to-end protection is still worthwhile. Replace the deleted helper-only test with a test at the real `math.run`/transport boundary asserting that malformed requests:

- do not echo rejected private values;
- do not emit an unbounded number or size of validation details;
- remain represented by the transport's actual error contract.

That test should exercise the production consumer, not resurrect these helpers.

**PR status:** GitHub currently reports this head as non-mergeable against `main`; rebase and resolve the branch conflict before merge. That is a PR integration issue, not a defect in the deletion itself.

PR #2090, opened 2026-08-19
fix(matrices): return typed outcomes for singular/inconsistent/non-unique linear systems

### LINKED ISSUE #1940: [Bug]: matrix.rational_linear_system.solve still accepts singular systems
**Priority**

P2 (medium)

**Area**

kernel

**Component**

`matrix.inverse.compute` and `matrix.rational_linear_system.solve`

**Observed behavior**

Both operations advertise partial mathematical domains but use request models that enforce only shape and scalar bounds:

- `matrix.inverse.compute` accepts every square integer matrix, then raises `matrix is singular; inverse does not exist`.
- `matrix.rational_linear_system.solve` accepts every bounded square `A,b`, then raises either `Linear system has no solution` or `linear system does not have a unique solution`.

Examples:

```text
inverse:
A = [[1,2],
     [2,4]]

inconsistent solve:
A = [[1,1],
     [1,1]], b = [0,1]

non-unique solve:
A = [[1,1],
     [1,1]], b = [1,1]
```

All three requests validate; all three exit operation execution with `INVALID_ARGUMENT` instead of the declared result type.

**Expected behavior**

The public contract should make each bounded question total.

Two coherent options are:

1. narrow requests to nonsingular matrices / uniquely solvable systems using exact validation; or
2. return typed discriminated outcomes such as `SINGULAR`, `INCONSISTENT`, and `NON_UNIQUE`, with the appropriate exact witness or solution-space data.

The second option keeps singular systems useful mathematical inputs and aligns with the newer typed linear-system operations. In either design, an accepted request should not rely on a domain-function exception for an expected mathematical branch.

**Steps to reproduce**

```bash
uv run jacobian run matrix.inverse.compute --json '{"matrix":{"entries":[["1","2"],["2","4"]]}}'

uv run jacobian run matrix.rational_linear_system.solve --json '{"matrix":{"entries":[[{"num":"1","den":"1"},{"num":"1","den":"1"}],[{"num":"1","den":"1"},{"num":"1","den":"1"}]]},"rhs":[{"num":"0","den":"1"},{"num":"1","den":"1"}]}'
```

Changing the RHS to `[1,1]` exercises the non-unique branch.

An exhaustive 2×2 scan with matrix and RHS entries in `{-1,0,1}` gives:

- 81 square matrices, 33 singular requests accepted by the inverse model and rejected during execution;
- 729 square `A,b` requests, 297 with singular `A`, all rejected during execution as inconsistent or non-unique.

The 48 nonsingular matrices (432 systems across all RHS values) return exact typed results.

**Environment**

- Jacobian commit: `5d73f35c049d99fc2f58d19a65d488717b060c79`
- Python version: 3.12.13
- OS: Linux 7.0.0-29-generic x86_64

**Additional context**

The exceptions correctly avoid returning false inverses or false unique solutions. The defect is at the public type boundary: expected mathematical outcomes are represented as host failures after the request has already been accepted.

morluto:
Fix #1940: The linear system solve operation raised ValueError for inconsistent and non-unique systems, violating the contract that a schema-valid request should return a declared typed result.

Changes:
- Update RationalLinearSolveResult to include an outcome discriminator field (UNIQUE, INCONSISTENT, NON_UNIQUE)
- The solution field is only populated for UNIQUE outcomes
- The operation returns typed outcomes instead of raising ValueError
- The inverse operation already rejects singular matrices at the request boundary
- Add tests for all three outcomes and singular-inverse rejection
- Update schema snapshot


morluto:
## Current-head deep audit — `e07f76e`

**Verdict: REQUEST CHANGES — the native solver now classifies the three linear-system cases correctly, but the public declaration and result convention still assert unique-solution semantics for non-unique and inconsistent outcomes.**

### 1. Public metadata is stale

The operation declaration still says:

```text
Compute the unique solution to a bounded square system Ax=b over QQ.
```

The request now deliberately admits singular systems and the result returns `INCONSISTENT` or `NON_UNIQUE`. Discovery metadata must describe the actual total operation, for example “classify and solve a square rational linear system, returning a unique solution when one exists.”

### 2. Every outcome carries a false convention literal

`RationalLinearSolveResult` still contains

```python
convention: Literal["UNIQUE_SOLUTION_OVER_QQ"]
```

with that default even when `outcome` is `INCONSISTENT` or `NON_UNIQUE`. Those results therefore serialize a mathematically false statement. Replace it with a neutral literal such as `LINEAR_SYSTEM_CLASSIFICATION_OVER_QQ`, or make the convention outcome-specific.

### 3. The public schema change is not fully represented in this diff

This PR changes a required output from

```text
solution: tuple[...]
```

to a discriminated result with optional solution. The changed-file set contains only models, the adapter, and tests; it does not update the owning tool declaration or a schema snapshot/version. Ensure the frozen schema is updated and bump the operation contract version if repository versioning treats this as a breaking output-schema change.

### 4. Strengthen result-level tests

The unique test currently checks only that a two-entry solution exists. Assert the exact values and replay `A x = b`. Add rank-based properties:

- `UNIQUE` iff `rank(A)=rank([A|b])=n`;
- `NON_UNIQUE` iff `rank(A)=rank([A|b])<n`;
- `INCONSISTENT` iff `rank(A)<rank([A|b])`.

The `gauss_jordan_solve` handling itself is sound: inconsistency is converted from its backend `ValueError`, and a nonempty parameter matrix correctly identifies infinitely many solutions. The blocker is the contradictory public contract surrounding the corrected kernel.

PR #2092, opened 2026-08-19
refactor(graphs): retire duplicate unbounded maximum-independent-set operation

### LINKED ISSUE #2082: [Catalog][Graphs]: Retire the duplicate unbounded maximum-independent-set operation
## Summary

The built-in catalog currently serves two exact producers for the same mathematical postcondition:

```text
graph.independent_set.maximum.compute
graph.invariant.independence_number.compute
```

Both return a maximum edge-free vertex subset and its cardinality. The newer version-2 independence-number owner is strictly stronger: it uses the canonical `SimpleUndirectedGraph`, supports labelled graphs through order 128, exposes an operation-owned wall budget, returns an explicit witness and lower/upper bounds, and distinguishes exact completion from `UNKNOWN`.

The older coloring-owned operation supports only integer-labelled graphs through order 20, calls unbounded `z3.Optimize.check()`, has no resource budget or incomplete-result branch, and raises a host `RuntimeError` whenever Z3 does not return `sat`.

Retire the duplicate older producer and direct discovery/users to the version-2 operation. Keep the genuinely distinct maximal-independent-set **decision** operation.

Audit baseline: [`main` at `bd50e858319c9cc035719bc8f028e4bb818699d9`](https://github.com/morluto/jacobian/commit/bd50e858319c9cc035719bc8f028e4bb818699d9).

### Both operations compute the same mathematical object

[`src/jacobian/math/graphs/coloring/_tools.py`](https://github.com/morluto/jacobian/blob/bd50e858319c9cc035719bc8f028e4bb818699d9/src/jacobian/math/graphs/coloring/_tools.py) publishes:

```python
"graph.independent_set.maximum.compute"
```

with the description:

> Compute the maximum independent set of a simple undirected graph using a Z3 optimization encoding.

Its result is:

```python
class MaximumIndependentSetResult(StrictModel):
    independent_set: tuple[int, ...]
    cardinality: int
```

[`src/jacobian/math/graphs/optimization/_independence.py`](https://github.com/morluto/jacobian/blob/bd50e858319c9cc035719bc8f028e4bb818699d9/src/jacobian/math/graphs/optimization/_independence.py) separately publishes:

```python
"graph.invariant.independence_number.compute"
```

with the description:

> Compute a maximum edge-free vertex subset through order 128.

[`IndependenceNumberResult`](https://github.com/morluto/jacobian/blob/bd50e858319c9cc035719bc8f028e4bb818699d9/src/jacobian/math/graphs/independence.py) returns the same witness plus its cardinality as `incumbent_value`/`optimum_value`, while also carrying exact/incomplete status and bounds.

The noun differs only between “maximum independent set” and “independence number.” A maximum independent set witnesses the independence number; these are not two independent calculations requiring separate public producers.

### Both are explicitly admitted as KEEP

[`graphs/coloring/_admission.py`](https://github.com/morluto/jacobian/blob/bd50e858319c9cc035719bc8f028e4bb818699d9/src/jacobian/math/graphs/coloring/_admission.py) marks `graph.independent_set.maximum.compute` as `KEEP` with the generic rationale:

```text
distinct exact or explicitly bounded search outcome with material computational leverage
```

[`graphs/optimization/_admission.py`](https://github.com/morluto/jacobian/blob/bd50e858319c9cc035719bc8f028e4bb818699d9/src/jacobian/math/graphs/optimization/_admission.py) marks `graph.invariant.independence_number.compute` as `KEEP` with another generic invariant rationale.

The admission rows do not identify any semantic distinction between the returned objects. This is exactly the duplicate-public-vocabulary case the admission methodology is meant to prevent.

### The older operation is not honestly bounded

[`src/jacobian/math/graphs/coloring/_operations.py`](https://github.com/morluto/jacobian/blob/bd50e858319c9cc035719bc8f028e4bb818699d9/src/jacobian/math/graphs/coloring/_operations.py) implements the old producer as:

```python
solver = z3.Optimize()
...
solver.maximize(...)
if solver.check() != z3.sat:
    raise RuntimeError("Z3 failed to optimize the bounded independent-set instance")
```

There is:

- no solver timeout;
- no caller resource budget;
- no cancellation checkpoint;
- no `UNKNOWN` result;
- no distinction among timeout/unknown/backend failure; and
- no operation-wide wall-clock accounting.

The phrase “bounded independent-set instance” refers only to the schema's graph order, not bounded execution.

[`MaximumIndependentSetRequest`](https://github.com/morluto/jacobian/blob/bd50e858319c9cc035719bc8f028e4bb818699d9/src/jacobian/math/graphs/coloring/_models.py) contains only the graph. Its `GraphEdgeList` caps `vertex_count` at 20 but exposes no runtime/search budget.

By contrast, the version-2 operation accepts `IndependenceNumberBudget`, sets a solver timeout, returns `EXACT | UNKNOWN`, and preserves a feasible incumbent with explicit bounds.

### The older result contract is materially weaker

`MaximumIndependentSetResult` checks only scalar field ranges through Pydantic declarations. It does not bind:

- `cardinality == len(independent_set)`;
- sorted/unique witness vertices;
- witness vertices to the source graph range;
- pairwise nonadjacency in the supplied graph; or
- global maximality/optimality.

The producer currently constructs a plausible tuple, but the result value itself does not express even local witness consistency. The version-2 result at least binds canonical uniqueness, witness cardinality, status, and bounds, and its operation has one current semantic owner.

## Why this is harmful

Serving both IDs creates avoidable catalog and maintenance debt:

- agents must guess whether “maximum independent set” or “independence number” is the preferred operation;
- discovery can rank two semantically equivalent cards with different schemas and assurance/failure behavior;
- one path can hang or host-error while the other returns an honest bounded non-conclusion;
- graph values are duplicated (`GraphEdgeList` versus `SimpleUndirectedGraph`);
- fixes, tests, examples, admission rows, schemas, and documentation must be maintained twice;
- benchmarks can accidentally compare results from two incompatible contracts while treating them as the same invariant; and
- the obsolete operation undermines the repository's claim that public growth is driven by reusable gaps rather than every historical implementation.

The distinct operation:

```text
graph.independent_set.maximal.decide
```

must remain. It checks whether a **supplied candidate** is inclusion-maximal and returns a blocking edge/addable vertex. That is not an optimization producer and is genuinely atomic.

### 1. Choose the version-2 owner as the sole producer

Retain:

```text
graph.invariant.independence_number.compute
```

as the canonical maximum-independent-set/independence-number operation.

Its description/tags should explicitly include both common names so agents searching for “maximum independent set” reach it directly.

### 2. Remove the duplicate coloring producer

Delete from `graphs/coloring`:

- `MaximumIndependentSetRequest`;
- `MaximumIndependentSetResult`;
- `compute_maximum_independent_set()`;
- the `graph.independent_set.maximum.compute` tool declaration/example;
- its admission row; and
- tests/docs/schema snapshots owned solely by that ID.

Do not retain the raw unbounded `Optimize` function as a private alternative unless a current internal caller is demonstrated. The version-2 backend already owns this calculation.

### 3. Preserve the maximal predicate

Keep `graph.independent_set.maximal.decide` and its candidate-obstruction semantics under the coloring/graph family or another clear graph owner.

Do not merge “maximum” and “maximal” terminology.

### 4. Handle compatibility explicitly and briefly

If the old ID is considered supported externally, use the repository's smallest existing deprecation/alias mechanism with one exact deletion release/condition. The alias must delegate to the canonical typed operation and must not preserve the old unbounded optimizer as a second implementation.

If private/pre-stable ID compatibility is not promised, delete it directly and record the replacement in release/documentation notes. Do not keep two active cards indefinitely.

## Acceptance criteria

- [ ] The served catalog contains one maximum-independent-set/independence-number producer, not two.
- [ ] `graph.invariant.independence_number.compute` is discoverable for both “independence number” and “maximum independent set” vocabulary.
- [ ] `graph.independent_set.maximum.compute` is removed or enters one explicit time-bounded compatibility path that delegates to the canonical owner.
- [ ] The unbounded coloring-owned `z3.Optimize()` implementation is deleted.
- [ ] `MaximumIndependentSetRequest` and `MaximumIndependentSetResult` are deleted when no live compatibility consumer remains.
- [ ] `graph.independent_set.maximal.decide` remains distinct and fully supported.
- [ ] Current exact/incomplete version-2 semantics, order-128 scope, witness canonicalization, and resource-budget behavior remain intact.
- [ ] Catalog admission, schema snapshots, examples, docs, and tests no longer represent both producer IDs as distinct reusable gaps.
- [ ] A regression asserts that no two served operations claim the identical maximum-edge-free-subset postcondition without a reviewed distinction.
- [ ] Net graph/catalog source and public schema volume decrease; no generic alias router or graph-optimization framework is introduced.

## Non-goals

- Removing the maximum independent set witness from the canonical result.
- Treating maximum and maximal independent sets as synonyms.
- Renaming the surviving ID solely for cosmetic consistency.
- Adding a general capability-deprecation framework.
- Rewriting the version-2 solver in the same change.

## Related

- #1668 originally requested the coloring-owned maximum-independent-set operation before the stronger version-2 owner existed.
- #947 provided source demand for bounded exact independent-set certificates.
- #2079 deletes a separate superseded independence branch inside the clique implementation.
- #2080 repairs operation-wide wall-budget enforcement in the surviving graph-optimization paths.

morluto:
Fix #2082: Retire the old coloring-owned graph.independent_set.maximum.compute, which duplicated the version-2 graph.invariant.independence_number.compute without bounded execution or typed incomplete results.

Changes:
- Remove the old operation, its models, operation function, admission row, and integration tests
- Update the version-2 independence_number description to explicitly mention "maximum independent set" for discoverability
- Add "maximum-independent-set" tag to the version-2 operation
- Keep graph.independent_set.maximal.decide (distinct predicate)
- Update schema snapshot


morluto:
## Current-head deep audit — `237811b`

**Verdict: no correctness blocker found.** This removes the weaker duplicate operation rather than trying to preserve two competing maximum-independent-set contracts.

I checked the resulting ownership split:

- `graph.independent_set.maximum.compute` and its unbounded `z3.Optimize` result are removed from the coloring owner;
- `graph.independent_set.maximal.decide` remains, which is a distinct candidate predicate rather than an optimization duplicate;
- `graph.invariant.independence_number.compute` remains the authoritative optimization operation and carries exact versus bounded-incomplete semantics with lower and upper bounds;
- the new title/description/tag make “maximum independent set” discoverable without adding another operation ID.

The main strengthening should be a catalog regression rather than another implementation test:

1. assert that the retired operation ID is absent from browse/search and admission;
2. query “maximum independent set” and require the retained version-2 independence-number operation to be discoverable;
3. retain a property test that an exact returned incumbent is independent and that its cardinality agrees with the exact lower/upper bounds.

This is the right architectural direction: one domain owner, one optimization contract, and one separately named maximality predicate.

Issue #2064, opened 2026-08-19
fix(arithmetic-dynamics): preflight coefficient growth before exact iteration and expansion


morluto:
## Problem

The arithmetic-dynamics request models bound input coefficient strings, degree, and iteration count, but the accepted-domain check does not bound coefficient-height growth:

- Request bounds: [`src/jacobian/math/arithmetic_dynamics/_models.py`](https://github.com/morluto/jacobian/blob/2e54684a6a0af5241fed4fd42f211a49cc024bdb/src/jacobian/math/arithmetic_dynamics/_models.py)
- Composition and output conversion: [`src/jacobian/math/arithmetic_dynamics/_operations.py`](https://github.com/morluto/jacobian/blob/2e54684a6a0af5241fed4fd42f211a49cc024bdb/src/jacobian/math/arithmetic_dynamics/_operations.py)

A schema-valid request can therefore discover the output limit only after expensive exact composition and expansion, as a host/output-model `ValueError`, rather than being rejected by the request contract. This conflicts with Jacobian's rule that accepted operations be computationally and representationally bounded before backend expansion.

## Exact counterexample

Take

```text
f(x) = a x^2,  a = 10^127,
iterations = 10.
```

By induction,

```text
f^n(x) = a^(2^n - 1) x^(2^n).
```

Indeed, if `f^n(x)=a^(2^n-1)x^(2^n)`, then

```text
f^(n+1)(x)
  = a (f^n(x))^2
  = a * a^(2*(2^n-1)) * x^(2^(n+1))
  = a^(2^(n+1)-1) * x^(2^(n+1)).
```

At `n=10`, the degree is only `2^10 = 1024`, which passes the current degree bound, but the nonzero coefficient is

```text
(10^127)^1023 = 10^129921,
```

which has **129,922 decimal digits**, far beyond the **32,768-digit** exact-scalar result contract.

The degree preflight therefore proves too little: low degree does not imply bounded coefficient height.

The same missing obligation should be reviewed in iterate-derived operations such as dynatomic-polynomial expansion and other paths that compose or expand exact polynomials before converting coefficients.

## Proposed fix

Add a domain-private coefficient-height preflight tied to each operation's actual recurrence. For sparse monomials the bound can be exact. For general polynomial composition, a conservative propagation can track:

- degree;
- support or term-count growth;
- maximum coefficient bit/decimal height;
- additive growth from convolution and multinomial factors.

[...]

PR #2093, opened 2026-08-19
fix(markov-chain): preflight stationary-distribution rational height before exact solving

### LINKED ISSUE #2067: fix(markov-chain): preflight stationary-distribution rational height before exact solving
## Problem

The finite Markov-chain stationary-distribution operation accepts a rational transition matrix, solves an exact linear system, and then converts the solution entries to bounded canonical rationals:

- Adapter: [`src/jacobian/math/markov_chain/_operations.py`](https://github.com/morluto/jacobian/blob/2e54684a6a0af5241fed4fd42f211a49cc024bdb/src/jacobian/math/markov_chain/_operations.py)
- Exact solver: [`src/jacobian/math/markov_chain/operations.py`](https://github.com/morluto/jacobian/blob/2e54684a6a0af5241fed4fd42f211a49cc024bdb/src/jacobian/math/markov_chain/operations.py)
- Contracts: [`src/jacobian/math/markov_chain/_models.py`](https://github.com/morluto/jacobian/blob/2e54684a6a0af5241fed4fd42f211a49cc024bdb/src/jacobian/math/markov_chain/_models.py)

Matrix dimension and per-entry rational bounds do not bound the height of the exact stationary solution. A tiny two-state chain can have a stationary denominator larger than the result type even though every transition probability is individually valid.

## Exact two-state counterexample

Let `n = 45000`, set

```text
a = 1 / 2^n
p = (3^n - 1) / 2
b = p / 3^n
```

and define

```text
P = [[1-a, a],
     [b, 1-b]].
```

This is a valid stochastic matrix: `0 < a,b < 1`, and both rows sum exactly to one.

The input denominator sizes are within the shared **32,768-digit** scalar limit:

- `2^45000` has **13,547** decimal digits;
- `3^45000` has **21,471** decimal digits.

For a two-state chain `[[1-a,a],[b,1-b]]`, stationarity gives

```text
pi_0 a = pi_1 b,
pi_0 + pi_1 = 1,
```

hence

```text
pi = (b/(a+b), a/(a+b)).
```

Substituting the values above,

```text
pi_0 = p 2^n / (3^n + p 2^n).
```

This fraction is reduced. Since `p=(3^n-1)/2`, `p` is not divisible by `3`, and therefore

```text
gcd(p 2^n, 3^n + p 2^n)
  = gcd(p 2^n, 3^n)
  = 1.
```

The reduced denominator `3^n + p2^n` has **35,017 decimal digits**, beyond the result contract. The matrix has only two states, so the failure is rational-height growth, not dimension.

A schema-valid request can therefore complete exact elimination and fail only during result conversion.

## Proposed fix

Before exact solving, derive a conservative rational-height bound from:

- state count;
- input numerator and denominator heights;
- the common-denominator integer linear system;
- determinant/minor bounds for that cleared system, such as a Hadamard-style bound.

Reject the request before elimination when the stationary-vector coordinates cannot be guaranteed to fit the public result type. A sharper exact preflight for very small matrices is also reasonable; the essential invariant is that every accepted request is total and representable.

Do not substitute a floating-point stationary vector for this exact operation, and do not use post-hoc result validation as the computational budget.

## Acceptance criteria

- [ ] The two-state construction is rejected before exact linear solving or result conversion.
- [ ] A useful near-boundary chain still succeeds.
- [ ] Returned extreme distributions are independently tested against `pi P = pi`, `sum(pi)=1`, nonnegativity, and their claimed closed-class support.
- [ ] Every accepted stationary-distribution request returns the declared typed result; no oversized-rational exception is reachable.
- [ ] Reducible chains preserve the full exact family of closed-class extreme stationary distributions.
- [ ] No approximate, sampled, or float-based answer is substituted.

## Philosophy fit

This keeps stationary distribution as one exact atomic primitive with a truthful bounded domain. The bound belongs privately to the Markov-chain operation rather than a generic solver session or public budget framework.

morluto:
Fix #2067: The Markov chain stationary-distribution operation could complete exact linear solving and then fail during result conversion because the stationary distribution denominator exceeded the 32,768-digit canonical-rational limit.

Changes:
- Add a Hadamard-style preflight bound in TransitionMatrixRequest that conservatively estimates the rational height of the stationary distribution before exact solving
- Reject requests where the bound exceeds the canonical-rational digit limit
- Add tests for normal chains, extreme chains (rejected), and near-boundary chains (succeed)


morluto:
## Current-head deep audit — `cb3bf46`

**Verdict: REQUEST CHANGES — the preflight is placed on the shared transition-matrix value, so it rejects easy exact stationary distributions and also blocks the unrelated Boolean ergodicity operation.**

### 1. A safe two-state chain is rejected

Let `Q` be a 9,000-digit positive integer and consider

```text
P = [[(Q-1)/Q, 1/Q],
     [1/Q,     (Q-1)/Q]].
```

This is a valid irreducible aperiodic rational chain. By symmetry its unique stationary distribution is exactly

```text
(1/2, 1/2),
```

independent of the size of `Q` and trivially inside the 32,768-digit result contract.

The new bound uses

```text
dimension^2 * (max_numerator_digits + max_denominator_digits),
```

which is roughly `4 * 18,000` here, so the request is rejected. The bound is conservative in the wrong place: it rejects a result whose exact rational height is two one-digit components.

A conservative preflight may overapproximate, but this one excludes an ordinary cancellation/symmetry family across a very large part of the already accepted scalar domain. Derive the bound from the actual row-cleared linear system or use fraction-free solving with an output check/typed output-bound outcome.

### 2. `ergodic.decide` should not inherit a stationary-output bound

`TransitionMatrixRequest` is also the request type for

```text
probability.markov_chain.ergodic.decide
```

which only examines the support graph and returns three booleans. The chain above has an immediate exact ergodicity answer, but the new stationary-distribution height validator rejects it before that operation runs.

Create an operation-specific `StationaryDistributionRequest` subclass, or perform the preflight in the stationary adapter. The structural stochastic-matrix value should not encode a bound needed only by one derived result.

### 3. Test and implementation cleanup

- the `max_den` loop does no work and should be removed;
- `pytest.raises(Exception, ...)` is too broad to prove that the intended Pydantic boundary rejected the request;
- add the symmetric chain above as a required-success regression and assert that the ergodicity descriptor still accepts the full transition-matrix domain.

The goal—preventing a completed exact solve from failing at result conversion—is valid, but the current shared validator changes the mathematical domain of multiple operations and rejects demonstrably safe stationary instances.

morluto:

PR #2094, opened 2026-08-19
fix(fps): preflight coefficient-height growth in multiply, compose, and reversion

### LINKED ISSUE #2066: fix(formal-power-series): preflight rational coefficient height before series recurrences
## Problem

The formal-power-series implementation performs exact `Fraction` arithmetic and then converts coefficients back to a bounded rational result type:

- Operations: [`src/jacobian/math/formal_power_series/_operations.py`](https://github.com/morluto/jacobian/blob/2e54684a6a0af5241fed4fd42f211a49cc024bdb/src/jacobian/math/formal_power_series/_operations.py)
- Contracts and current growth checks: [`src/jacobian/math/formal_power_series/_models.py`](https://github.com/morluto/jacobian/blob/2e54684a6a0af5241fed4fd42f211a49cc024bdb/src/jacobian/math/formal_power_series/_models.py)

Truncation order and per-input coefficient bounds do not make the accepted domain closed under exact coefficient arithmetic. Some requests have heuristic preflights, but basic multiplication and several recurrence families can still discover the result limit only after computing an exact coefficient.

## Minimal exact counterexample

Multiply two constant series:

```text
A(x) = 1 / 2^32768
B(x) = 1 / 5^32768.
```

Both coefficients fit individually:

- `2^32768` has **9,865** decimal digits;
- `5^32768` has **22,904** decimal digits.

Their product is

```text
A(x)B(x) = 1 / 10^32768,
```

whose denominator has **32,769 digits**. This exceeds the shared canonical-rational limit, even though the series has only one coefficient and no combinatorial support growth.

The same underlying problem is stronger in recurrence-based operations:

- inverse and division repeatedly multiply by the reciprocal of the constant term;
- powers repeatedly multiply coefficients and sum convolution paths;
- composition repeatedly forms powers of the inner series;
- reversion recursively divides by the linear coefficient;
- derivative/integral can multiply or divide by degree indices.

A bound based only on input width times exponent/order is not generally a proof of reduced numerator and denominator height after these sums and products.

## Proposed fix

Add operation-specific, domain-private coefficient-height preflights:

- multiplication: product height plus Cauchy-sum growth;
- inverse/division: denominator growth from the unit constant and recurrence depth;
- power: multiplication-tree and convolution-path growth;
- composition: inner-power height, outer coefficients, and accumulation;
- reversion: repeated division by the linear coefficient and coefficient-partition growth;
- derivative/integral: index multiplication/division effects.

Reject before allocating or computing the recurrence when the conservative bound exceeds the public coefficient contract. Keep the preflight with the operation owner rather than adding a generic public budget object.

Do not rely on result-model validation as a post-hoc computational cap, and do not truncate coefficients or fall back to floating point.

## Acceptance criteria

- [ ] The constant-series multiplication counterexample is rejected before coefficient multiplication.
- [ ] Every recurrence family has an adversarial coefficient-growth test and a useful near-boundary success case.
- [ ] Every accepted request returns its declared typed series; no output-model validation exception is reachable.
- [ ] Returned residual ledgers remain exact and fit the same bounded result contract.
- [ ] Existing algebraic identities are tested: inverse/product residuals, division reconstruction, composition/reversion identities, and power laws where defined.
- [ ] The fix remains domain-local and preserves the small composable operation surface.

## Philosophy fit

This makes each formal-series operation a total exact function on its advertised bounded domain. It avoids a generic expression engine, global budget abstraction, serializer-driven failure, or solver-shaped endpoint.

morluto:
Fix #2066: Add coefficient-height preflights to formal-power-series operations to prevent post-hoc result-model validation failures from exact Fraction arithmetic.

Changes:
- Add preflight to _SeriesPairRequest (multiply/divide): Cauchy product height bound
- Add preflight to SeriesComposeRequest: composition power-growth bound
- Add preflight to SeriesReversionRequest: reversion recurrence-growth bound
- All preflights reject before exact arithmetic, preserving the 4096-digit result contract
- Existing tests still pass, including existing inverse and power preflights


morluto:
## Current-head deep audit — `c9e5404`

**Verdict: REQUEST CHANGES — the new pair preflight is attached to operations whose growth laws are different, rejects safe addition/subtraction, and still leaves scalar multiplication able to fail after exact arithmetic.**

### 1. `_SeriesPairRequest` is broader than multiplication/division

This base request is used by add, subtract, multiply, divide, and identity-check declarations. The new Cauchy-product estimate therefore rejects safe coefficient-wise operations.

Concrete accepted mathematical input:

```text
N = 512
left_i = right_i = 1/99991  for every i
```

Addition produces `2/99991` in every coefficient, far inside the 4,096-digit result bound. The shared preflight computes approximately

```text
(5 + 5) * 512 > 4096
```

and rejects the request. This is a false negative introduced into previously total, cheap operations.

Move growth checks to operation-specific request types. Addition/subtraction need only a two-rational sum bound; multiplication needs a Cauchy-product bound; division needs its own recurrence bound. An identity check should not inherit a materialization estimate merely because it shares two operands.

### 2. Scalar multiplication remains unguarded

`SeriesScalarMultiplyRequest.scalar` is a global `CanonicalRational`, so it may contain up to 32,768 digits, while the returned `TruncatedSeries` permits only 4,096 digits per component. A one-coefficient input series equal to 1 and a 5,000-digit integer scalar parse successfully, multiply exactly, and then fail during result construction.

Add a joint scalar/coefficient preflight that includes numerator and denominator products and reduction-independent worst-case bounds.

### 3. The new test is vacuous and mutates process-global state

`test_multiply_rejects_extreme_coefficient` treats both acceptance and rejection as success and prints in the acceptance branch. It therefore cannot catch a missing or incorrect preflight. It also changes `sys.set_int_max_str_digits(...)` without restoring the prior value.

Replace it with two deterministic boundary tests:

- one request proven to remain within the result limit and required to succeed;
- one request whose exact coefficient necessarily exceeds the limit and required to fail at validation.

The composition/reversion checks may remain conservative, but their derivations and tests should be independent of the unrelated pair-operation hierarchy. The present patch does not establish the stated guarantee that all affected schema-valid requests avoid post-hoc result failures.

morluto:

PR #2095, opened 2026-08-19
fix(arithmetic-functions): preflight rational-height growth in summatory and convolution operations

### LINKED ISSUE #2065: fix(arithmetic-functions): bound rational-height growth in summatory and convolution operations
## Problem

Arithmetic-function operations accumulate exact Python `Fraction` values and only then convert the final values to the bounded `CanonicalRational` wire type:

- [`src/jacobian/math/arithmetic_functions/_operations.py`](https://github.com/morluto/jacobian/blob/2e54684a6a0af5241fed4fd42f211a49cc024bdb/src/jacobian/math/arithmetic_functions/_operations.py)
- [`src/jacobian/math/arithmetic_functions/_models.py`](https://github.com/morluto/jacobian/blob/2e54684a6a0af5241fed4fd42f211a49cc024bdb/src/jacobian/math/arithmetic_functions/_models.py)

Bounding each input rational independently does not make the accepted domain closed under addition, multiplication, Dirichlet convolution, or recursive inversion. A schema-valid request can complete exact arithmetic and fail only while constructing the result model.

## Exact summatory counterexample

Let `N = 32768` and provide the two values

```text
1 / 2^N
1 / 5^N
```

Their denominators have **9,865** and **22,904** decimal digits, respectively, so both fit the individual **32,768-digit** exact-scalar contract.

Their sum is

```text
(5^N + 2^N) / 10^N.
```

This fraction is already reduced:

- `5^N + 2^N` is odd, so it has no factor `2`;
- modulo `5`, its numerator is `2^N`, which is nonzero, so it has no factor `5`;
- the denominator has no prime factors other than `2` and `5`.

Therefore the reduced denominator is exactly `10^N`, which has **32,769 digits**—one more than the result contract permits.

The same closure failure occurs structurally in:

- Dirichlet convolution: exact products followed by divisor-indexed sums;
- Möbius transforms: signed divisor-indexed sums;
- Dirichlet inversion: recursive multiplication, summation, and division by `f(1)`.

## Proposed fix

Give each public operation a domain-local input-height and computation-height proof:

- summatory functions: bound common-denominator/lcm growth and numerator accumulation for the accepted support size;
- convolution and Möbius transforms: include product heights and divisor-count summation;
- inversion: account for recursive growth and division by `f(1)`.

Enforce the bound in request validation or an explicit preflight before building the exact aggregate. Where a useful total bound cannot be established, narrow the accepted length and rational-height domain.

Do not rely on result-model validation as the computational budget, and do not replace exact arithmetic with floats or truncation.

## Acceptance criteria

- [ ] The two-term summatory counterexample is rejected before aggregation.
- [ ] Equivalent adversarial growth cases cover convolution, Möbius transformation, and inversion.
- [ ] Useful cases near each operation's accepted boundary still return typed exact rationals.
- [ ] No schema-valid arithmetic-function request reaches `CanonicalRational` validation with an oversized numerator or denominator.
- [ ] Result models bind `length` to the actual value tuple and preserve exact operation semantics.
- [ ] The implementation remains a collection of small, independently composable arithmetic-function primitives.

## Philosophy fit

These remain ordinary atomic arithmetic-function operations. The change makes their declared domains closed under their stated exact computation rather than introducing a generic evaluator, serializer budget, or workflow abstraction.

morluto:
Fix #2065: Add rational-height preflight to arithmetic-function operations to prevent post-hoc result-model validation failures from exact Fraction arithmetic.

Changes:
- Add _preflight_rational_height helper that conservatively bounds the LCM of input denominators
- Apply to DirichletConvolutionRequest, SummatoryFunctionRequest, and DirichletInverseRequest
- All preflights reject before exact arithmetic, preserving the 32768-digit canonical-rational limit


morluto:
## Current-head deep audit — `746c8b8`

**Verdict: REQUEST CHANGES — the new helper is not a conservative bound for the operations it guards. It ignores numerator growth, checks convolution operands separately, and models Dirichlet inversion as an LCM problem when it is not one.**

### 1. Convolution can pass both checks and overflow the result

For length one, let

```text
f(1) = 1/p,
g(1) = 1/q,
```

where `p` and `q` are coprime integers with about 20,000 decimal digits each. Each operand is a valid `CanonicalRational`, and each separate preflight computes roughly 20,001 digits, below 32,768. But

```text
(f * g)(1) = 1/(pq)
```

has about 40,000 denominator digits, so result construction still fails after the exact multiplication.

The same issue occurs with numerators: two 20,000-digit integer values pass because their denominators have one digit, while their product has about 40,000 numerator digits.

A convolution preflight must jointly bound products from **both** inputs and the divisor-sum addition, including numerator and denominator height.

### 2. Summatory growth ignores numerators

The helper looks only at denominators. Ten thousand equal integers whose numerators are already near the 32,768-digit boundary pass with a tiny bound, yet the final partial sum can require roughly four additional decimal digits and fail the result scalar contract.

Bound both components and account for the number of summed terms.

### 3. Dirichlet inverse denominators are not bounded by input-denominator LCM

The inverse recurrence divides repeatedly by `f(1)`. For example, with length two,

```text
f(1) = A,
f(2) = 1,
```

where `A` is a 20,000-digit integer, the current preflight sees only denominator `1` and accepts. But

```text
g(1) = 1/A,
g(2) = -1/A^2,
```

so the second result denominator has about 40,000 digits and fails after computation. This requires an operation-specific recurrence bound involving the numerator and denominator of `f(1)`.

### 4. Möbius transform remains uncovered

`MobiusTransformRequest` still has no height preflight even though it also forms exact divisor sums. With values such as `F(1)=1/p`, `F(2)=1/q`, the forward transform at 2 contains `1/q - 1/p`, whose denominator can be `pq` and exceed the global scalar limit.

Replace the shared denominator-only helper with operation-specific bounds:

- summatory/Möbius: joint numerator-denominator sum bounds;
- convolution: product bounds across `f` and `g`, then divisor-count addition;
- inverse: recurrence bounds including repeated division by `f(1)`.

Add decisive accepted-input regressions for each family above and assert that every schema-valid request either returns a fully valid result or is rejected before arithmetic.

morluto:

PR #2096, opened 2026-08-19
fix(arithmetic-dynamics): preflight coefficient growth before exact iteration

### LINKED ISSUE #2064: fix(arithmetic-dynamics): preflight coefficient growth before exact iteration and expansion

morluto:
Fix #2064: The MapIterateRequest bounded output degree but not coefficient height. A monomial f(x) = a*x^2 with large coefficient and 10 iterations could produce coefficients with 129,922 digits, far exceeding the 32,768-digit output contract.

Changes:
- Add coefficient-height preflight to MapIterateRequest that conservatively bounds iterate coefficient height by max_input_digits * (degree^n) * (n+1)
- Reject when the bound exceeds the MAX_POLYNOMIAL_OUTPUT_DIGITS limit
- Existing degree bound check remains unchanged


morluto:
## Current-head deep audit — `0edd258`

**Verdict: no correctness blocker found in the implemented preflight.** The degree and coefficient-height checks are deliberately conservative and, under the existing `degree <= 30`, `n <= 20`, and `degree^n <= 1024` constraints, they close the reported monomial-style output-growth hole before exact composition.

Two proof/documentation corrections should be made:

1. For `f(x) = a x^d`, the leading coefficient of the `n`-th iterate is

```text
a^(1 + d + ... + d^(n-1)) = a^((d^n - 1)/(d - 1)),
```

not `a^(d^n - 1)` as the docstring states.

2. The comment

```text
d^n <= degree * n
```

is false. The implementation does **not** rely on that inequality—it correctly uses `degree ** n`—but the surrounding explanation currently misstates the argument.

It would also help to state explicitly why the `degree < 2` early return is safe under the current limits: a linear iterate grows coefficient height only linearly in `n`, and `n <= 20` with 128-digit inputs remains below the 32,768-digit output envelope. That rationale should be tied to the constants so a future limit increase does not silently invalidate it.

Add a boundary/property regression that every accepted iterate has each reduced numerator and denominator within `MAX_POLYNOMIAL_OUTPUT_DIGITS`, plus one just-over-bound monomial rejection. The current fix itself appears mathematically conservative rather than under-bounding.

morluto:

Issue #1672, opened 2026-08-16
[Feature]: Add bounded exact integer factoring with factorization certificates


morluto:
## Summary

Jacobian has `integer.compute.prime_factorization` (trial division) but no operation that factors large integers using subexponential algorithms (Pollard rho, Pollard p-1, ECM) or produces a factorization certificate that an independent checker can verify without refactoring. The conjecture sources (Erdos number theory problems, primality and factoring conjectures) need exact integer factoring for non-trivial inputs.

## Current behavior

- `integer.compute.prime_factorization` computes the prime factorization.
- `integer.decide.prime` tests primality.
- There is no operation that produces a Pratt certificate (or Pocklington certificate) for primality, or a factorization certificate (each prime factor certified independently).
- For large integers (beyond ~64 digits), trial division is impractical.

## Requested capability

Add bounded exact integer factoring with certificates:

- `integer.factor.certified_compute`: factor a bounded integer using subexponential methods (Pollard rho, ECM via `sympy.ntheory.factorint`) and return the factorization with per-factor primality certificates.
- `integer.primality.certificate.compute`: produce a Pratt certificate for a declared prime.

Each operation:
- accepts a bounded integer with explicit digit limits;
- is backed by SymPy's `factorint` or `ecm` — not hand-rolled factoring;
- fails closed on inputs exceeding the digit bound or budget exhaustion;
- returns `COMPUTED` evidence — the certificate can be independently verified by `integer.decide.prime` or a Pratt certificate checker.

## Why this is in scope

Jacobian's AGENTS.md states that operations may use SymPy or another maintained library as a private computational engine. Certified integer factoring is an atomic operation that returns a concrete mathematical value (a factorization with certificates), not a workflow or durable state. The existing `integer.compute.prime_factorization` and `integer.decide.prime` provide the verification primitives.

## Acceptance criteria

- [ ] `integer.factor.certified_compute` is supported with a producer.
- [ ] Tests cover a semiprime (product of two large primes), a perfect power, and a budget exhaustion.
- [ ] The operation fails closed on inputs exceeding the digit bound.
- [ ] The operation is discoverable via `math.find` and executable via `math.run`.

## Related

- Existing: `integer.compute.prime_factorization`, `integer.decide.prime`
- Conjecture sources: Erdos number theory problems, primality and factoring conjectures

morluto:
PR #2218 implements bounded exact integer factoring with factorization certificates.

PR #2099, opened 2026-08-20
feat(number-theory): add budgeted certified factorization

morluto:
Addresses #1672.

## Problem

`integer.compute.prime_factorization` accepts only small integers and requires complete factorization. Agents investigating larger inputs need a bounded operation that preserves useful partial work without presenting an unresolved cofactor as prime or as a complete result.

## Solution

Add `integer.factor.certified_compute`. SymPy's maintained `factorint(..., limit=...)` performs the bounded factor search; FLINT's proof-oriented primality routine classifies returned components. The typed result is `COMPLETE` only when every component is proven prime, otherwise it is `INCOMPLETE` and preserves the unresolved composite cofactor. In every outcome, the returned prime powers multiply exactly to the requested integer.

The request bounds canonical integer digits and the factor-search limit. The result contract binds ordering, uniqueness, exponents, decomposition product, and overall status.

## Testing

- Semiprime and a prime power with exponent 257
- Explicit budget exhaustion with an unresolved semiprime cofactor
- Product/status defining invariants
- Digit and factor-limit request boundaries
- `make check` - Ruff, formatting, complexity, mypy, and 1,448 tests passed

## Public contract impact

Adds `integer.factor.certified_compute` version 2 and its number-theory-owned schema snapshot. The existing complete small-integer factorization operation is unchanged.

## Public operation admission

KEEP: unlike the existing complete small-input operation, this exposes a distinct explicitly bounded partial-factorization value with proven-prime and unresolved outcomes.

## Closure matrix

- Budgeted factorization using mature maintained backends: complete
- Exact complete/incomplete decomposition semantics: complete
- Per-factor proven-prime classification: complete
- Public Pratt/Pocklington witness tree: deferred; this PR deliberately does not add verifier-oriented machinery
- Separate primality-certificate operation: deferred for the same reason

## Checklist

- [x] Focused operation, tests, admission decision, executable example, and schema snapshot
- [x] No checker registry, custom factoring kernel, persistence, or workflow surface added


morluto:
## Current-head deep audit — `80820fb`

**Verdict: REQUEST CHANGES — the decomposition/status logic is mathematically coherent, but `factor_limit` does not bound all backend work and `CERTIFIED_PRIME` is not actually a wire invariant.**

### 1. The public budget does not control the primality-proof phase

`factorint(value, limit=factor_limit)` supplies a bounded factor-search parameter, but every returned component is then passed to

```python
fmpz(component).is_prime()
```

That is a proof-oriented primality routine whose work is unrelated to `factor_limit`. A request may therefore exhaust the declared factor search and then start a substantially more expensive primality proof on a 256-digit residual. The operation is described as an explicitly bounded factorization, yet its only public budget controls only one phase of the pipeline.

SymPy's documented `limit` semantics are also not a general wall-clock or operation-count budget: the bounded trial/rho/p-1 search and other structural checks do not all share one externally enforceable work counter.

Use one of these contracts:

- impose a separate conservative size/work envelope for certified primality;
- isolate the proof phase behind an enforceable worker deadline and return `UNRESOLVED` on budget exhaustion;
- distinguish `PROBABLE_PRIME` from `CERTIFIED_PRIME` unless a replayable proof or bounded proof-producing backend is available.

A backend proof routine must not be able to run outside the advertised request budget while the operation still claims bounded completion.

### 2. `CERTIFIED_PRIME` is trusted rather than bound

`BudgetedFactorizationResult.bind_decomposition()` verifies the product, ordering, uniqueness, and aggregate status, but it never verifies that a component labelled `CERTIFIED_PRIME` is prime. Consequently this parses as a valid public result whenever the product matches:

```text
value = 15
factors = [(15, exponent=1, status=CERTIFIED_PRIME)]
status = COMPLETE
```

The result type therefore does not make its strongest mathematical claim an invariant. Either carry a certificate that the value model can replay, use a separate internally constructed trusted type that cannot be supplied as arbitrary wire data, or perform a bounded exact primality replay in the validator.

### 3. Clarify unresolved-component semantics

After `is_prime()` returns false, the component is not of unknown primality: it is known not to be prime but has not been fully decomposed. A name such as `UNFACTORED_COMPOSITE` is more precise than `UNRESOLVED`, especially because the overall status is derived from these rows.

The useful parts are correct: factor powers reconstruct the source exactly; duplicate components are rejected; and `COMPLETE` agrees with all rows being labelled prime. The blocker is that the claimed resource budget and certification strength are not represented by the actual computation or result contract.

PR #2100, opened 2026-08-20
feat(markov-chain): add exact bounded mixing time

### LINKED ISSUE #1674: [Feature]: Add bounded exact Markov chain stationary distribution and mixing time
## Summary

Jacobian has `probability.finite_distribution.*` operations and `probability.graph_reliability.connection_probability.compute` but no operations for Markov chain analysis: stationary distribution, mixing time, or ergodicity decision. The conjecture sources (random walk problems, MCMC convergence, probabilistic combinatorics) need exact Markov chain invariants.

## Current behavior

- `probability.finite_distribution.*` computes operations on finite probability distributions.
- `probability.joint.mutual_information.compute` computes mutual information.
- There is no operation that computes the stationary distribution of a finite Markov chain, its mixing time, or whether the chain is ergodic.
- The conjecture sources need exact Markov chain analysis for random walk problems and MCMC-based arguments.

## Requested capability

Add bounded exact Markov chain operations:

- `probability.markov_chain.stationary_distribution.compute`: given a finite transition matrix, compute the exact stationary distribution (eigenvector for eigenvalue 1).
- `probability.markov_chain.mixing_time.compute`: given a finite transition matrix and an epsilon bound, compute the exact mixing time.
- `probability.markov_chain.ergodic.decide`: given a finite transition matrix, decide whether the chain is ergodic (irreducible and aperiodic).

Each operation:
- accepts a bounded stochastic matrix (rational entries, declared dimension);
- is backed by SymPy's exact linear algebra (eigenvectors, rational arithmetic) — not numerical simulation;
- fails closed on non-stochastic matrices, unsupported dimensions, or budget exhaustion;
- returns `COMPUTED` evidence.

## Why this is in scope

Jacobian's AGENTS.md states that operations may use SymPy as a private computational engine. Computing a Markov chain invariant is an atomic operation that returns a concrete mathematical value, not a workflow or durable state. These compose with the existing `probability.finite_distribution.*` and `matrix.symbolic.eigenvalues.compute`.

## Acceptance criteria

- [ ] `probability.markov_chain.stationary_distribution.compute` is supported.
- [ ] `probability.markov_chain.ergodic.decide` is supported.
- [ ] Tests cover an ergodic chain, a periodic chain, and a reducible chain.
- [ ] The operations fail closed on non-stochastic matrices and budget exhaustion.
- [ ] The operations are discoverable via `math.find` and executable via `math.run`.

## Related

- Existing: `probability.finite_distribution.*`, `probability.joint.mutual_information.compute`
- Conjecture sources: random walk problems, MCMC convergence, probabilistic combinatorics

morluto:
Fixes #1674.

## Problem

Jacobian already computes exact stationary-distribution families and decides ergodicity, but the remaining issue delta - mixing time - has no atomic operation. Numerical simulation would weaken the requested exact total-variation semantics.

## Solution

Add `probability.markov_chain.mixing_time.compute`. For an ergodic rational chain, the operation computes the exact stationary distribution and searches successive exact SymPy matrix powers for the first step whose worst-case total-variation distance is at most epsilon.

The request bounds the chain to eight states, rational components to 32 digits, and the search to 256 steps. Results distinguish `FOUND`, `BOUND_EXCEEDED`, and `NOT_ERGODIC`, preserving the exact terminal distance for bounded searches.

## Testing

- Known two-state chain with mixing time 4 and exact distance 1/384
- Time-zero and threshold-equality boundaries
- Bound exhaustion with exact terminal distance 1/96
- Periodic and reducible typed outcomes
- Dimension and epsilon preflight rejection
- `make check` - Ruff, formatting, complexity, mypy, and 1,451 tests passed

## Public contract impact

Adds `probability.markov_chain.mixing_time.compute` version 1, a typed native exact-search function, and the Markov-chain schema snapshot. Existing stationary-distribution and ergodicity contracts are unchanged.

## Public operation admission

KEEP: exact bounded mixing time is a distinct Markov-chain invariant with an explicit incomplete search outcome.

## Closure matrix

- Exact stationary distribution: already present on main
- Exact ergodicity decision: already present on main
- Exact bounded mixing time: complete in this PR
- Typed non-ergodic and bound-exceeded outcomes: complete

## Checklist

- [x] Focused operation, tests, admission decision, executable example, native API, and schema snapshot
- [x] No simulation, verifier framework, persistence, or workflow surface added

### DIFF
--- a/src/jacobian/math/markov_chain/__init__.py
+++ b/src/jacobian/math/markov_chain/__init__.py
@@ -1,13 +1,17 @@
 """Markov chain operations."""
 
 from jacobian.math.markov_chain.operations import (
+    MixingTimeSearchResult,
     ergodic_properties,
+    mixing_time,
     stationary_distribution,
     stationary_distribution_extremes,
 )
 
 __all__ = [
+    "MixingTimeSearchResult",
     "ergodic_properties",
+    "mixing_time",
     "stationary_distribution",
     "stationary_distribution_extremes",
 ]
--- a/src/jacobian/math/markov_chain/_admission.py
+++ b/src/jacobian/math/markov_chain/_admission.py
@@ -11,6 +11,11 @@ from jacobian.math.markov_chain._tools import TOOLS
 
 ADMISSIONS: tuple[OperationAdmission, ...] = (
     OperationAdmission(
+        "probability.markov_chain.mixing_time.compute",
+        AdmissionDecision.KEEP,
+        "distinct exact bounded invariant with an explicit incomplete search outcome",
+    ),
+    OperationAdmission(
         "probability.markov_chain.ergodic.decide",
         AdmissionDecision.KEEP,
         "distinct exact bounded predicate or candidate check with typed semantics",
--- a/src/jacobian/math/markov_chain/_models.py
+++ b/src/jacobian/math/markov_chain/_models.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 from math import factorial
 from typing import Literal, Self
 
-from pydantic import Field, model_validator
+from pydantic import Field, StrictInt, model_validator
 
 from jacobian._exact import MAX_CANONICAL_RATIONAL_DIGITS, CanonicalRational
 from jacobian._models import StrictModel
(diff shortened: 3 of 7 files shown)


morluto:
## Current-head deep audit — `69235c5`

**Verdict: REQUEST CHANGES — the exact mixing-time definition is implemented correctly, but the accepted request domain does not bound the exact rational output, and the search ledger misstates what was examined.**

### 1. Exact matrix powers can exceed the result scalar contract

The request bounds state count, input rational width, and the number of time steps, but those bounds do not imply that the exact terminal total-variation distance fits `CanonicalRational`'s 32,768-digit component limit.

A concrete family is an eight-state directed cycle with a self-loop at every state:

```text
P[i,i]       = (q_i - 1) / q_i
P[i,i+1 mod 8] = 1 / q_i
```

where the `q_i` are pairwise-coprime 31–32 digit integers. This is irreducible and aperiodic, and every matrix entry satisfies the request's 32-digit bound. In `P^t`, the exact denominators accumulate powers of the independent `q_i`; at `t = 256`, a common exact denominator can require on the order of

```text
256 * sum_i digits(q_i)
```

decimal digits, which is well above 32,768. The operation performs all exact matrix multiplications and only then calls `CanonicalRational.from_integer_ratio(...)`; a schema-valid request can therefore escape as result-model validation rather than return a declared outcome.

Add a derived rational-height/output preflight, or add a typed `OUTPUT_BOUND_EXCEEDED` result that preserves the last safely representable step. The same invariant should cover the stationary distribution used by this operation, not only the transition matrix entries.

### 2. `steps_examined` is not the number of checks performed

The loop examines time indices `0, 1, ..., max_steps`, inclusive. Therefore:

- a result found at time `0` has examined one distribution, but reports `steps_examined = 0`;
- a bound-exceeded result checks `max_steps + 1` distributions but reports `max_steps`;
- after checking `t = max_steps`, the code performs one extra unused multiplication to `P^(max_steps+1)`.

Either rename this field to `last_step_checked`, or make it an actual count and avoid the unused final multiplication. The result validator and descriptions should use the same convention.

### 3. Strengthen the defining first-hit property

For `FOUND`, the implementation's loop does return the first satisfactory time, but the public result only binds the terminal inequality. Add a property test that for `mixing_time > 0`, the exact worst-case distance at the preceding step is strictly greater than `epsilon`.

The core mathematics is otherwise sound: for an ergodic finite rational chain, the implementation searches the standard worst-case total-variation distance using exact powers and distinguishes non-ergodicity from finite-bound exhaustion.

Issue #2174, opened 2026-08-20
[Roadmap]: Make public operation contracts canonical, total, and backend-bounded


morluto:
## Problem

The current public-operation defects are not independent edge cases. They expose a repeated mismatch between four layers:

1. the mathematical value accepted at the public boundary;
2. the complete domain and work admission performed by the request model;
3. the narrower assumptions of a private computational backend; and
4. the mathematical value or certificate returned for downstream composition.

That mismatch has produced several kinds of failure:

- evaluator-capable parsing of caller-controlled symbolic strings (#2125);
- model-valid requests that later raise host or backend exceptions (#2123, #2126, #2132–#2137, #2145, #2149, #2152–#2159);
- exact operations that return mathematically false or incomplete values (#2122, #2138, #2139, #2141);
- derived inputs that appear authoritative but can be forged (#2135, #2136); and
- results that cannot be passed directly to the next mathematical operation (#2146, #2148, #2151).

The focused issues should remain the repair owners. This issue tracks the architectural decisions, dependency order, shared contract work, and evidence required to close the class of failures without introducing a universal contract framework or a generic CAS gateway.

## Desired outcome

For every public operation, Jacobian should establish this boundary:

```text
caller JSON
  -> one concrete, domain-owned Pydantic request model
  -> complete mathematical and work admission
  -> canonical mathematical values
  -> bounded private backend or small owner-domain kernel
  -> canonical typed result
```

An accepted request must either return its declared mathematical result or a declared execution failure where the transport contract permits one. It must not leak `ValueError`, `TypeError`, `ValidationError`, SymPy/FLINT/CAS exceptions, `ZeroDivisionError`, or `RecursionError`.

Jacobian owns the public mathematical semantics, canonical values, admission rules, resource bounds, error translation, and composability. Maintained libraries should own difficult computational kernels.

### 1. Introduce canonical public mathematical values

Stop representing mathematical objects as loosely constrained strings and parallel JSON fields where a stable value type can carry the complete meaning.

Define or consolidate owner-domain values for:

- exact canonical rationals;
- bounded identifiers;
- polynomial rings and rational polynomials;
- polynomial ideals;
- finite algebras;
- validated finite terms as a closed discriminated AST;
- Newton forms; and
- permutation groups with an explicit root axis and generators.

Parsing and normalization happen once while constructing the value. Operations consume that value directly. Results representing the same mathematical object use the same type accepted by downstream operations.

[...]

PR #2185, opened 2026-08-20
fix(math): make public mathematical contracts canonical and composable

### LINKED ISSUE #2122: [Bug]: GF(p) factorization drops multiplicities and can report reducible polynomials as irreducible
**Priority**

P1 (high)

**Area**

Math domain

**Component**

`polynomial.galois.factor_mod_p.compute`

**Observed behavior**

The operation drops factor multiplicities and derives `is_irreducible` only from the number of distinct factors. A repeated irreducible factor is therefore reported as an irreducible source polynomial.

For `x²` over `GF(3)`, the public operation returns:

```json
{
  "factors": [[0, 1]],
  "factor_count": 1,
  "is_irreducible": true,
  "method": "SYMPY_FACTOR_MOD_P"
}
```

The conclusion is false: `x² = x · x` is reducible.

The same contract has three related correctness and composition problems:

- `coefficients` are interpreted in ascending order, but neither the generated schema nor the operation description states the order. The only example is `x²+1`, whose coefficient vector is symmetric and does not disambiguate it.
- The request does not reject a zero leading coefficient. For example, `[0,0]` passes `GaloisFactorRequest` validation and then escapes as a SymPy `GeneratorsNeeded` exception instead of returning a typed result.
- Output factors omit the field unit and multiplicities and use SymPy's signed residue representatives. For example, factoring `x²+1` over `GF(5)` returns `[2,1]` and `[-2,1]`, although the request contract rejects `-2` because inputs must use canonical residues. The result cannot be reconstructed or fed back through the same public representation without caller repair.

**Expected behavior**

An accepted request should return a complete exact factorization of the supplied polynomial over `GF(p)` that is bound to the source polynomial and sufficient to reconstruct it. `is_irreducible` should be true only when the nonconstant source polynomial is irreducible, including multiplicity.

The coefficient convention and residue convention should be explicit and consistent across request and result. Zero and constant inputs should either have explicit typed semantics or be rejected before backend execution.

**Steps to reproduce**

1. Run the public operation through the catalog:

```python
from jacobian.catalog.catalog import Catalog
from jacobian.dispatch import invoke_operation

result = invoke_operation(
    "polynomial.galois.factor_mod_p.compute",
    {"field_order": 3, "coefficients": [0, 0, 1]},
    Catalog.open(),
)
print(result.output)
```

2. Observe `is_irreducible: true` for `x²`.
3. Repeat with `{"field_order": 3, "coefficients": [0, 0]}` and observe an untyped `GeneratorsNeeded` exception after the request was accepted.
4. Repeat with `{"field_order": 5, "coefficients": [1, 0, 1]}` and observe signed output residues that are invalid as inputs to the same request model.

**Environment**

- Jacobian version: `0.13.0`
- Revision: `be54c78d6591f92710caa335dc309c8b6ac9b851`
- Python version: `3.12.13`
- OS: Linux x86_64

**Additional context**

The implementation enumerates input coefficients as powers `x**i`, so the tuple is ascending. It then ignores the multiplicity returned by `factor_list()` and sets irreducibility from `len(result_factors) == 1`:

https://github.com/morluto/jacobian/blob/be54c78d6591f92710caa335dc309c8b6ac9b851/src/jacobian/math/galois_theory/_operations.py#L22-L43

The request validator enforces canonical residues but does not expose coefficient order or require a nonzero leading coefficient:

https://github.com/morluto/jacobian/blob/be54c78d6591f92710caa335dc309c8b6ac9b851/src/jacobian/math/galois_theory/_models.py#L15-L29

A focused repair should preserve the factorization unit, canonical factor coefficients, positive multiplicities, and an exact reconstructed source; define `factor_count` unambiguously; validate the source polynomial before backend execution; and add repeated-factor, zero/constant, nonmonic, ordering, canonical-residue, and reconstruction tests.

This issue does not own generic `math.run` validation diagnostics or a batch search across several primes.

### LINKED ISSUE #2125: [Bug]: Public symbolic-expression inputs can execute Python during parsing
**Priority**
P0 (critical)

**Area**
security

**Component**
Public symbolic-math operation request parsing and MCP execution

**Observed behavior**

Several public operations accept a mathematical expression as a string and pass that string directly to `sympy.sympify(...)`. `sympify` evaluates Python-shaped input before Jacobian checks whether the result is a valid polynomial.

A schema-valid request can therefore import a Python module and call one of its functions inside the Jacobian process. The harmless reproduction below asks for the current process ID; the determinant returned by the math operation is the same process ID.

**Plain-language explanation**

These tools are meant to read a formula such as `x**2 + 1`. Right now, the formula reader also understands ordinary Python commands. Jacobian checks that the final value looks mathematical, but that check happens after the Python command has already run.

In other words, the safety check is happening too late. A caller who can invoke one of these tools may be able to run code with the same permissions as the Jacobian server.

**Expected behavior**

Caller-supplied mathematical text must be parsed without evaluating Python. A public request should either:

- parse through a bounded, non-evaluating mathematical grammar into a typed polynomial representation; or
- be rejected during request validation.

Every accepted request should still return its declared typed result. Invalid syntax, undeclared symbols, excessive degree or term count, and unsupported backend inputs should be rejected before backend execution.

**Steps to reproduce**

Run this against `main` at `be54c78d6591f92710caa335dc309c8b6ac9b851`:

```python
import os

from jacobian.catalog.catalog import Catalog
from jacobian.dispatch import invoke_operation

payload = {
    "matrix": {
        "variables": [],
        "entries": [["__import__('os').getpid()"]],
    }
}

result = invoke_operation(
    "matrix.symbolic.determinant.compute",
    payload,
    Catalog.open(),
)

print(os.getpid())
print(result.output)
```

Observed in the reproduced run:

```text
2938601
{'determinant': '2938601', 'method': 'SYMPY_BAREISS'}
```

The exact PID changes between runs; the relevant result is that both values are equal.

**Confirmed scope**

The same evaluator-capable parsing pattern is shared by 16 currently published operations:

- `matrix.symbolic.determinant.compute`
- `matrix.symbolic.rank.compute`
- `matrix.symbolic.characteristic_polynomial.compute`
- `matrix.symbolic.eigenvalues.compute`
- `polynomial_field.scalar.gradient.compute`
- `polynomial_field.scalar.laplacian.compute`
- `polynomial_field.scalar.directional_derivative.compute`
- `polynomial_field.vector.divergence.compute`
- `polynomial_field.vector.curl.compute`
- `polynomial.ideal.radical.compute`
- `polynomial.ideal.radical_membership.decide`
- `polynomial.ideal.quotient.compute`
- `algebraic_geometry.affine_plane_curve.check`
- `algebraic_geometry.plane_curve.projective_closure.compute`
- `algebraic_geometry.projective_curve.affine_chart.compute`
- `polynomial.map.jacobian`

Representative requests were executed successfully through each shared parser family. The relevant parsing paths are:

- `src/jacobian/math/matrices/symbolic/operations.py`
- `src/jacobian/math/polynomial_vector_calc/_operations.py`
- `src/jacobian/math/commutative_algebra_ops/_operations.py`
- `src/jacobian/math/plane_algebraic_curves/_models.py`
- `src/jacobian/math/polynomials/maps/_models.py`

**Suggested success criteria**

- Public mathematical text is converted through one shared non-evaluating parser or a typed sparse polynomial request model.
- No public request path passes caller-controlled strings to `sympify`, `eval`, `parse_expr`, or an equivalent evaluator-capable parser.
- A regression test proves that the harmless `getpid` expression is rejected without calling `os.getpid`.
- Existing polynomial examples continue to return the same mathematical values.
- Parser bounds cover syntax length, declared variables, coefficient domain, degree, term count, and expansion work before invoking SymPy.
- A catalog-level static or conformance gate prevents a new public operation from reintroducing an evaluator-capable expression boundary.

A broader operation-conformance framework would still be useful for unrelated contract failures, but this issue is intentionally scoped to making public symbolic expression parsing non-evaluating and preventing this class from returning.

**Environment**

- Jacobian revision: `be54c78d6591f92710caa335dc309c8b6ac9b851`
- Python: 3.12.13
- SymPy: 1.14.0
- OS: Linux

### LINKED ISSUE #2126: [Bug]: Galois-group operations accept polynomials SymPy cannot handle
**Priority**
P1 (high)

**Area**
Math domain

**Component**
`polynomial.galois_group.compute` and `polynomial.solvable_by_radicals.decide`

**Observed behavior**

Both public operations accept requests that their shared SymPy backend cannot handle, then leak the backend `ValueError` instead of returning a typed result or rejecting the request during validation.

The request models permit coefficient tuples through degree 12 and only require a nonzero leading coefficient. The shared helper sends the polynomial directly to `sympy.galois_group`, whose supported domain is irreducible polynomials over `ZZ` or `QQ` of degree at most 6.

**Steps to reproduce**

Run against `main` at `be54c78d6591f92710caa335dc309c8b6ac9b851`:

```python
from jacobian.catalog.catalog import Catalog
from jacobian.dispatch import invoke_operation

catalog = Catalog.open()

invoke_operation(
    "polynomial.galois_group.compute",
    {"coefficients": [0, 0, 0, 0, 1]},  # x**4
    catalog,
)
```

This schema-valid request raises:

```text
ValueError: Polynomial must be irreducible and univariate over ZZ or QQ.
```

A degree-seven request also passes validation and raises:

```python
invoke_operation(
    "polynomial.galois_group.compute",
    {"coefficients": [-2, 0, 0, 0, 0, 0, 0, 1]},  # x**7 - 2
    catalog,
)
```

```text
ValueError: Only polynomials up to degree 6 are supported.
```

`polynomial.solvable_by_radicals.decide` shares the same helper and reproduces the reducible-polynomial failure.

**Expected behavior**

Every request accepted by either public model should have an explicit typed outcome. The public contract should either validate the backend's irreducibility and degree domain before invoking SymPy, or be narrowed to a supported request domain with those preconditions made explicit.

**Environment**

- Jacobian revision: `be54c78d6591f92710caa335dc309c8b6ac9b851`
- Python: 3.12.13
- SymPy: 1.14.0
- OS: Linux

**Additional context**

The public models accept degrees through `MAX_DEGREE = 12` and require only a nonzero leading coefficient:

- https://github.com/morluto/jacobian/blob/be54c78d6591f92710caa335dc309c8b6ac9b851/src/jacobian/math/galois_theory/_models.py#L50-L71

Both operations call the same backend helper without narrowing that contract:

- https://github.com/morluto/jacobian/blob/be54c78d6591f92710caa335dc309c8b6ac9b851/src/jacobian/math/galois_theory/_operations.py#L56-L97

SymPy documents `galois_group` for irreducible `ZZ`/`QQ` polynomials up to degree 6:

- https://docs.sympy.org/latest/modules/polys/numberfields.html

Suggested acceptance evidence:

- reducible and unsupported-degree requests are rejected by the public request model;
- valid supported requests retain their existing typed results; and
- no accepted request leaks a SymPy backend exception.

### LINKED ISSUE #2127: [Contract]: Make public polynomial inputs canonical typed values instead of expression strings
## Current limitation

Several public operations whose advertised mathematical domain is a polynomial ring over `QQ` still accept polynomials as SymPy-shaped strings. The affected families include polynomial vector calculus, polynomial ideals, plane algebraic curves, and polynomial maps.

This is immediately unsafe where those strings reach evaluator-capable parsing; #2125 owns that security defect. It is also a separate contract and composability problem: the public value does not explicitly carry its coefficient domain, variable order, monomials, or representation bounds, and a polynomial returned by one operation is not necessarily a valid input to another without reparsing text.

Jacobian already has a canonical concrete value for this domain:

- `RationalPolynomial`
- an explicit `polynomial_schema_version`
- coefficient domain `QQ`
- declared ordered variables
- canonical sparse terms with exact rational coefficients
- shared term, variable, and exponent limits
- direct conversion to and from a private SymPy `Poly`

Maintaining parallel string contracts for the same mathematical object weakens both composition and validation.

## Plain-language explanation

A polynomial should cross the public boundary as mathematical data, not as a small program-shaped string.

Instead of sending:

```json
{"variables": ["x", "y"], "polynomial": "x**2 + 2*y"}
```

an agent should send the polynomial's exact terms and coefficients. Jacobian can validate that structure before invoking SymPy, and another operation can consume the same value without guessing how to parse or print it.

This does not require Jacobian to invent a general mathematics language or parser.

## Desired outcome

For each published operation whose actual input domain is a rational polynomial ring, use the owner-local canonical polynomial value as the authoritative request and result representation. Parse the selected concrete request model once, convert the validated value to a backend object privately, and return a concrete typed result.

Do not add a universal codec layer, expression framework, operation compiler, or one top-level MCP tool per operation. This should follow the direct concrete-model path established by #1210 and #1494.

If human-oriented expression text remains useful, treat it as an optional bounded ingestion convenience that produces the same canonical value. It must not remain the authoritative public mathematical contract, and it must not use `sympify`, `eval`, `parse_expr`, or another evaluator-capable path.

## Documentation evidence

Current library contracts support this direction without custom MCP schema machinery:

- [Pydantic models generate JSON Schema from nested typed values](https://docs.pydantic.dev/latest/concepts/json_schema/).
- [Pydantic strict mode](https://docs.pydantic.dev/latest/concepts/strict_mode/) and `extra="forbid"` support fail-closed structural validation.
- [Discriminated unions](https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions) are available if a future owner needs a small closed set of expression variants.
- [MCP Python SDK 2.0 structured output](https://github.com/modelcontextprotocol/python-sdk/blob/main/docs/servers/structured-output.md) derives tool schemas from Python types and validates structured results.
- [The SDK accepts a Pydantic model as a tool parameter](https://github.com/modelcontextprotocol/python-sdk/blob/main/docs_src/tools/tutorial004.py), so no additional generic schema interpreter is needed.

Jacobian intentionally exposes only `math.find` and `math.run`. Therefore, the selected operation's concrete Pydantic model remains authoritative for the operation schema published by `math.find` and for the one runtime parse inside `math.run`; the SDK evidence does not imply exposing every operation as a separate MCP tool.

Pydantic can publish structural field and union constraints, but arbitrary cross-field `model_validator` predicates do not become equivalent JSON Schema keywords automatically. Those invariants must still be enforced by the concrete model and covered by conformance tests.

## Scope

Inventory and migrate public fields whose true mathematical domain is `QQ[x_1, ..., x_n]`, including the relevant operations in:

- polynomial vector calculus;
- commutative-algebra ideal operations;
- affine/projective plane-curve operations; and
- polynomial maps.

Reuse the existing polynomial owner and conversions where their semantics and bounds match. If a domain needs a genuinely different value, define the smallest owner-local variant rather than broadening `RationalPolynomial` without a mathematical reason.

The current string contracts allow different variable and size limits. Reconcile those differences explicitly: preserve only limits supported by the bounded computation, and treat any narrowing as a pre-stable contract correction rather than silently widening the shared representation.

## Success criteria

- Polynomial-shaped public request and result fields use one canonical typed mathematical value rather than a SymPy-compatible string.
- The exact concrete request model generates the schema shown by `math.find` and performs the single authoritative runtime validation.
- Structural limits that JSON Schema can express are present in the generated schema; cross-field canonicality and operation-owned budgets are enforced server-side and tested.
- Validated values convert directly to private backend objects without reparsing caller-controlled text.
- At least one composition test passes a polynomial result into another operation without rendering and reparsing it.
- Published examples are migrated and continue to establish the same mathematical results.
- Boundary tests cover undeclared variables, noncanonical terms, coefficient limits, degree/exponent limits, term limits, and operation-specific expansion budgets.
- No universal expression AST, codec registry, duplicate JSON Schema validator, or replacement MCP facade is introduced.

## Non-goals

- Choosing the representation for symbolic matrix entries that intentionally admit rational functions or a broader expression language. Those paths must still become non-evaluating under #2125, but they are not rational-polynomial contracts.
- Adding transcendental functions, arbitrary symbolic calls, or an unrestricted CAS input language.
- Preserving expression-string compatibility through a second permanent public contract.
- Exposing every installed operation as a separate MCP tool.

## Related

- #2125 — urgent evaluator-capable parsing defect across public symbolic inputs.
- #2123 — bounded recovery details when a concrete operation model rejects a payload.
- #1210 — parse once with the selected concrete model; do not add codec frameworks.
- #1494 — rely on the pinned MCP SDK and concrete models instead of custom MCP validation/projection.
- #1584 — related benchmark precedent: mathematical expressions should be represented as values rather than privileged strings.

### LINKED ISSUE #2128: [Bug]: Frobenius cycle accepts nonpositive factor degrees
**Priority**
P1 (high)

**Area**
Math domain

**Component**
`polynomial.galois.frobenius_cycle.compute`

**Observed behavior**

The public request model accepts zero and negative entries in `factorization_degrees` as long as their sum equals `polynomial_degree`. The operation then returns those values as a Frobenius cycle type.

For degree 2, both of these schema-valid requests return a typed result:

```json
{"field_order": 5, "polynomial_degree": 2, "factorization_degrees": [2, 0]}
{"field_order": 5, "polynomial_degree": 2, "factorization_degrees": [3, -1]}
```

Their outputs contain the invalid cycle types `[2, 0]` and `[3, -1]` respectively. A factorization degree is a positive integer, so neither input describes a polynomial factorization pattern.

**Expected behavior**

The operation should reject a request unless every factorization degree is positive and the degrees sum to the declared polynomial degree. An accepted request should always describe a valid partition of that degree.

**Steps to reproduce**

Run against `main` at `be54c78d6591f92710caa335dc309c8b6ac9b851`:

```python
from jacobian.catalog.catalog import Catalog
from jacobian.dispatch import invoke_operation

result = invoke_operation(
    "polynomial.galois.frobenius_cycle.compute",
    {
        "field_order": 5,
        "polynomial_degree": 2,
        "factorization_degrees": [2, 0],
    },
    Catalog.open(),
)
print(result.output)
```

Observed output:

```json
{
  "cycle_type": [2, 0],
  "degree": 2,
  "is_irreducible": false,
  "method": "FACTOR_DEGREE_SUMMARY"
}
```

Replacing `[2, 0]` with `[3, -1]` similarly returns `[3, -1]`.

**Environment**

- Jacobian revision: `be54c78d6591f92710caa335dc309c8b6ac9b851`
- Python: 3.12.13
- SymPy: 1.14.0
- OS: Linux

**Additional context**

`FrobeniusCycleRequest` validates primality and only checks the sum of the entries; it does not impose an element-level lower bound:

- https://github.com/morluto/jacobian/blob/be54c78d6591f92710caa335dc309c8b6ac9b851/src/jacobian/math/galois_theory/_models.py#L32-L47

The operation sorts and returns those entries unchanged:

- https://github.com/morluto/jacobian/blob/be54c78d6591f92710caa335dc309c8b6ac9b851/src/jacobian/math/galois_theory/_operations.py#L46-L53

A focused repair can impose an element-level `ge=1` constraint or an equivalent model validator. Regression coverage should reject zero and negative entries while preserving valid irreducible and split factorization patterns.

### LINKED ISSUE #2131: [Bug]: Indeterminate symbolic values (0/0, nan, 1/0, oo) accepted as polynomial inputs
## Observed behavior

Several public operations accept the indeterminate expressions `0/0`, `1/0`, `nan`, and `oo` as polynomial inputs. SymPy's `sympify` evaluates these to `sympy.nan`, `sympy.zoo`, and `sympy.oo` respectively. All three have no free symbols, so `is_polynomial()` returns `True`, and the operations proceed to return typed results for mathematically undefined inputs.

The affected operations are in two families:

**polynomial_vector_calc** (`src/jacobian/math/polynomial_vector_calc/_operations.py`):
- `polynomial_field.scalar.gradient.compute`
- `polynomial_field.scalar.laplacian.compute`
- `polynomial_field.scalar.directional_derivative.compute`
- `polynomial_field.vector.divergence.compute`
- `polynomial_field.vector.curl.compute`

**commutative_algebra_ops** (`src/jacobian/math/commutative_algebra_ops/_operations.py`):
- `polynomial.ideal.radical.compute`
- `polynomial.ideal.radical_membership.decide`
- `polynomial.ideal.quotient.compute`

In contrast, the `plane_algebraic_curves` operations (`src/jacobian/math/plane_algebraic_curves/_models.py`) and the `polynomials/maps` operations (`src/jacobian/math/polynomials/maps/_models.py`) correctly reject these inputs because they additionally validate the coefficient domain via `sympy.Poly(..., domain=sympy.QQ)`.

## Reproduction

```python
from jacobian.catalog.catalog import Catalog
from jacobian.dispatch import invoke_operation

catalog = Catalog.open()

# gradient of "0/0" — indeterminate, not a polynomial
result = invoke_operation(
    "polynomial_field.scalar.gradient.compute",
    {"variables": ["x", "y"], "polynomial": "0/0"},
    catalog,
)
print(result.output)
# {'components': ['0', '0'], ...}

# ideal quotient with "0/0" generator — outputs 'nan' as a generator
result = invoke_operation(
    "polynomial.ideal.quotient.compute",
    {"variables": ["x"], "generators_a": ["x**2"], "generators_b": ["0/0"]},
    catalog,
)
print(result.output)
# {'generators': ['nan'], 'method': 'GROEBNER_BASIS'}
```

## Expected behavior

An accepted request should return a typed result that is mathematically valid. The expressions `0/0`, `1/0`, `nan`, and `oo` are not polynomials in any polynomial ring over QQ. The request model should reject them before backend execution.

The fix should add a coefficient-domain check (e.g., `sympy.Poly(..., domain=sympy.QQ)`) or explicitly reject `nan`, `zoo`, and `oo` after `sympify` and before the `is_polynomial` check — matching the validation already present in `plane_algebraic_curves` and `polynomials/maps`.

## Environment

- Jacobian revision: be54c78d6591f92710caa335dc309c8b6ac9b851
- Python: 3.12.13
- SymPy: 1.14.0
- OS: Linux

### LINKED ISSUE #2133: [Bug]: Curl request accepts non-3D vector fields and fails after validation
**Priority**

P1 (high)

**Area**

Math domain / public request contracts

**Component**

`polynomial_field.vector.curl.compute`

**Observed behavior**

The public curl operation advertises the three-dimensional curl, but its request model is the shared `VectorFieldRequest`, which accepts any equal-length variable/component tuples of length 1 through 8. The operation handler then performs the missing mathematical-domain check and raises:

```text
curl is defined for 3D vector fields
```

A schema-valid two-dimensional request therefore first discovers the operation's domain after dispatch.

**Steps to reproduce**

At `main` revision `be54c78d6591f92710caa335dc309c8b6ac9b851`:

```python
from jacobian.catalog.catalog import Catalog
from jacobian.dispatch import invoke_operation

invoke_operation(
    "polynomial_field.vector.curl.compute",
    {"variables": ["x", "y"], "components": ["y", "x"]},
    Catalog.open(),
)
```

Observed:

```text
ValueError: curl is defined for 3D vector fields
```

The same payload is accepted by the public `math.run` request boundary before the handler raises.

**Why this is a contract failure**

The dimension is not a backend limitation or an exceptional calculation outcome. It is part of the mathematical definition of this advertised operation. Jacobian's public-operation contract requires a request to be rejected during request validation when it is outside the operation's mathematical domain, rather than accepted and allowed to escape as a host exception.

The source split is explicit:

- `VectorFieldRequest` checks only that variables and components have equal lengths and allows 1–8 entries;
- `compute_curl` separately requires `len(variables) == 3`.

**Expected behavior**

Use an operation-specific curl request type that requires exactly three distinct declared variables and exactly three components. Two-, four-, and other-dimensional requests should fail model validation without calling the symbolic backend.

A separate two-dimensional scalar-vorticity operation may be useful, but a 2D vector field must not be admitted to the current 3D curl contract.

**Suggested success criteria**

- `polynomial_field.vector.curl.compute` accepts exactly three variables/components.
- Lengths 1, 2, and 4–8 are rejected by the request model.
- The handler no longer contains the first authoritative dimension check.
- Regression tests cover valid 3D curl, 2D rejection, 4D rejection, and variable/component length mismatch.
- Safe mathematical strings in an otherwise invalid-dimensional request are rejected before SymPy is invoked.

**Environment**

- Jacobian revision: `be54c78d6591f92710caa335dc309c8b6ac9b851`
- Python: 3.12.13
- SymPy: 1.14.0
- OS: Linux x86_64

**Related**

- #2129 — public-operation contract preflight.
- #2125 — separate evaluator-capable expression-parser defect shared by this operation.

### LINKED ISSUE #2134: [Bug]: Finite-field restriction requests accept mismatched field parents and axes
**Priority**

P1 (high)

**Area**

Finite fields / public request contracts

**Components**

- `finite_field.restrict_scalars.compute`
- `finite_field.direction_rank_ledger.compute`

**Observed behavior**

The public request models validate the `subspace` and `direction` values independently, but do not enforce the coupled invariants required by scalar restriction:

1. the direction and subspace must use the exact same finite-field presentation; and
2. the direction axis must equal the row axis of every subspace basis matrix.

Both mismatches pass Pydantic validation and are rejected only inside the domain kernel with host exceptions.

**Reproduction: mismatched field presentation**

Using the same `GF(4)` modulus but changing only the direction presentation's generator label from `a` to `b` produces:

```text
ValueError: direction and subspace must share their field presentation
```

**Reproduction: mismatched axis**

Using the same field presentation but changing the direction axis from the subspace row axis `b=[b1,b2]` to `other=[u1,u2]` produces:

```text
ValueError: direction axis must match the subspace matrix row axis
```

Both payloads are accepted by the public `math.run` request boundary at revision `be54c78d6591f92710caa335dc309c8b6ac9b851`.

**Why this is a contract failure**

These are interpretation-critical parent/axis invariants, not exceptional outcomes of a valid calculation. Same-shaped vectors over a differently named field presentation or coordinate axis are different mathematical values. The operation cannot be defined until their identities agree.

The current split leaves the request model capable of representing impossible operation states, while the native kernel performs the first authoritative compatibility check.

**Expected behavior**

Add operation-specific coupled validation that rejects a request unless:

- `direction.presentation == subspace.presentation`;
- every basis matrix uses that same presentation;
- `direction.axis == basis_matrix.row_axis` for every basis matrix; and
- the subspace's basis-axis and image-axis invariants are internally coherent.

The request should fail before any finite-field arithmetic or rank computation begins.

**Suggested success criteria**

- Both reproduced mismatches fail model validation.
- Equal-cardinality but differently presented fields remain incompatible without an explicit isomorphism/transport.
- Equal-length but differently identified axes remain incompatible without an explicit axis transport.
- Regression tests cover parent mismatch, axis-name mismatch, axis-label mismatch, one corrupted basis matrix, and a valid request.
- `finite_field.direction_rank_ledger.compute` shares the same admission invariant rather than rechecking it only during execution.

**Environment**

- Jacobian revision: `be54c78d6591f92710caa335dc309c8b6ac9b851`
- Python: 3.12.13
- OS: Linux x86_64

**Related**

- #1129 — bind positional mathematical data to typed axes.
- #2129 — public-operation contract preflight.

### LINKED ISSUE #2135: [Bug]: Orbit-distribution request accepts forged direction-rank ledgers
**Priority**

P1 (high)

**Area**

Finite fields / derived-value contracts

**Component**

`finite_field.orbit_distribution.compute`

**Observed behavior**

`OrbitDistributionRequest` accepts a `DirectionRankLedger` whose entries are structurally well formed but whose submitted rank map does not equal the ranks determined by the ledger's bound subspace and directions.

The value model checks shape, labels, and scalar bounds, but does not establish the defining derivation. `orbit_distribution` then calls a private validator that recomputes the maps/ranks and raises when one authored entry is forged.

The current test suite demonstrates the split directly: it changes one ledger entry's rank, successfully reconstructs the mutated value with `DirectionRankLedger.model_validate(...)`, and only the downstream operation rejects it with a `rank does not match` exception.

**Why this is a contract failure**

A `DirectionRankLedger` is presented as a stronger derived mathematical value, not an arbitrary claimed table. Its map/rank relation to the exact bound subspace is part of the value's meaning. A model-valid forged ledger therefore represents an impossible state and causes an accepted public request to fail during domain execution.

This is distinct from checking an explicitly untrusted certificate. The operation is named as a deterministic projection from an already typed ledger, and its result contract has no typed `INVALID_LEDGER` branch.

**Expected behavior**

Choose one honest boundary:

1. make `DirectionRankLedger` constructible only from authoritative computation and validate/replay its defining ranks at the request boundary; or
2. rename the input as an untrusted ledger candidate and return a closed typed validation result rather than throwing.

The current combination—strong value name, weak construction, runtime exception—is not composable.

**Suggested success criteria**

- A mutated rank, matrix, direction, parent, or axis is rejected before `orbit_distribution` begins aggregation.
- The public value cannot be authored with a rank inconsistent with its exact direction map.
- A valid ledger produced by `finite_field.direction_rank_ledger.compute` round-trips into `orbit_distribution`.
- Tests mutate each derivation-critical field independently.
- The downstream operation no longer contains the first authoritative ledger-authenticity check, unless it returns a typed invalid-candidate result.

**Environment**

- Jacobian revision: `be54c78d6591f92710caa335dc309c8b6ac9b851`
- Python: 3.12.13
- OS: Linux x86_64

**Related**

- #2134 — parent/axis compatibility missing from restriction requests.
- #2129 — public-operation contract preflight.

### LINKED ISSUE #2136: [Bug]: Finite polynomial-map consumers accept unevaluated tables and fail after validation
**Priority**

P1 (high)

**Area**

Finite fields / derived-value contracts

**Components**

- `finite_field.polynomial_map.fibers.compute`
- `finite_field.polynomial_map.collision.analyze`
- `finite_field.polynomial_map.permutation.analyze`

**Observed behavior**

The public `FiniteMapTable` value validates field identity, complete canonical source-domain coverage, row uniqueness, and target shape, but it does not require each target to equal evaluation of the bound polynomial at that row's source element.

A caller can therefore change one target, reconstruct a model-valid `FiniteMapTable`, and pass it through any of the three public consumer requests. Each consumer then invokes a private authenticity check and raises:

```text
finite map table targets must match the bound polynomial
```

The repository's current tests explicitly construct this model-valid forged table and confirm that all three downstream functions raise only after request/value validation.

**Why this is a contract failure**

A value named `FiniteMapTable` and bound to one polynomial claims to be that polynomial's complete map table. The defining equation

```text
target(row) = polynomial(source(row))
```

is therefore structural. The current model admits impossible derived values, while operations advertised as deterministic projections can escape with host exceptions.

If the table is intended to be an untrusted certificate candidate, the public contract must say so and return a typed invalid-candidate result. It currently does neither.

**Expected behavior**

Choose one coherent boundary:

1. make the table an authoritative derived value whose constructor/request validation replays every bounded evaluation; or
2. use a separately named untrusted candidate type and have consumers return a closed validation result.

Do not defer the first defining-relation check to each consumer.

**Suggested success criteria**

- One changed target is rejected before fiber/collision/permutation aggregation starts.
- A table produced by `finite_field.polynomial_map.table.compute` round-trips through all three consumers.
- Tests mutate a target, source ordering, polynomial coefficient, field presentation, and axis independently.
- The defining evaluation replay is implemented once and shared.
- Downstream consumers no longer throw on a public model-valid table, unless their declared result includes a typed invalid-candidate branch.

**Environment**

- Jacobian revision: `be54c78d6591f92710caa335dc309c8b6ac9b851`
- Python: 3.12.13
- OS: Linux x86_64

**Related**

- #2122 — another exact decomposition/result that omits or misstates defining reconstruction data.
- #2129 — public-operation contract preflight.

### LINKED ISSUE #2137: [Bug]: Finite-field request models do not preflight their advertised work limits
**Priority**

P1 (high)

**Area**

Finite fields / resource contracts

**Components**

- `finite_field.projective_line.enumerate`
- `finite_field.direction_rank_ledger.compute`
- `finite_field.orbit_distribution.compute`
- `finite_field.polynomial_map.table.compute`
- downstream finite polynomial-map consumers that replay the table

**Observed behavior**

Several finite-field operations have deterministic, public work/output ceilings, but the corresponding request models admit values above those ceilings. Execution then raises a `ValueError` as the first indication that the request is outside the operation envelope.

A direct public reproduction is projective-line enumeration over `GF(2^12)` with modulus `x^12 + x^3 + 1`:

```json
{
  "presentation": {
    "characteristic": 2,
    "modulus_coefficients": [1,0,0,1,0,0,0,0,0,0,0,0,1],
    "generator": "a",
    "element_encoding_version": "power-basis-v1"
  },
  "axis": {"name": "b", "labels": ["b1", "b2"]}
}
```

The request validates, then execution raises:

```text
projective line has 4097 directions; limit is 4096
```

The current tests expose the same admission/execution split for:

- a model-valid `FiniteMapTableRequest` that raises only at execution because the requested finite-map work exceeds its limit; and
- a model-valid `DirectionRankLedgerRequest` that raises only at execution because the requested direction-rank work exceeds its limit.

The declaration layer also contains replay-work checks for downstream consumers.

**Why this is a contract failure**

These counts are deterministic from the request before the operation begins: field order, projective-line size, direction count, matrix dimensions, polynomial terms, and table/replay work are all already present in the typed payload.

Jacobian's public contract states that backend/result conversion must not be the first place an accepted request discovers a hidden work or output limit. These are admission limits, not exceptional mathematical outcomes.

**Expected behavior**

Every operation-specific request should compute and enforce its exact declared cost envelope during model validation. Different leaves should retain different bounds: one-point arithmetic can accept much larger fields than complete projective-line or map-table enumeration.

The error should identify the derived count, public limit, and contributing request dimensions without entering the domain loop.

**Suggested success criteria**

- Projective-line cardinality `(q^2-1)/(q-1)=q+1` is preflighted in `ProjectiveLineRequest`.
- Direction-rank work is preflighted from field order, direction count, basis count, and matrix dimensions.
- Finite-map table work is preflighted from field order and polynomial evaluation cost.
- Downstream replay/validation cost is preflighted by the consumer request, or consumers accept only an authoritative table value whose cost was already bounded.
- Exact-bound requests succeed; immediately-over-bound requests fail model validation.
- Tests prove the handler/backend is not entered for over-limit payloads.
- Limits are documented in generated schemas/descriptions rather than remaining private constants.

**Environment**

- Jacobian revision: `be54c78d6591f92710caa335dc309c8b6ac9b851`
- Python: 3.12.13
- OS: Linux x86_64

**Related**

- #1936 — earlier projective-line dimension admission bug; the current report concerns cardinality/work even with the now-correct two-coordinate axis.
- #2129 — public-operation contract preflight.
- #2135 and #2136 — separate derived-value authenticity failures.

### LINKED ISSUE #2141: [Bug]: Projective chart transition is not invariant under homogeneous rescaling
**Priority**

P0 (critical mathematical correctness)

**Area**

Projective coordinates

**Component**

`projective.chart_transition.compute`

**Observed behavior**

The operation is not a well-defined function on projective points: two homogeneous tuples representing the same point can produce different transition coordinates.

At revision `be54c78d6591f92710caa335dc309c8b6ac9b851`:

```json
{
  "point": {"coordinates": [
    {"num":"1","den":"1"},
    {"num":"2","den":"1"},
    {"num":"3","den":"1"}
  ]},
  "chart_i": 0,
  "chart_j": 1
}
```

returns:

```json
{"transition":[{"num":"3","den":"2"}],"chart_i":0,"chart_j":1}
```

But the equivalent homogeneous representative `[2:4:6]` returns:

```json
{"transition":[{"num":"3","den":"1"}],"chart_i":0,"chart_j":1}
```

Since `[1:2:3]=[2:4:6]` in projective space, the outputs must agree.

**Root cause**

The implementation forms each retained coordinate as:

```text
X_k * X_i / X_j
```

for `k` different from both `i` and `j`. The extra factor `X_i` makes the result scale by the homogeneous representative. Under `X -> lambda X`, it changes by a factor of `lambda`.

The result also omits the target-chart coordinate corresponding to the old source normalization index `i`. A chart of `P^n` has `n` affine coordinates; transitioning from `U_i` to `U_j` should return every ratio `X_k/X_j` for `k != j`, including `X_i/X_j`. The current implementation returns only `n-1` coordinates by excluding both indices.

For the example above, the target `U_1` coordinates, ordered by homogeneous indices with index 1 omitted, are:

```text
(X_0/X_1, X_2/X_1) = (1/2, 3/2)
```

for both representatives.

**Why this violates the accepted design**

#1892 defines the exact standard-chart transition contract:

- every target ratio is divided by the target normalization coordinate;
- equivalent projective rescalings normalize identically;
- `U_i -> U_j -> U_i` is the identity on the overlap; and
- source/target affine axes are explicit.

The published implementation violates all four requirements.

**Expected behavior**

Represent the projective point or source chart point canonically and return the complete target chart point:

```text
(X_k / X_j) for every k != j
```

with the target affine axis and overlap condition `X_j != 0`. If the point is outside `U_j`, return a typed `OUTSIDE_TARGET_CHART` result rather than dividing or throwing.

**Suggested success criteria**

- `[1:2:3]` and every nonzero rational rescaling yield the same target chart point.
- The example returns `(1/2,3/2)` under the documented target-axis order.
- Output dimension for `P^n` is exactly `n`.
- `U_i -> U_j -> U_i` round-trips on the overlap.
- A point with `X_j=0` returns a closed outside-chart result.
- Coordinate-axis permutations transport points, chart indices, and outputs coherently.
- Tests cover `P^1`, `P^2`, negative/rational rescalings, every chart pair, and denominator-zero boundaries.

**Environment**

- Jacobian revision: `be54c78d6591f92710caa335dc309c8b6ac9b851`
- Python: 3.12.13
- OS: Linux x86_64

**Related**

- #1892 — exact projective points and standard-chart transition contract.
- #1129 — bind positional data to typed axes.

### LINKED ISSUE #2145: [Bug]: universal_algebra terms admit cyclic ASTs and equation profiles lack work bounds
**Priority**

P1 (high)

**Area**

Math domain / bounded request contracts

**Component**

- `universal_algebra.term.evaluate.compute`
- `universal_algebra.equation.profile.compute`

**Observed behavior**

`FlatTerm` validates only that its root index is in range. It does not validate node kinds, child-index bounds, acyclicity, operation arity against the bound algebra, variable bounds, node count, or depth.

A term whose application node references itself therefore passes the public request model and reaches recursive evaluation, which ends with:

```text
RecursionError: maximum recursion depth exceeded
```

Separately, `EquationProfileRequest.variable_count` has no upper or derived work bound. The Boolean-algebra catalog example still validates after changing `variable_count` to `251`; execution would enumerate:

```text
2 ** 251
```

assignments before returning an exact profile.

**Steps to reproduce**

At revision `be54c78d6591f92710caa335dc309c8b6ac9b851`, start from the published term-evaluation example and make the application node reference itself:

```python
from copy import deepcopy

from jacobian.catalog.catalog import Catalog
from jacobian.dispatch import invoke_operation

catalog = Catalog.open()
op = catalog.operation("universal_algebra.term.evaluate.compute")
payload = deepcopy(op.examples[0].input)
payload["term"]["nodes"][2]["children"][0] = 2

invoke_operation(op.operation_id, payload, catalog)
# RecursionError: maximum recursion depth exceeded
```

The unbounded equation request is accepted without executing the expansion:

```python
op = catalog.operation("universal_algebra.equation.profile.compute")
payload = deepcopy(op.examples[0].input)
payload["variable_count"] = 251
op.request_type.model_validate(payload)  # accepted
```

The kernel enumerates `product(range(n), repeat=variable_count)`.

**Expected behavior**

A public term value should be a bounded, source-bound, acyclic AST before evaluation begins. Request validation should establish:

- closed node kinds;
- every child index is in range and the reachable graph is acyclic;
- application nodes reference a declared operation with matching arity;
- variable nodes are well formed and fit the assignment/declared variable domain; and
- public node-count and depth limits.

`equation.profile.compute` should preflight its exact assignment count from carrier size and variable count and reject requests above a documented work/output envelope.

**Additional context**

This is separate from #2132, which owns incomplete partitions and a `variable_count` smaller than referenced variables. This issue concerns cyclic/invalid term structure and missing upper work bounds.

The original acceptance criteria in #1912 explicitly required bounded term nodes, term depth, and assignment count.

Source:

- https://github.com/morluto/jacobian/blob/be54c78d6591f92710caa335dc309c8b6ac9b851/src/jacobian/math/universal_algebra/values.py#L64-L91
- https://github.com/morluto/jacobian/blob/be54c78d6591f92710caa335dc309c8b6ac9b851/src/jacobian/math/universal_algebra/operations.py#L18-L88

**Environment**

- Jacobian revision: `be54c78d6591f92710caa335dc309c8b6ac9b851`
- Python: 3.12.13
- OS: Linux x86_64

### LINKED ISSUE #2146: [Bug]: universal_algebra.quotient.compute does not return a composable FiniteAlgebra
**Priority**

P2 (medium)

**Area**

Math domain / result contracts

**Component**

`universal_algebra.quotient.compute`

**Observed behavior**

A successful quotient computation returns the three pieces of an algebra, but its `operations` field uses positional pairs:

```json
{
  "carrier": ["B0"],
  "operations": [["and", 2], ["or", 2]],
  "tables": [[0], [0]]
}
```

The domain's canonical `FiniteAlgebra` type instead requires typed operation symbols:

```json
{
  "operations": [
    {"operation_id": "and", "arity": 2},
    {"operation_id": "or", "arity": 2}
  ]
}
```

As a result, the exact output of `universal_algebra.quotient.compute` cannot be supplied as the `algebra` input to term evaluation, equation profiling, subalgebra generation, congruence checking, or another quotient without caller-side reconstruction.

**Steps to reproduce**

```python
from jacobian.catalog.catalog import Catalog
from jacobian.dispatch import invoke_operation

catalog = Catalog.open()
op = catalog.operation("universal_algebra.quotient.compute")
result = invoke_operation(op.operation_id, op.examples[0].input, catalog)

finite_algebra_type = op.request_type.model_fields["algebra"].annotation
finite_algebra_type.model_validate(result.output)
```

Observed:

```text
ValidationError: operations.0 Input should be a valid dictionary or instance of OperationSymbol
ValidationError: operations.1 Input should be a valid dictionary or instance of OperationSymbol
```

**Expected behavior**

A quotient operation should return the domain's canonical typed `FiniteAlgebra` value, together with the quotient map if that map is part of the advertised postcondition. Its output should round-trip directly into every operation accepting a `FiniteAlgebra`.

A separate display-oriented tuple rendering may be derived metadata, but it should not replace the executable mathematical result.

**Additional context**

This is a result-composability defect, distinct from #2132's invalid quotient requests. The original #1912 contract required quotient operations to return an exact finite algebra and canonical quotient map.

Source:

- https://github.com/morluto/jacobian/blob/be54c78d6591f92710caa335dc309c8b6ac9b851/src/jacobian/math/universal_algebra/_models.py#L126-L131
- https://github.com/morluto/jacobian/blob/be54c78d6591f92710caa335dc309c8b6ac9b851/src/jacobian/math/universal_algebra/values.py#L23-L61

**Environment**

- Jacobian revision: `be54c78d6591f92710caa335dc309c8b6ac9b851`
- Python: 3.12.13
- OS: Linux x86_64

### LINKED ISSUE #2147: [Bug]: Interpolation treats equal rational nodes as distinct and returns zoo coefficients
**Priority**

P1 (high)

**Area**

Polynomial interpolation / exact rational contracts

**Components**

- `polynomial.interpolation.divided_differences.compute`
- `polynomial.interpolation.newton_form.compute`
- `polynomial.interpolation.newton_evaluate.compute`

**Observed behavior**

Interpolation nodes are represented as rational strings, but the request validator checks pairwise distinctness using string equality rather than exact rational equality.

Consequently, the mathematically equal nodes `"1"` and `"1/1"` are accepted as distinct. Divided differences then divide by zero, and `newton_form.compute` returns a typed result containing SymPy's indeterminate value `zoo`:

```json
{
  "coefficients": ["1", "zoo", "zoo"],
  "nodes": ["1", "1/1", "2"],
  "method": "NEWTON_FORM"
}
```

This is not merely a malformed-string exception: both spellings are valid exact rationals, but they denote the same interpolation node.

**Steps to reproduce**

```python
from jacobian.catalog.catalog import Catalog
from jacobian.dispatch import invoke_operation

catalog = Catalog.open()

invoke_operation(
    "polynomial.interpolation.newton_form.compute",
    {
        "nodes": ["1", "1/1", "2"],
        "values": ["1", "2", "5"],
    },
    catalog,
).output
```

Observed coefficients are `['1', 'zoo', 'zoo']`.

**Expected behavior**

Requests should normalize nodes to Jacobian's canonical exact rational value before testing mathematical invariants. Pairwise distinctness must be checked on exact rational values, not source spelling.

Every accepted request should return finite canonical rational coefficients/results or a declared typed mathematical outcome. `zoo`, `nan`, and infinities must not cross this exact-rational result contract.

**Suggested success criteria**

- `"1"` and `"1/1"` normalize to the same rational and are rejected as repeated nodes.
- Equivalent sign and reduction variants, such as `"-1/-2"` and `"1/2"`, receive the same treatment.
- Results use the canonical rational representation and cannot contain `zoo`, `nan`, or infinities.
- Distinct canonical rational nodes continue to produce exact finite coefficients.
- Tests cover equivalent spellings and points immediately at the public count/digit bounds.

**Additional context**

The current `_require_distinct` implementation stores raw strings in a set, while the kernel parses them only afterward with `sympy.Rational`.

- https://github.com/morluto/jacobian/blob/be54c78d6591f92710caa335dc309c8b6ac9b851/src/jacobian/math/polynomial_interpolation_ops/_models.py#L14-L63
- https://github.com/morluto/jacobian/blob/be54c78d6591f92710caa335dc309c8b6ac9b851/src/jacobian/math/polynomial_interpolation_ops/_operations.py#L17-L70

#2154 separately owns invalid rational syntax and zero-denominator strings. This issue owns equality after valid exact-rational parsing and the invalid typed `zoo` result.

Related: #2129 and #1043.

**Environment**

- Jacobian revision: `be54c78d6591f92710caa335dc309c8b6ac9b851`
- Python: 3.12.13
- SymPy: 1.14.0
- OS: Linux x86_64

### LINKED ISSUE #2148: [Contract]: Make Newton-form results directly evaluable
## Current limitation

`polynomial.interpolation.newton_form.compute` produces an exact Newton presentation:

```text
nodes + divided-difference coefficients
```

but `polynomial.interpolation.newton_evaluate.compute` does not accept that value. Its request instead takes:

```text
nodes + sample values + evaluation point
```

and recomputes the complete divided-difference table before evaluating.

This leaves the producer and nominal consumer non-composable. Despite the title “Evaluate a polynomial in Newton form,” the evaluator consumes interpolation samples rather than a Newton form.

## Reproduction

The published form operation returns fields:

```json
{
  "nodes": ["0", "1", "2"],
  "coefficients": ["1", "1", "1"]
}
```

The evaluator requires:

```json
{
  "nodes": ["0", "1", "2"],
  "values": ["1", "2", "5"],
  "evaluation_point": "3"
}
```

Passing the form result directly fails schema validation because `values` and `evaluation_point` are missing and `coefficients` is not accepted. Renaming `coefficients` to `values` is mathematically wrong because the evaluator treats those values as samples and interpolates them again.

## Desired outcome

Introduce or reuse one canonical typed Newton-form value containing exact nodes and Newton coefficients. The form producer should return it, and the Newton evaluator should accept that exact value plus an evaluation point and perform only nested multiplication.

If evaluation from raw samples remains useful, it should retain a name and contract that clearly says it interpolates and evaluates from samples rather than consuming Newton coefficients.

## Success criteria

- Output from `newton_form.compute` round-trips directly into the Newton-form evaluator with only an evaluation point added.
- Evaluation does not recompute divided differences from coefficient data.
- Exact-rational types and bounds agree across producer and consumer.
- A composition test proves `samples -> Newton form -> evaluation` agrees with direct polynomial evaluation.
- Discovery descriptions distinguish “evaluate supplied Newton form” from “interpolate and evaluate samples.”

## Context

This is separate from the invalid-rational/repeated-node bug. It concerns the stable mathematical value exposed between two otherwise related operations.

Source:

- https://github.com/morluto/jacobian/blob/be54c78d6591f92710caa335dc309c8b6ac9b851/src/jacobian/math/polynomial_interpolation_ops/_models.py#L37-L81
- https://github.com/morluto/jacobian/blob/be54c78d6591f92710caa335dc309c8b6ac9b851/src/jacobian/math/polynomial_interpolation_ops/_operations.py#L50-L71

Related: #1883 and #2129.

### LINKED ISSUE #2149: [Bug]: tree_decomposition.restrict accepts subset vertices not in the source graph
## Observed behavior

`graph.tree_decomposition.restrict.compute` accepts a `subset` containing vertices that are not declared in the source graph. The request passes validation, but the backend raises `ValueError: subset must contain only declared source vertices`.

## Reproduction

```python
from jacobian.catalog.catalog import Catalog
from jacobian.dispatch import invoke_operation

catalog = Catalog.open()

payload = {
    "decomposition": {
        "graph": {
            "graph_schema_version": "1",
            "vertices": ["a", "b", "c"],
            "edges": [["a", "b"], ["b", "c"]],
        },
        "tree_nodes": ["t0", "t1"],
        "tree_edges": [["t0", "t1"]],
        "bags": [["a", "b"], ["b", "c"]],
    },
    "subset": ["x"],  # "x" not in graph vertices
}

result = invoke_operation(
    "graph.tree_decomposition.restrict.compute",
    payload,
    catalog,
)
# ValueError: subset must contain only declared source vertices
```

## Expected behavior

The request model should validate that every element of `subset` is a declared vertex in `decomposition.graph.vertices`. An accepted request should either return a typed result or be rejected during validation.

## Environment

- Jacobian revision: be54c78d6591f92710caa335dc309c8b6ac9b851
- Python: 3.12.13
- OS: Linux

### LINKED ISSUE #2151: [Bug]: Galois-group results omit the permutation generators required for composition
**Priority**

P2 (medium)

**Area**

Galois theory / cross-domain result contracts

**Component**

`polynomial.galois_group.compute`

**Observed behavior**

The operation computes a concrete SymPy permutation group but discards its generators at the public boundary. The result contains only:

```text
group_name: string rendering of the backend object
order
degree
is_solvable
```

For example, `x^3 - 2` returns:

```json
{
  "group_name": "PermutationGroup([\n    (0 1 2),\n    (2)(0 1)])",
  "order": 6,
  "degree": 3,
  "is_solvable": true,
  "method": "SYMPY_GALOIS_GROUP"
}
```

The string is backend display text, not an executable or stable mathematical group value. It cannot be passed to Jacobian's finite-group operations, whose shared request requires `degree` plus explicit generator permutations.

**Steps to reproduce**

```python
from jacobian.catalog.catalog import Catalog
from jacobian.dispatch import invoke_operation

catalog = Catalog.open()
result = invoke_operation(
    "polynomial.galois_group.compute",
    {"coefficients": [-2, 0, 0, 1]},  # x^3 - 2, ascending order
    catalog,
)
print(result.output)
```

There is no field from which to construct the required `generators` input for:

- `group.order.compute`
- `group.orbit.compute`

without reparsing an undocumented SymPy representation or recomputing the group outside Jacobian.

**Expected behavior**

A successful exact Galois-group result should expose the root action as Jacobian-owned permutation generators on a stable degree-sized root axis. Backend names or database labels may remain derived metadata, but they should not replace the mathematical value.

At minimum, output from `polynomial.galois_group.compute` should round-trip directly into the existing permutation-group consumers after selecting the embedded group value.

**Suggested success criteria**

- The result contains bounded explicit generator permutations with a declared root-axis convention.
- Every generator is validated as a permutation of `0..degree-1`.
- Recomputed group order and solvability agree with the reported invariants.
- Relabelling the root axis conjugates the group coherently rather than changing its mathematical identity.
- A composition test sends the produced group directly to `group.order.compute` and `group.orbit.compute`.
- No consumer parses `group_name` or another backend rendering.

**Additional context**

This is distinct from #2126, which owns reducible and unsupported-degree request admission. It is a successful-result representation and composability problem.

The original #1862 acceptance criteria explicitly required public groups to be root-action permutation groups rather than names alone.

Source:

- https://github.com/morluto/jacobian/blob/be54c78d6591f92710caa335dc309c8b6ac9b851/src/jacobian/math/galois_theory/_models.py#L91-L96
- https://github.com/morluto/jacobian/blob/be54c78d6591f92710caa335dc309c8b6ac9b851/src/jacobian/math/galois_theory/_operations.py#L74-L86
- https://github.com/morluto/jacobian/blob/be54c78d6591f92710caa335dc309c8b6ac9b851/src/jacobian/math/group/_models.py#L15-L30

**Environment**

- Jacobian revision: `be54c78d6591f92710caa335dc309c8b6ac9b851`
- Python: 3.12.13
- SymPy: 1.14.0
- OS: Linux x86_64

### LINKED ISSUE #2152: [Bug]: finite_field.linear_map.rank malformed matrices escape request validation as TypeError
**Priority**

P2 (medium)

**Area**

Math request boundary / finite fields

**Component**

`finite_field.linear_map.rank.compute`

**Observed behavior**

The operation's nested matrix is a plain `PrimeFieldMatrix` dataclass decoded by the shared `StrictModel` before-validator. Missing required fields are passed directly to the dataclass constructor, so malformed public JSON raises raw Python `TypeError` rather than a structured request-validation error.

Removing any of these fields reproduces the problem:

```text
linear_map.matrix.prime
linear_map.matrix.entries
linear_map.matrix.columns
```

Observed errors include:

```text
TypeError: PrimeFieldMatrix.__init__() missing 1 required positional argument: 'prime'
TypeError: PrimeFieldMatrix.__init__() missing 1 required positional argument: 'entries'
TypeError: PrimeFieldMatrix.__init__() missing 1 required positional argument: 'columns'
```

**Steps to reproduce**

```python
from copy import deepcopy

from jacobian.catalog.catalog import Catalog
from jacobian.dispatch import invoke_operation

catalog = Catalog.open()
op = catalog.operation("finite_field.linear_map.rank.compute")
payload = deepcopy(op.examples[0].input)
del payload["linear_map"]["matrix"]["prime"]

invoke_operation(op.operation_id, payload, catalog)
```

The exception occurs while parsing the request, before the domain operation runs.

**Expected behavior**

Malformed nested JSON must produce ordinary field-addressable Pydantic validation diagnostics at the public request boundary. Dataclass construction errors should never escape as raw `TypeError`.

A maintained Pydantic model for the public matrix value is preferable when it carries the complete mathematical contract. If the dataclass adapter remains, it must validate required fields and nested shapes without invoking a constructor that can throw outside Pydantic's error model.

**Suggested success criteria**

- Missing `prime`, `entries`, and `columns` each produce a `ValidationError` bound to `linear_map.matrix.<field>`.
- Malformed rows, wrong JSON types, extra fields, and inconsistent dimensions also fail through the same request-validation channel.
- The valid published example remains unchanged mathematically.
- A generic regression test covers every public request field backed by a non-Pydantic dataclass.

**Additional context**

The shared adapter currently calls `cls(**prepared)` inside a `mode="before"` model validator:

- https://github.com/morluto/jacobian/blob/be54c78d6591f92710caa335dc309c8b6ac9b851/src/jacobian/_models.py#L64-L100

At the audited catalog revision, this is the only public operation whose request graph exposes such a dataclass field. This is separate from #2123's loss of details after a normal Pydantic `ValidationError`.

**Environment**

- Jacobian revision: `be54c78d6591f92710caa335dc309c8b6ac9b851`
- Python: 3.12.13
- OS: Linux x86_64

### LINKED ISSUE #2153: [Bug]: Doob martingale accepts invalid payoff rationals and fails during execution
**Priority**

P2 (medium)

**Area**

Probability / public request contracts

**Component**

`probability.process.doob_martingale.compute`

**Observed behavior**

`DoobMartingaleRequest` checks only payoff length and observation lengths. It does not validate that each payoff entry is an exact rational.

A payoff containing `"not-a-rational"` therefore passes the public request model. During execution, the kernel constructs a `FiniteRandomVariable`, whose validator finally parses the value and raises a nested Pydantic `ValidationError`:

```text
Value error, Invalid literal for Fraction: 'not-a-rational'
```

The accepted request never returns the declared `DoobMartingaleResult`.

**Steps to reproduce**

```python
from copy import deepcopy

from jacobian.catalog.catalog import Catalog
from jacobian.dispatch import invoke_operation

catalog = Catalog.open()
op = catalog.operation("probability.process.doob_martingale.compute")
payload = deepcopy(op.examples[0].input)
payload["payoff"][0] = "not-a-rational"

invoke_operation(op.operation_id, payload, catalog)
```

**Expected behavior**

The request model should establish the complete payoff domain before computing the filtration. Every payoff entry should be a canonical exact rational value under the same representation used by `FiniteRandomVariable`.

Invalid, zero-denominator, noncanonical, and over-budget values should fail request validation with the precise payoff index. Every accepted request should return a typed martingale result.

**Suggested success criteria**

- Invalid payoff text is rejected at `payoff.<index>` before filtration construction.
- Payoff and returned martingale values use the repository's canonical rational representation.
- Equivalent rational spellings normalize consistently rather than remaining distinct wire values.
- Tests cover invalid text, zero denominators, reducible fractions, sign normalization, and digit bounds.

**Additional context**

The request model omits the rational check performed later by `FiniteRandomVariable`:

- https://github.com/morluto/jacobian/blob/be54c78d6591f92710caa335dc309c8b6ac9b851/src/jacobian/math/finite_stochastic_processes/_models.py#L72-L86
- https://github.com/morluto/jacobian/blob/be54c78d6591f92710caa335dc309c8b6ac9b851/src/jacobian/math/finite_stochastic_processes/operations.py#L114-L131

Related: #1043 and #2129.

**Environment**

- Jacobian revision: `be54c78d6591f92710caa335dc309c8b6ac9b851`
- Python: 3.12.13
- OS: Linux x86_64

### LINKED ISSUE #2154: [Bug]: polynomial interpolation operations accept non-rational strings as nodes and values
## Observed behavior

Three polynomial interpolation operations accept non-rational strings in `nodes`, `values`, and `evaluation_point` fields. The request models validate only that nodes are pairwise distinct and that lengths match — they do not validate that the string values are valid rational numbers. The operations then try to parse the strings as `Fraction` objects and raise untyped `TypeError` or `ZeroDivisionError` exceptions.

The affected operations:
- `polynomial.interpolation.divided_differences.compute`
- `polynomial.interpolation.newton_evaluate.compute`
- `polynomial.interpolation.newton_form.compute`

## Reproduction

```python
from jacobian.catalog.catalog import Catalog
from jacobian.dispatch import invoke_operation

catalog = Catalog.open()

# Non-numeric string in nodes
result = invoke_operation(
    "polynomial.interpolation.divided_differences.compute",
    {"nodes": ["x", "y", "z"], "values": ["1", "2", "3"]},
    catalog,
)
# TypeError: invalid input: x

# 0/0 as evaluation point
result = invoke_operation(
    "polynomial.interpolation.newton_evaluate.compute",
    {
        "nodes": ["0", "1", "2"],
        "values": ["1", "2", "5"],
        "evaluation_point": "0/0",
    },
    catalog,
)
# ZeroDivisionError: Fraction(1, 0)
```

The request models in `src/jacobian/math/polynomial_interpolation_ops/_models.py` accept any string for `nodes`, `values`, and `evaluation_point`. They only check that nodes are pairwise distinct and lengths match. No pattern validation or rational-number parsing is applied.

## Expected behavior

The request model should validate that `nodes`, `values`, and `evaluation_point` are valid rational numbers (or at least valid integers) before backend execution. An accepted request should return a typed result rather than leaking a `TypeError` or `ZeroDivisionError`.

## Environment

- Jacobian revision: be54c78d6591f92710caa335dc309c8b6ac9b851
- Python: 3.12.13
- OS: Linux

### LINKED ISSUE #2156: [Bug]: commutative algebra and polynomial vector calculus operations accept non-polynomial expression strings
## Observed behavior

The commutative algebra and polynomial vector calculus operations accept a wide range of non-polynomial strings in their `polynomial` or `generators` fields. SymPy's `sympify` parses these into non-polynomial objects (booleans, lazy functions, or fails with a `SympifyError`), and the operations leak the resulting untyped exceptions instead of rejecting the request during validation.

## Affected operations

**`polynomial.ideal.radical_membership.decide`** (`src/jacobian/math/commutative_algebra_ops/_operations.py`):
- `"true"` → `TypeError: BooleanAtom not allowed in this context`
- `"false"` → `TypeError: BooleanAtom not allowed in this context`
- `"test"` → `TypeError: unsupported operand type(s) for *: 'Symbol' and 'LazyFunction'`
- `"   "` → `SympifyError: could not parse '   '`

**`polynomial_field.scalar.gradient.compute`**, **`polynomial_field.scalar.laplacian.compute`**, **`polynomial_field.scalar.directional_derivative.compute`** (`src/jacobian/math/polynomial_vector_calc/_operations.py`):
- `"true"` → `AttributeError: 'BooleanTrue' object has no attribute 'is_polynomial'`
- `"false"` → `AttributeError: 'BooleanFalse' object has no attribute 'is_polynomial'`
- `"test"` → `AttributeError: 'LazyFunction' object has no attribute 'is_polynomial'`
- `"   "` → `ValueError: failed to parse expression:    `
- `"null"` → `ValueError: expression contains undeclared symbols: ['null']`

## Reproduction

```python
from jacobian.catalog.catalog import Catalog
from jacobian.dispatch import invoke_operation

catalog = Catalog.open()

# radical_membership with "true"
result = invoke_operation(
    "polynomial.ideal.radical_membership.decide",
    {"variables": ["x"], "generators": ["x**2"], "polynomial": "true"},
    catalog,
)
# TypeError: BooleanAtom not allowed in this context.

# gradient with "test"
result = invoke_operation(
    "polynomial_field.scalar.gradient.compute",
    {"variables": ["x", "y"], "polynomial": "test"},
    catalog,
)
# AttributeError: 'LazyFunction' object has no attribute 'is_polynomial'
```

## Expected behavior

The request model should validate that the parsed expression is a polynomial with rational coefficients over the declared variables before backend execution. The `plane_algebraic_curves` operations (`src/jacobian/math/plane_algebraic_curves/_models.py`) already do this correctly — they use `sympy.Poly(..., domain=sympy.QQ)` and reject booleans, lazy functions, and other non-polynomial SymPy objects. The `commutative_algebra_ops` and `polynomial_vector_calc` operations should match this pattern.

## Related

- #2125 owns the security dimension of `sympify` evaluating Python code
- #2127 owns the architectural direction of making polynomial inputs typed values
- #2131 owns the specific `0/0`/`nan` acceptance gap

This issue is about the broader class of non-polynomial strings (booleans, lazy functions, whitespace, reserved words) that pass `sympify` but are not polynomials, and raise untyped backend exceptions.

## Environment

- Jacobian revision: be54c78d6591f92710caa335dc309c8b6ac9b851
- Python: 3.12.13
- SymPy: 1.14.0
- OS: Linux

### LINKED ISSUE #2157: [Bug]: Doob martingale accepts 0/0 and 1/0 in payoff field
## Observed behavior

`probability.process.doob_martingale.compute` accepts the indeterminate expression `"0/0"` and the undefined expression `"1/0"` in the `payoff` field. The request passes validation, but the backend raises `ZeroDivisionError: Fraction(0, 0)` or `ZeroDivisionError: Fraction(1, 0)`.

## Reproduction

```python
from jacobian.catalog.catalog import Catalog
from jacobian.dispatch import invoke_operation

catalog = Catalog.open()

result = invoke_operation(
    "probability.process.doob_martingale.compute",
    {
        "space": {"samples": ["H", "T"], "masses": ["1/2", "1/2"]},
        "observations": [["heads", "tails"]],
        "payoff": ["0/0", "0"],
    },
    catalog,
)
# ZeroDivisionError: Fraction(0, 0)
```

## Expected behavior

The request model should validate that `payoff` values are valid rational numbers before backend execution. An accepted request should return a typed result rather than leaking a `ZeroDivisionError`.

## Environment

- Jacobian revision: be54c78d6591f92710caa335dc309c8b6ac9b851
- Python: 3.12.13
- OS: Linux

### LINKED ISSUE #2158: [Bug]: universal_algebra operations accept arbitrary strings for node kind
## Observed behavior

`universal_algebra.equation.profile.compute` and `universal_algebra.term.evaluate.compute` accept arbitrary strings for the `kind` field of term tree nodes. The request models use a `str` type for `kind` without restricting it to the valid values `"variable"` and `"application"`. The backend then raises `ValueError: unknown node kind: <value>`.

## Affected operations

- `universal_algebra.equation.profile.compute`
- `universal_algebra.term.evaluate.compute`

## Reproduction

```python
from jacobian.catalog.catalog import Catalog
from jacobian.dispatch import invoke_operation

catalog = Catalog.open()

# Any invalid kind string is accepted
payload = {
    "algebra": {
        "carrier": ["0", "1"],
        "operations": [
            {"operation_id": "and", "arity": 2},
            {"operation_id": "or", "arity": 2},
        ],
        "tables": [[0, 0, 0, 1], [0, 1, 1, 1]],
    },
    "term": {
        "nodes": [
            {"kind": "test", "variable_id": 0, "operation": None, "children": []},
        ],
        "root": 0,
    },
    "assignment": {"values": ["0"]},
    "variable_count": 1,
}

result = invoke_operation(
    "universal_algebra.term.evaluate.compute",
    payload,
    catalog,
)
# ValueError: unknown node kind: test
```

## Expected behavior

The `kind` field should be a `Literal["variable", "application"]` (or equivalent discriminated union) so that invalid values are rejected during request validation. An accepted request should return a typed result rather than raising a backend `ValueError`.

## Related

- #2132 covers the cross-field validation gaps in these same operations (variable_count / incomplete assignment / empty partition)

## Environment

- Jacobian revision: be54c78d6591f92710caa335dc309c8b6ac9b851
- Python: 3.12.13
- OS: Linux

### LINKED ISSUE #2159: [Bug]: symbolic matrix and commutative algebra operations accept empty string variable names
## Observed behavior

Several operations accept empty string `""` as a variable name in their `variables` field. SymPy's `sympify` raises `ValueError: cannot use "" in local_dict`, which leaks as an untyped backend exception.

## Affected operations

- `matrix.symbolic.determinant.compute`
- `matrix.symbolic.rank.compute`
- `matrix.symbolic.characteristic_polynomial.compute`
- `matrix.symbolic.eigenvalues.compute`
- `polynomial.ideal.radical.compute`
- `polynomial.ideal.radical_membership.decide`
- `polynomial.ideal.quotient.compute`

## Reproduction

```python
from jacobian.catalog.catalog import Catalog
from jacobian.dispatch import invoke_operation

catalog = Catalog.open()

result = invoke_operation(
    "polynomial.ideal.radical.compute",
    {"variables": ["", "y"], "generators": ["x**2", "x*y"]},
    catalog,
)
# ValueError: cannot use "" in local_dict
```

## Expected behavior

Variable names should be validated as non-empty, valid Python identifiers during request validation. The `plane_algebraic_curves` operations (`src/jacobian/math/plane_algebraic_curves/_models.py`) already implement this check via `_require_valid_variables()`. The affected operations should match this pattern.

## Environment

- Jacobian revision: be54c78d6591f92710caa335dc309c8b6ac9b851
- Python: 3.12.13
- OS: Linux

morluto:
## Problem

Public mathematical operations accepted loosely related strings, axes, parents,
and derived structures that did not establish the domains required by their
implementations. Several results also discarded units, multiplicities,
generators, quotient maps, or other data needed for reconstruction and
composition.

Advances #2174.

Closes #2122.
Closes #2125.
Closes #2126.
Closes #2127.
Closes #2128.
Closes #2131.
Closes #2133.
Closes #2134.
Closes #2135.
Closes #2136.
Closes #2137.
Closes #2141.
Closes #2145.
Closes #2146.
Closes #2147.
Closes #2148.
Closes #2149.
Closes #2151.
Closes #2152.
Closes #2153.
Closes #2154.
Closes #2156.
Closes #2157.
Closes #2158.
Closes #2159.

## Solution

Replace expression-shaped polynomial and rational inputs with canonical exact
values carrying their ordered domains. Enforce parent, axis, dimension,
degeneracy, and work constraints in operation request models before backend
execution. Return composable certificates for finite-field factorizations,
Galois groups, Newton forms, quotient algebras, projective coordinates, and
authoritative finite-field derivations.

The established mathematical kernels remain delegated to SymPy and FLINT. This
PR changes Jacobian's public contracts and conversions around them.

## Testing

- `make check` — 1,986 tests passed; Ruff, formatting, complexity, and mypy passed.
- Specialist validation run (if any): none for this branch.

## Public contract impact

Breaking pre-stable request and result schema changes across affected operations.
Canonical polynomial, rational-function, finite-field, universal-algebra, and
derived certificate values replace expression strings and incomplete result
shapes. Operation versions and invocation examples are updated with their
contracts; no new operation IDs are introduced.

## Public operation admission

- Concrete gap: existing operations accepted values outside their mathematical or backend domains and returned values that were not reliably composable.
- Why existing operations or typed values are insufficient: the old string and field-bag representations did not carry complete parent, axis, or reconstruction information.
- Stable mathematical result: exact canonical values and complete bounded certificates for the existing operations.
- Admission decision: retain the existing owner-local `KEEP` decisions with corrected pre-stable contracts.

## Closure matrix

| --- | --- | --- |
| Canonical exact inputs | delivered | Existing affected operation IDs |
| Complete Galois certificates | delivered | `polynomial.galois.*` |
| Authentic finite-field derivations | delivered | `finite_field.*` |
| Closed universal-algebra terms and quotients | delivered | `universal_algebra.*` |

## Checklist

- [x] `make check` passes
- [x] Harbor task or verifier changes are not applicable
- [x] Public operation changes retain owner-local admission decisions
- [x] Result semantics distinguish exact and unavailable outcomes where applicable
- [x] Shared canonical values serve multiple production domains

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -225,13 +225,10 @@ source_modules = ["jacobian.dispatch"]
 forbidden_modules = ["jacobian.cli", "jacobian.mcp"]
 
 [[tool.importlinter.contracts]]
-name = "Finite-field values do not import private backends"
+name = "Finite-field values do not import the optional FLINT backend"
 type = "forbidden"
 source_modules = ["jacobian.math.finite_fields.values"]
-forbidden_modules = [
-    "jacobian.math.finite_fields._flint",
-    "jacobian.math.finite_fields._sympy",
-]
+forbidden_modules = ["jacobian.math.finite_fields._flint"]
 
 [tool.coverage.run]
 branch = true
--- a/src/jacobian/_models.py
+++ b/src/jacobian/_models.py
@@ -1,106 +1,15 @@
 """Narrow strict-model primitive shared by unrelated wire owners."""
 
-from __future__ import annotations
-
-from dataclasses import fields as dataclass_fields
-from dataclasses import is_dataclass
-from types import UnionType
-from typing import Annotated, Any, Union, get_args, get_origin
-
-from pydantic import BaseModel, ConfigDict, model_validator
-
-
-def _unwrap_annotation(annotation: Any) -> Any:
-    origin = get_origin(annotation)
-    if origin is Annotated:
-        args = get_args(annotation)
-        return _unwrap_annotation(args[0]) if args else annotation
-    return annotation
-
-
-def _tuple_annotation(annotation: Any) -> Any | None:
-    annotation = _unwrap_annotation(annotation)
-    origin = get_origin(annotation)
-    if origin is tuple:
-        return annotation
-    if origin in {Union, UnionType}:
-        tuple_args = [
-            arg
-            for arg in get_args(annotation)
-            if arg is not type(None) and get_origin(_unwrap_annotation(arg)) is tuple
-        ]
-        if len(tuple_args) == 1:
-            return tuple_args[0]
-    return None
-
-
-def _lists_to_tuples(value: Any) -> Any:
-    if isinstance(value, list):
-        return tuple(_lists_to_tuples(item) for item in value)
-    return value
-
-
-def _dataclass_annotation(annotation: Any) -> type[Any] | None:
-    annotation = _unwrap_annotation(annotation)
-    origin = get_origin(annotation)
-    if origin in {Union, UnionType}:
-        dataclass_args = [
-            arg
-            for arg in get_args(annotation)
-            if arg is not type(None) and _dataclass_annotation(arg) is not None
-        ]
-        if len(dataclass_args) == 1:
-            return _dataclass_annotation(dataclass_args[0])
-        return None
-    if (
-        isinstance(annotation, type)
-        and is_dataclass(annotation)
-        and not issubclass(annotation, BaseModel)
-    ):
-        return annotation
-    return None
-
-
-def _dataclass_from_json(cls: type[Any], value: Any) -> Any:
-    if isinstance(value, cls):
-        return value
-    if not isinstance(value, dict):
-        return value
-    allowed = {field.name for field in dataclass_fields(cls)}
-    unexpected = sorted(set(value) - allowed)
-    if unexpected:
-        raise ValueError(f"unexpected dataclass fields: {', '.join(unexpected)}")
-    prepared = {
-        name: _lists_to_tuples(item) if isinstance(item, list) else item
-        for name, item in value.items()
-    }
-    return cls(**prepared)
+from pydantic import BaseModel, ConfigDict
 
 
 class StrictModel(BaseModel):
-    """Closed, immutable model with JSON tuple/dataclass decoding."""
+    """Closed immutable model; Pydantic owns nested and JSON validation."""
 
-    model_config = ConfigDict(extra="forbid", frozen=True)
+    model_config = ConfigDict(
+        extra="forbid",
+        frozen=True,
+    )
 
-    @model_validator(mode="before")
-    @classmethod
-    def accept_json_wire_shapes(cls, data: Any) -> Any:
-        """Decode JSON arrays and objects into declared tuple and dataclass fields."""
 
-        if isinstance(data, cls) or not isinstance(data, dict):
-            return data
-        coerced = dict(data)
-        for name, field in cls.model_fields.items():
-            if name not in coerced:
-                continue
-            value = coerced[name]
-            dataclass_type = _dataclass_annotation(field.annotation)
-            if dataclass_type is not None:
-                coerced[name] = _dataclass_from_json(dataclass_type, value)
-                continue
-            if _tuple_annotation(field.annotation) is None or not isinstance(
-                value, list
-            ):
-                continue
-            coerced[name] = _lists_to_tuples(value)
-        return coerced
+__all__ = ["StrictModel"]
--- a/src/jacobian/math/arithmetic_dynamics/_models.py
+++ b/src/jacobian/math/arithmetic_dynamics/_models.py
@@ -8,6 +8,7 @@ from typing import Literal, Self
 
 from pydantic import Field, model_validator
 
+from jacobian._exact import CanonicalRational, require_bounded_rational
 from jacobian._models import StrictModel
 from jacobian.math._rational_height import RationalHeight, sum_heights
 
@@ -146,26 +147,19 @@ def _mobius(value: int) -> int:
     return -1 if factors % 2 else 1
 
 
-def parse_canonical_rational(value: str, *, label: str) -> Fraction:
-    if len(value) > 2 * MAX_COEFFICIENT_DIGITS + 2:
-        raise ValueError(f"{label} exceeds the rational digit bound")
-    try:
-        parsed = Fraction(value)
-    except (ValueError, ZeroDivisionError):
-        raise ValueError(f"{label} must be a canonical rational") from None
-    if str(parsed) != value:
-        raise ValueError(f"{label} must be a reduced canonical rational")
-    if (
-        len(str(abs(parsed.numerator))) > MAX_COEFFICIENT_DIGITS
-        or len(str(parsed.denominator)) > MAX_COEFFICIENT_DIGITS
-    ):
-        raise ValueError(f"{label} exceeds the rational digit bound")
-    return parsed
+def _bounded_fraction(
+    value: CanonicalRational, *, max_digits: int, label: str
+) -> Fraction:
+    require_bounded_rational(value, max_digits=max_digits, label=label)
+    return value.as_fraction()
 
 
-def parse_polynomial_coefficients(values: tuple[str, ...]) -> tuple[Fraction, ...]:
+def parse_polynomial_coefficients(
+    values: tuple[CanonicalRational, ...],
+) -> tuple[Fraction, ...]:
     coefficients = tuple(
-        parse_canonical_rational(value, label="coefficient") for value in values
+        _bounded_fraction(value, max_digits=MAX_COEFFICIENT_DIGITS, label="coefficient")
+        for value in values
     )
     if len(coefficients) > 1 and coefficients[-1] == 0:
         raise ValueError("polynomial coefficients must omit trailing zeros")
@@ -173,7 +167,9 @@ def parse_polynomial_coefficients(values: tuple[str, ...]) -> tuple[Fraction, ..
 
 
 class PolynomialCoefficientRequest(StrictModel):
-    coefficients: tuple[str, ...] = Field(min_length=1, max_length=MAX_DEGREE + 1)
+    coefficients: tuple[CanonicalRational, ...] = Field(
+        min_length=1, max_length=MAX_DEGREE + 1
+    )
 
     @model_validator(mode="after")
     def require_canonical_coefficients(self) -> Self:
@@ -207,12 +203,12 @@ class MapIterateRequest(PolynomialCoefficientRequest):
 class OrbitPrefixRequest(PolynomialCoefficientRequest):
     """Compute until a first repeat or an explicit finite/output bound."""
 
-    start: str
+    start: CanonicalRational
     max_steps: int = Field(ge=0, le=MAX_ORBIT_STEPS)
 
     @model_validator(mode="after")
     def require_canonical_start(self) -> Self:
-        parse_canonical_rational(self.start, label="start")
+        _bounded_fraction(self.start, max_digits=MAX_COEFFICIENT_DIGITS, label="start")
         return self
 
 
@@ -253,12 +249,17 @@ class DynatomicPolynomialRequest(PolynomialCoefficientRequest):
 class CycleMultiplierRequest(PolynomialCoefficientRequest):
     """Compute the multiplier of a supplied, validated exact rational cycle."""
 
-    cycle: tuple[str, ...] = Field(min_length=1, max_length=MAX_ORBIT_STEPS)
+    cycle: tuple[CanonicalRational, ...] = Field(
+        min_length=1, max_length=MAX_ORBIT_STEPS
+    )
 
     @model_validator(mode="after")
     def require_exact_cycle(self) -> Self:
         points = tuple(
-            parse_canonical_rational(value, label="cycle point") for value in self.cycle
+            _bounded_fraction(
+                value, max_digits=MAX_COEFFICIENT_DIGITS, label="cycle point"
+            )
+            for value in self.cycle
         )
         if len(set(points)) != len(points):
             raise ValueError("cycle points must be distinct")
@@ -287,11 +288,11 @@ class FiniteFieldMapRequest(StrictModel):
 
 
 class MapIterateResult(StrictModel):
-    source_coefficients: tuple[str, ...] = Field(
+    source_coefficients: tuple[CanonicalRational, ...] = Field(
         min_length=1, max_length=MAX_DEGREE + 1
     )
     n: int = Field(ge=0, le=MAX_ITERATE)
-    coefficients: tuple[str, ...] = Field(
+    coefficients: tuple[CanonicalRational, ...] = Field(
         min_length=1, max_length=MAX_ITERATE_DEGREE + 1
     )
     degree: int = Field(ge=0, le=MAX_ITERATE_DEGREE)
@@ -301,13 +302,14 @@ class MapIterateResult(StrictModel):
     @model_validator(mode="after")
     def bind_degree_and_coefficients(self) -> Self:
         parse_polynomial_coefficients(self.source_coefficients)
-        values = tuple(Fraction(value) for value in self.coefficients)
-        if any(
-            str(Fraction(value)) != value
-            or len(value) > MAX_POLYNOMIAL_OUTPUT_DIGITS * 2 + 2
+        values = tuple(
+            _bounded_fraction(
+                value,
+                max_digits=MAX_POLYNOMIAL_OUTPUT_DIGITS,
+                label="iterate coefficient",
+            )
             for value in self.coefficients
-        ):
-            raise ValueError("iterate coefficients must be bounded and canonical")
+        )
         expected_degree = 0 if values == (Fraction(0),) else len(values) - 1
         if self.degree != expected_degree:
             raise ValueError("degree must match the canonical coefficient tuple")
@@ -330,11 +332,13 @@ class OrbitRepeatEvidence(StrictModel):
 
 
 class OrbitPrefixResult(StrictModel):
-    source_coefficients: tuple[str, ...] = Field(
+    source_coefficients: tuple[CanonicalRational, ...] = Field(
         min_length=1, max_length=MAX_DEGREE + 1
     )
-    start: str
-    orbit: tuple[str, ...] = Field(min_length=1, max_length=MAX_ORBIT_STEPS + 1)
+    start: CanonicalRational
+    orbit: tuple[CanonicalRational, ...] = Field(
+        min_length=1, max_length=MAX_ORBIT_STEPS + 1
+    )
     requested_steps: int = Field(ge=0, le=MAX_ORBIT_STEPS)
     computed_steps: int = Field(ge=0, le=MAX_ORBIT_STEPS)
     termination: Literal["REPEAT_FOUND", "STEP_BOUND_REACHED", "OUTPUT_BOUND_REACHED"]
@@ -374,10 +378,10 @@ class OrbitPrefixResult(StrictModel):
 
 
 class DynatomicPolynomialResult(StrictModel):
-    source_coefficients: tuple[str, ...] = Field(
+    source_coefficients: tuple[CanonicalRational, ...] = Field(
         min_length=1, max_length=MAX_DEGREE + 1
     )
-    coefficients: tuple[str, ...] = Field(
+    coefficients: tuple[CanonicalRational, ...] = Field(
         min_length=1, max_length=MAX_DYNATOMIC_DEGREE + 1
     )
     degree: int = Field(ge=0, le=MAX_DYNATOMIC_DEGREE)
@@ -390,13 +394,14 @@ class DynatomicPolynomialResult(StrictModel):
     @model_validator(mode="after")
     def bind_degree_and_coefficients(self) -> Self:
         parse_polynomial_coefficients(self.source_coefficients)
-        values = tuple(Fraction(value) for value in self.coefficients)
-        if any(
-            str(Fraction(value)) != value
-            or len(value) > MAX_POLYNOMIAL_OUTPUT_DIGITS * 2 + 2
+        values = tuple(
+            _bounded_fraction(
+                value,
+                max_digits=MAX_POLYNOMIAL_OUTPUT_DIGITS,
+                label="dynatomic coefficient",
+            )
             for value in self.coefficients
-        ):
-            raise ValueError("dynatomic coefficients must be bounded and canonical")
+        )
         expected_degree = 0 if values == (Fraction(0),) else len(values) - 1
         if self.degree != expected_degree:
             raise ValueError("degree must match the canonical coefficient tuple")
@@ -404,11 +409,13 @@ class DynatomicPolynomialResult(StrictModel):
 
 
 class CycleMultiplierResult(StrictModel):
-    source_coefficients: tuple[str, ...] = Field(
+    source_coefficients: tuple[CanonicalRational, ...] = Field(
         min_length=1, max_length=MAX_DEGREE + 1
     )
-    multiplier: str
-    cycle: tuple[str, ...] = Field(min_length=1, max_length=MAX_ORBIT_STEPS)
+    multiplier: CanonicalRational
+    cycle: tuple[CanonicalRational, ...] = Field(
+        min_length=1, max_length=MAX_ORBIT_STEPS
+    )
     period: int = Field(ge=1, le=MAX_ORBIT_STEPS)
     validated_cycle: Literal[True] = True
     complete: Literal[True] = True
@@ -416,9 +423,7 @@ class CycleMultiplierResult(StrictModel):
     @model_validator(mode="after")
     def bind_period_and_multiplier(self) -> Self:
         source = parse_polynomial_coefficients(self.source_coefficients)
-        points = tuple(
-            parse_canonical_rational(value, label="cycle point") for value in self.cycle
-        )
+        points = tuple(value.as_fraction() for value in self.cycle)
         if len(set(points)) != len(points) or any(
             _evaluate(source, point) != points[(index + 1) % len(points)]
             for index, point in enumerate(points)
@@ -426,13 +431,11 @@ class CycleMultiplierResult(StrictModel):
             raise ValueError("cycle must be a distinct ordered cycle of the bound map")
         if self.period != len(self.cycle):
             raise ValueError("period must match cycle length")
-        value = Fraction(self.multiplier)
-        if (
-            str(value) != self.multiplier
-            or len(str(abs(value.numerator))) > MAX_POLYNOMIAL_OUTPUT_DIGITS
-            or len(str(value.denominator)) > MAX_POLYNOMIAL_OUTPUT_DIGITS
-        ):
-            raise ValueError("multiplier must be a bounded canonical rational")
+        require_bounded_rational(
+            self.multiplier,
+            max_digits=MAX_POLYNOMIAL_OUTPUT_DIGITS,
+            label="multiplier",
+        )
         return self
 
 
@@ -514,14 +517,19 @@ def _evaluate(coefficients: tuple[Fraction, ...], point: Fraction) -> Fraction:
 
 
 def _require_bound_orbit(
-    source_coefficients: tuple[str, ...],
-    start: str,
-    orbit_values: tuple[str, ...],
+    source_coefficients: tuple[CanonicalRational, ...],
+    start: CanonicalRational,
+    orbit_values: tuple[CanonicalRational, ...],
 ) -> None:
     source = parse_polynomial_coefficients(source_coefficients)
-    initial = parse_canonical_rational(start, label="start")
+    initial = _bounded_fraction(start, max_digits=MAX_COEFFICIENT_DIGITS, label="start")
     orbit = tuple(
-        parse_canonical_rational(value, label="orbit value") for value in orbit_values
+        _bounded_fraction(
+            value,
+            max_digits=MAX_ORBIT_VALUE_DIGITS,
+            label="orbit value",
+        )
+        for value in orbit_values
     )
     if orbit[0] != initial:
         raise ValueError("orbit must begin at the bound start point")
--- a/src/jacobian/math/arithmetic_dynamics/_operations.py
+++ b/src/jacobian/math/arithmetic_dynamics/_operations.py
@@ -2,9 +2,9 @@
 
 from __future__ import annotations
 
-from fractions import Fraction
 from typing import Any
 
+from jacobian._exact import CanonicalRational
 from jacobian.math.arithmetic_dynamics._models import (
     CycleMultiplierRequest,
     CycleMultiplierResult,
@@ -34,11 +34,11 @@ def _polynomial(request: PolynomialCoefficientRequest) -> Any:
     return polynomial_from_coefficients(request.coefficient_values())
 
 
-def _format_coefficients(polynomial: Any) -> tuple[str, ...]:
-    values = tuple(str(value) for value in polynomial_coefficients(polynomial))
-    if any(len(value) > 65_538 for value in values):
-        raise ValueError("polynomial coefficient output exceeds digit bound")
-    return values
+def _format_coefficients(polynomial: Any) -> tuple[CanonicalRational, ...]:
+    return tuple(
+        CanonicalRational.from_fraction(value)
+        for value in polynomial_coefficients(polynomial)
+    )
 
 
 def compute_map_iterate(request: MapIterateRequest) -> MapIterateResult:
@@ -54,7 +54,7 @@ def compute_map_iterate(request: MapIterateRequest) -> MapIterateResult:
 def compute_orbit_prefix(request: OrbitPrefixRequest) -> OrbitPrefixResult:
     result = orbit_prefix(
         _polynomial(request),
-        Fraction(request.start),
+        request.start.as_fraction(),
         request.max_steps,
     )
     repeat = (
@@ -71,7 +71,7 @@ def compute_orbit_prefix(request: OrbitPrefixRequest) -> OrbitPrefixResult:
     return OrbitPrefixResult(
         source_coefficients=request.coefficients,
         start=request.start,
-        orbit=tuple(str(value) for value in result.orbit),
+        orbit=tuple(CanonicalRational.from_fraction(value) for value in result.orbit),
         requested_steps=request.max_steps,
         computed_steps=len(result.orbit) - 1,
         termination=result.termination,
@@ -96,10 +96,12 @@ def compute_dynatomic_polynomial(
 def compute_cycle_multiplier(
     request: CycleMultiplierRequest,
 ) -> CycleMultiplierResult:
-    points = tuple(Fraction(value) for value in request.cycle)
+    points = tuple(value.as_fraction() for value in request.cycle)
     return CycleMultiplierResult(
         source_coefficients=request.coefficients,
-        multiplier=str(cycle_multiplier(_polynomial(request), points)),
+        multiplier=CanonicalRational.from_fraction(
+            cycle_multiplier(_polynomial(request), points)
+        ),
         cycle=request.cycle,
         period=len(points),
     )
--- a/src/jacobian/math/arithmetic_dynamics/_tools.py
+++ b/src/jacobian/math/arithmetic_dynamics/_tools.py
@@ -70,9 +70,17 @@ ARITHMETIC_DYNAMICS_OPERATIONS: tuple[MathTool[Any, Any], ...] = (
             example(
                 "f_x_squared_plus_1_iterate_2",
                 "Compute f^2 for f(x)=x^2+1; n must be non-negative.",
-                {"coefficients": ["1", "0", "1"], "n": 2},
+                {
+                    "coefficients": [
+                        {"num": "1", "den": "1"},
+                        {"num": "0", "den": "1"},
+                        {"num": "1", "den": "1"},
+                    ],
+                    "n": 2,
+                },
             ),
         ),
+        version="2",
     ),
     _op(
         "arithmetic_dynamics.point.orbit.compute",
@@ -92,9 +100,18 @@ ARITHMETIC_DYNAMICS_OPERATIONS: tuple[MathTool[Any, Any], ...] = (
                 "orbit_of_0_under_x2",
                 "Orbit of 0 under f(x)=x^2 for 5 steps; "
                 "start must be a rational number.",
-                {"coefficients": ["0", "0", "1"], "start": "0", "max_steps": 5},
+                {
+                    "coefficients": [
+                        {"num": "0", "den": "1"},
+                        {"num": "0", "den": "1"},
+                        {"num": "1", "den": "1"},
+                    ],
+                    "start": {"num": "0", "den": "1"},
+                    "max_steps": 5,
+                },
             ),
         ),
+        version="2",
     ),
     _op(
         "arithmetic_dynamics.dynatomic_polynomial.compute",
@@ -112,9 +129,17 @@ ARITHMETIC_DYNAMICS_OPERATIONS: tuple[MathTool[Any, Any], ...] = (
             example(
                 "dynatomic_n1_x2",
                 "Dynatomic polynomial for n=1 of f(x)=x^2; n must be at least 1.",
-                {"coefficients": ["0", "0", "1"], "n": 1},
+                {
+                    "coefficients": [
+                        {"num": "0", "den": "1"},
+                        {"num": "0", "den": "1"},
+                        {"num": "1", "den": "1"},
+                    ],
+                    "n": 1,
+                },
             ),
         ),
+        version="2",
     ),
     _op(
         "arithmetic_dynamics.cycle.multiplier.compute",
@@ -134,9 +159,17 @@ ARITHMETIC_DYNAMICS_OPERATIONS: tuple[MathTool[Any, Any], ...] = (
                 "multiplier_fixed_0_x2",
                 "Multiplier of the fixed point 0 under f(x)=x^2; "
                 "cycle points must be rational.",
-                {"coefficients": ["0", "0", "1"], "cycle": ["0"]},
+                {
+                    "coefficients": [
+                        {"num": "0", "den": "1"},
+                        {"num": "0", "den": "1"},
+                        {"num": "1", "den": "1"},
+                    ],
+                    "cycle": [{"num": "0", "den": "1"}],
+                },
             ),
         ),
+        version="2",
     ),
     _op(
         "arithmetic_dynamics.finite_field.functional_graph.compute",
--- a/src/jacobian/math/finite_fields/_models.py
+++ b/src/jacobian/math/finite_fields/_models.py
@@ -16,11 +16,26 @@ from jacobian.math.finite_fields import (
     ProjectiveLine,
     ProjectivePoint,
 )
+from jacobian.math.finite_fields.values import _direction_rank_work
+
+_MAX_PROJECTIVE_POINTS = 4_096
+_MAX_FINITE_MAP_WORK = 1_000_000
+_MAX_DIRECTION_RANK_WORK = 1_000_000
 
 
 class RestrictScalarsRequest(StrictModel):
     subspace: FiniteDimensionalSubspace
     direction: ProjectivePoint
+
+    @model_validator(mode="after")
+    def require_same_parent_and_axis(self) -> Self:
+        if self.direction.presentation != self.subspace.presentation:
+            raise ValueError("direction and subspace must share their presentation")
+        if self.direction.axis != self.subspace.row_axis:
+            raise ValueError("direction axis must match the subspace row axis")
+        if _direction_rank_work(self.subspace, 1) > _MAX_DIRECTION_RANK_WORK:
+            raise ValueError("restriction exceeds the operation work budget")
+        return self
 
 
 class LinearMapRankRequest(StrictModel):
@@ -46,12 +61,28 @@ class ProjectiveLineRequest(StrictModel):
             raise ValueError(
                 "projective-line enumeration requires a two-coordinate axis"
             )
+        count = self.presentation.order + 1
+        if count > _MAX_PROJECTIVE_POINTS:
+            raise ValueError("projective line exceeds the output-size budget")
         return self
 
 
 class DirectionRankLedgerRequest(StrictModel):
     subspace: FiniteDimensionalSubspace
     directions: ProjectiveLine
+
+    @model_validator(mode="after")
+    def require_same_parent_axis_and_work(self) -> Self:
+        if self.directions.presentation != self.subspace.presentation:
+            raise ValueError("directions and subspace must share their presentation")
+        if self.directions.axis != self.subspace.row_axis:
+            raise ValueError("direction axis must match the subspace row axis")
+        if (
+            _direction_rank_work(self.subspace, len(self.directions.points))
+            > _MAX_DIRECTION_RANK_WORK
+        ):
+            raise ValueError("direction-rank ledger exceeds the operation work budget")
+        return self
 
 
 class OrbitDistributionRequest(StrictModel):
@@ -60,6 +91,17 @@ class OrbitDistributionRequest(StrictModel):
 
 class FiniteMapTableRequest(StrictModel):
     polynomial_map: FinitePolynomialMap
+
+    @model_validator(mode="after")
+    def require_bounded_enumeration(self) -> Self:
+        work = (
+            self.polynomial_map.domain.order
+            * len(self.polynomial_map.polynomial.coefficients)
+            * self.polynomial_map.domain.degree
+        )
+        if work > _MAX_FINITE_MAP_WORK:
+            raise ValueError("finite map exceeds the operation work budget")
+        return self
 
 
 class FiberPartitionRequest(StrictModel):
--- a/src/jacobian/math/finite_fields/_operations.py
+++ b/src/jacobian/math/finite_fields/_operations.py
@@ -6,7 +6,6 @@ from jacobian.math.finite_fields import (
     CollisionResult,
     DirectionRankLedger,
     FiberPartition,
-    FiniteDimensionalSubspace,
     FiniteLinearMap,
     FiniteMapTable,
     OrbitDistribution,
@@ -34,12 +33,6 @@ from jacobian.math.finite_fields._models import (
     ProjectiveLineRequest,
     RestrictScalarsRequest,
 )
-
-_MAX_PROJECTIVE_POINTS = 4096
-_MAX_FINITE_MAP_ELEMENTS = 4096
-_MAX_FINITE_MAP_WORK = 1_000_000
-_MAX_FINITE_MAP_REPLAY_WORK = 1_000_000
-_MAX_DIRECTION_RANK_WORK = 1_000_000
 
 _FIELD: dict[str, object] = {
     "characteristic": 2,
@@ -131,13 +124,6 @@ _TABLE: dict[str, object] = {
 
 
 def _enumerate_projective_line(request: ProjectiveLineRequest) -> ProjectiveLine:
-    count = (request.presentation.order ** len(request.axis.labels) - 1) // (
-        request.presentation.order - 1
-    )
-    if count > _MAX_PROJECTIVE_POINTS:
-        raise ValueError(
-            f"projective line has {count} directions; limit is {_MAX_PROJECTIVE_POINTS}"
-        )
     return projective_line(request.presentation, request.axis)
 
 
@@ -150,92 +136,33 @@ def _rank(request: LinearMapRankRequest) -> RankResult:
 
 
 def _ledger(request: DirectionRankLedgerRequest) -> DirectionRankLedger:
-    _require_direction_rank_work(request.subspace, len(request.directions.points))
     return direction_rank_ledger(request.subspace, request.directions)
 
 
-def _direction_rank_work(
-    subspace: FiniteDimensionalSubspace,
-    direction_count: int,
-) -> int:
-    source_dimension = len(subspace.basis)
-    target_dimension = len(subspace.column_axis.labels) * subspace.presentation.degree
-    restriction_work = (
-        source_dimension * len(subspace.row_axis.labels) * target_dimension
-    )
-    rank_work = (
-        target_dimension * source_dimension * min(target_dimension, source_dimension)
-    )
-    return direction_count * (restriction_work + rank_work)
-
-
-def _require_direction_rank_work(
-    subspace: FiniteDimensionalSubspace,
-    direction_count: int,
-) -> None:
-    work = _direction_rank_work(subspace, direction_count)
-    if work > _MAX_DIRECTION_RANK_WORK:
-        raise ValueError(
-            f"direction-rank work is {work}; limit is {_MAX_DIRECTION_RANK_WORK}",
-        )
-
-
 def _orbit_distribution(request: OrbitDistributionRequest) -> OrbitDistribution:
-    _require_direction_rank_work(request.ledger.subspace, len(request.ledger.entries))
     return orbit_distribution(request.ledger)
 
 
 def _finite_map_table(request: FiniteMapTableRequest) -> FiniteMapTable:
-    polynomial_map = request.polynomial_map
-    count = polynomial_map.domain.order
-    if count > _MAX_FINITE_MAP_ELEMENTS:
-        raise ValueError(
-            f"finite map has {count} inputs; limit is {_MAX_FINITE_MAP_ELEMENTS}"
-        )
-    work = (
-        count
-        * len(polynomial_map.polynomial.coefficients)
-        * polynomial_map.domain.degree
-    )
-    if work > _MAX_FINITE_MAP_WORK:
-        raise ValueError(f"finite map work is {work}; limit is {_MAX_FINITE_MAP_WORK}")
     return finite_map_table(request.polynomial_map)
 
 
 def _fiber_partition(request: FiberPartitionRequest) -> FiberPartition:
-    _require_finite_map_replay_work(request)
     return fiber_partition(request.table)
 
 
 def _analyze_collisions(request: CollisionRequest) -> CollisionResult:
-    _require_finite_map_replay_work(request)
     return analyze_collisions(request.table)
 
 
 def _analyze_permutation(request: PermutationRequest) -> PermutationResult:
-    _require_finite_map_replay_work(request)
     return analyze_permutation(request.table)
-
-
-def _require_finite_map_replay_work(
-    request: FiberPartitionRequest | CollisionRequest | PermutationRequest,
-) -> None:
-    table = request.table
-    work = (
-        len(table.entries)
-        * len(table.map.polynomial.coefficients)
-        * table.map.domain.degree
-    )
-    if work > _MAX_FINITE_MAP_REPLAY_WORK:
-        raise ValueError(
-            f"finite map replay work is {work}; limit is {_MAX_FINITE_MAP_REPLAY_WORK}",
-        )
 
 
 def finite_field_operations() -> MathTools:
     projective_line_operation = MathTool(
         operation_id="finite_field.projective_line.enumerate",
-        version="1",
+        version="2",
         request_type=ProjectiveLineRequest,
         result_type=ProjectiveLine,
         run=_enumerate_projective_line,
@@ -252,7 +179,7 @@ def finite_field_operations() -> MathTools:
     )
     restrict_operation = MathTool(
         operation_id="finite_field.restrict_scalars.compute",
-        version="1",
+        version="2",
         request_type=RestrictScalarsRequest,
         result_type=FiniteLinearMap,
         run=_restrict,
@@ -269,7 +196,7 @@ def finite_field_operations() -> MathTools:
     )
     rank_operation = MathTool(
         operation_id="finite_field.linear_map.rank.compute",
-        version="1",
+        version="2",
         request_type=LinearMapRankRequest,
         result_type=RankResult,
         run=_rank,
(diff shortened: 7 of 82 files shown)


morluto:
@codex

PR #2219, opened 2026-08-21
Add pinned Mathlib declaration search

N0zoM1z0:
## Problem

Agents using Jacobian could elaborate a Lean snippet only after they already knew the relevant declaration. Jacobian had no bounded local way to search the fixed Mathlib environment by declaration name, type constants, namespace, or declaration kind, so premise retrieval depended on model memory, filesystem text search, or an external service.

The existing `lean.check` boundary also invoked the core Lean executable outside the repository's Lake environment. A candidate returned from Mathlib lookup could therefore fail verification with `unknown module prefix 'Mathlib'`, even though the compiled Mathlib environment was installed locally.

This left three related gaps:

- no typed Mathlib declaration-search operation behind `math.find` / `math.run`;
- no single pinned environment shared by lookup and elaboration;
- no reusable evaluation that separates formal answer correctness from whether an agent discovers and selects the new operation.

### Add a bounded declaration-search operation

This PR admits `lean.declarations.search` as a new immutable logic operation. Its request supports:

- one literal, case-sensitive declaration-name substring;
- up to four exact type-constant names;
- up to eight namespace prefixes;
- declaration-kind filters;
- a result limit from 1 to 20;
- a timeout from 1 to 60 seconds.

Type constants and namespaces use a conservative ASCII dotted-name grammar. They are serialized as JSON data and compared against Lean names; caller input is never inserted into Lean source or passed to an evaluator.

The result returns the fixed Mathlib revision, the original typed query, scanned-declaration count, `EXHAUSTED` versus `RESULT_LIMIT`, and bounded declaration records containing:

- exact declaration name;
- kind;
- defining module;
- an 8,000-character pretty-printed type preview;
- an explicit `type_truncated` flag.

Matching is performed against the complete Lean expression, not the displayed preview. Timeout, unavailable runtime, process-limit failure, and malformed backend output remain typed execution outcomes and do not become negative mathematical conclusions.

The private Lean helper imports the complete pinned Mathlib environment, sorts the immutable declaration set, filters private names, and performs one request-scoped scan. The Python adapter validates the complete `lake-manifest.json` SHA-256, Lean toolchain, Mathlib commit, and `Mathlib.olean` before invoking it with fixed process, output, memory, and time bounds.

### Use the same pinned environment for `lean.check`

`lean.check` now executes through `lake env lean` in the validated Mathlib workspace. Self-contained snippets may therefore `import Mathlib` and directly verify a declaration returned by search. Its default timeout moves from 10 to the existing 30-second maximum because a cold umbrella import was observed at approximately 10 seconds; the public maximum is unchanged.

Both operations remain stateless. Source and query files live only in request-scoped temporary directories and are deleted when each subprocess exits.

### Pin and provision Mathlib

The repository now pins:

- Lean `v4.31.0`;
- Mathlib `v4.31.0` / commit `fabf563a7c95a166b8d7b6efca11c8b4dc9d911f`;
- the complete Lake manifest by SHA-256.

`make setup-lean`, the Lean CI action, and the runtime Docker image install the fixed toolchain and compiled Mathlib cache. The Dockerfile creates the Mathlib layer before copying Python source, so ordinary source edits do not rebuild the multi-gigabyte cache.

The resulting local image is 11,244,185,392 bytes. The Lean processes use a 10,240 MB memory threshold, so the backend documentation calls out the corresponding deployment cost explicitly.

### Add a reusable Harbor adoption regression

The new `mathlib-gcd-premise-retrieval` task asks an agent to submit one fully qualified declaration and its ordered local arguments for a frozen Lean goal. The hidden verifier:

- validates the strict submission schema;
- binds evaluation to the frozen input;
- accepts only conservative declaration/identifier grammars;
- constructs only `exact <declaration> <arguments>`;
- independently elaborates that source with pinned Lean;
- fails closed on missing runtime, timeout, invalid schema, injection text, or elaboration failure.

Host regressions cover the Oracle witness, a wrong declaration, a missing argument, executable text, and an extra submission field. The task intentionally uses a core declaration imported by the pinned Mathlib environment so the independent verifier remains small; the Lean process lane separately requires and verifies `sq_abs` from `Mathlib.Algebra.Order.Ring.Abs`.

The evaluation guide includes matched uncued and explicitly labeled routing-cue arms, with three attempts per arm. Formal verifier reward and MCP trajectory telemetry are reported separately: calling a tool is not itself success.

### Keep evaluation image selection machine-readable

While running the Harbor treatment, the dirty-tree image builder exposed a pre-existing integration bug: Docker diagnostics were printed to stdout, so Make command substitution passed build logs as `JACOBIAN_IMAGE` and Compose rejected the value. Build diagnostics now go to stderr and stdout contains only the selected image reference. A focused regression preserves this contract.

## Testing

Automated validation on commit `71dae2f36bedd5b74b61a8cb92b68b6f854b0e58`:

```sh
make check
# 2284 passed, 37 skipped

make check-external
# 2 passed; includes import Mathlib + exact sq_abs a and a Mathlib.* module lookup

make docs-linkcheck

make harbor-prepare-task \
  DATASET=mathematical-benchmarks-v1 \
  TASKS=mathlib-gcd-premise-retrieval
# no generated or formatting changes

make harbor-validate-task \
  DATASET=mathematical-benchmarks-v1 \
  TASKS=mathlib-gcd-premise-retrieval
# 5 host verifier tests passed
# Oracle: 1 trial, 0 exceptions, reward 1.0
# digest: sha256:f2bc2aa54837f755082540fcabc571c0ea7d2fc2fcf770f3bcb71c37759f9d4c

make container-image IMAGE=jacobian:local
# image built successfully
```

The rebuilt image was also smoke-tested directly: `lean.declarations.search` returned `sq_abs` from `Mathlib.Algebra.Order.Ring.Abs`, and `lean.check` elaborated the matching `import Mathlib` proof with no diagnostics.

Manual Codex CLI observations used `gpt-5.6-terra` with high reasoning effort, a fresh local MCP name, apps disabled, and the public `math.find` / `math.run` interface:

- true Mathlib case: search returned `sq_abs` in 15,233 ms; `lean.check` returned `ELABORATED` with no diagnostics in 9,986 ms;
- additional declaration-search cases found `Nat.add_comm`, refined a broad reverse query to `List.reverse_reverse`, and returned an empty `COMPLETED` / `EXHAUSTED` result for a missing name without confusing it with timeout or error;
- Harbor uncued arm: the model made no Jacobian call, guessed an obsolete declaration name, and received reward 0;
- a weak generic lookup cue still produced no MCP call and reward 0;
- explicit routing-cue arm: the model used `math.find`, inspected `lean.declarations.search`, executed it, selected `Nat.gcd_zero_right`, checked the proof, and received formal reward 1.0.

The paired result supports a narrow conclusion: availability, discovery after routing, execution, and formal use work; uncued tool selection remains a presentation/salience risk and is intentionally retained as a regression signal.

No benchmark snapshot or model output is committed.

## Public contract impact

- Adds operation ID `lean.declarations.search`, version `1`, with new typed request and result schemas.
- Materially changes `lean.check`'s fixed environment to include the pinned Mathlib Lake workspace.
- Changes `lean.check.timeout_seconds` default from 10 to 30 seconds; its maximum remains 30 seconds.
- Keeps the MCP surface at exactly two tools: `math.find` and `math.run`.
- Adds a required multi-gigabyte Mathlib runtime dependency for the service image and Lean specialist lane.
- No compatibility shim is added; this repository is pre-stable, and an absent or changed environment returns a typed `UNAVAILABLE` outcome.

## Public operation admission

- Concrete gap: agents cannot retrieve a declaration from the fixed local Mathlib environment when they do not already know its exact name.
- Why existing operations or typed values are insufficient: `math.find` discovers Jacobian operations, not Lean declarations, while `lean.check` can only validate a source snippet chosen by the caller.
- Stable mathematical result: a bounded query-bound projection of exact declaration names, kinds, defining modules, and type previews from one immutable Mathlib environment, with explicit exhaustion and truncation semantics.
- Admission decision: admitted owner-locally in `jacobian.math.logic._admission` as `lean.declarations.search`; the shared admission policy remains fail closed.

## Closure matrix

| --- | --- | --- |
| Bounded fixed-environment Mathlib declaration lookup | delivered | `lean.declarations.search` |

No additional public operation candidates are proposed by this PR.

## Checklist

- [x] `make check` passes
- [x] Explicitly relevant specialist validation is listed above (Lean and Harbor/Oracle)
- [x] Harbor task/verifier changes ran `make harbor-prepare-task` then `make harbor-validate-task`
- [x] Result semantics distinguish completed/exhausted, result-limited, unavailable, timeout, and error outcomes
- [x] New shared abstractions replace duplication in at least two surviving production paths (not applicable; no shared abstraction added)


morluto:
closing this for now since adding lean would increase dependency size

PR #2326, opened 2026-08-23
feat(matrices): add exact rational spectrum claim checks for symmetric matrices

### LINKED ISSUE #2283: [Operation]: Check complete rational spectra of bounded symmetric matrices
## Parent gap / RFC

This is a narrow checker follow-up to #1670 and #1691. The former produces graph spectra; the latter introduced exact rational symmetric-matrix inertia. Neither checks a supplied complete rational spectrum claim while binding every multiplicity to the exact source matrix.

For a bounded symmetric matrix over `QQ` and a claimed list of distinct rational eigenvalues with multiplicities, decide exactly whether that list is the matrix's complete spectrum.

## Proposed operation ID

`matrix.symmetric.rational_spectrum_claim.check`

## Input

- one bounded symmetric rational matrix `A`;
- a nonempty list of pairwise-distinct rational values `lambda_i`;
- one positive claimed multiplicity `m_i` for each value.

The matrix and claims must use canonical Jacobian rational values. Reject nonsymmetric matrices, duplicate claimed values, nonpositive multiplicities, and requests outside the declared dimension/coefficient/work budget before invoking the backend.

## Result

Return a source-bound exact check result containing:

- the submitted matrix and claimed profile;
- for each `lambda_i`, the exact nullity of `A - lambda_i I`;
- whether each nullity equals the claimed multiplicity;
- the sum of established multiplicities;
- `valid_complete_rational_spectrum`;
- the first failed condition when invalid.

For a real symmetric matrix, algebraic and geometric multiplicities agree. Therefore the claim is complete exactly when every claimed multiplicity equals the corresponding nullity and the multiplicities sum to the matrix order.

An optional additional ledger may replay that the product of the distinct linear factors annihilates `A`, but the nullity and dimension conditions are already the defining proof.

## Completeness semantics

Every accepted request returns `VALID` or `INVALID` exactly. Backend timeout, allocation failure, or an exceeded intermediate/output budget is an operational non-conclusion and must not be returned as `INVALID`.

This operation is deliberately limited to symmetric rational matrices whose entire claimed spectrum is rational. It does not check nonsymmetric Jordan multiplicities, algebraic-number spectra, or approximate eigenpairs.

## Bounds and cost class

The natural implementation performs one exact shifted rank or inertia computation per claimed eigenvalue. Admission should bound:

- matrix order and nonzero entries;
- number of claimed values;
- input numerator/denominator digits;
- shifted-entry and exact-elimination intermediate digits;
- aggregate matrix work; and
- result bytes.

The existing symmetric-inertia contract admits order 50. The motivating fixture is also `50 x 50`, so that scale is a useful target only if benchmarked; otherwise publish a smaller honest initial bound.

## Maintained implementation

Reuse the exact rational rank/inertia kernel privately. For each claim, compute the nullity of `A - lambda I`, then replay the multiplicity and dimension equalities in domain-owned code.

The result must not accept authored nullities, multiplicities, or a Boolean independently of the source matrix. Rational eigenvalues should remain exact values rather than display strings.

A caller could make several shifted-inertia calls, but the current inertia result does not retain its source matrix or dimension. The caller would still own source binding, distinct-value validation, multiplicity matching, completeness accounting, and the theorem-specific inference from symmetric nullities to the complete spectrum.

This operation packages one stable claim-replay postcondition. It does not search for eigenvalues or replace the general spectrum producer.

## Defining fixtures

The Atlas Graffiti 284 certificate supplies a substantial exact fixture: its `50 x 50` rational symmetric distance matrix has the complete profile

```text
91^1, (-4)^28, 1^21.
```

The certificate independently checks the shifted nullities, an annihilating polynomial, traces, and a nonzero integer eigenvector, and rejects perturbed eigenvalues and vectors:

https://github.com/techno-optimist/erdos-frontier-atlas/tree/main/certificates/graffiti-284-refutation

Smaller defining tests should include:

- diagonal matrices with repeated rational eigenvalues;
- the zero matrix;
- a rational symmetric matrix with an irrational spectrum, for which any claimed rational list is invalid rather than complete;
- a correct subset of eigenvalues whose multiplicities do not sum to the order;
- wrong multiplicity, duplicate value, and source-matrix mutations;
- simultaneous row/column permutation invariance;
- equivalent rational encodings; and
- requests at and immediately over every public budget.

## Deferred operation family

- eigenvalue discovery;
- general algebraic-number spectra;
- nonsymmetric Jordan or generalized-eigenspace checking;
- approximate residual/tolerance checks;
- eigenvector-basis certification; and
- graph-specific spectrum wrappers.

morluto:
Closes #2283.

## Problem

The catalog can compute rational matrix rank and symmetric inertia, but neither operation binds a submitted list of rational eigenvalues and multiplicities to one source matrix or decides whether that list is the complete spectrum. Repeating shifted-rank calls leaves the caller responsible for distinct-value validation, source association, multiplicity matching, and the symmetric-matrix completeness inference.

## Solution

Add `matrix.symmetric.rational_spectrum_claim.check`. The request accepts the canonical materialized `RationalMatrix` plus a nonempty profile of distinct rational eigenvalues and positive multiplicities. It rejects nonsquare or nonsymmetric matrices, duplicate values, and requests outside the digit, work, intermediate-height, or result-size envelope before calling the backend.

For each claim, the operation clears the denominators of every row of `A - λI` and calls python-flint's exact integer-matrix rank. Row scaling preserves rank over `QQ`, so `nullity(A - λI) = n - rank(A - λI)` exactly. Since rational symmetric matrices are diagonalizable over `RR`, the profile is complete exactly when every claimed multiplicity equals its shifted nullity and the claimed multiplicities sum to the matrix order.

The result retains the matrix and submitted profile, reports every exact nullity and comparison, records claimed and established multiplicity sums, and returns a deterministic first failure. Result validation recomputes the complete ledger from the retained source, so authored source, claim, ledger, or conclusion mutations cannot revalidate. Operational failures are not converted into mathematical `INVALID` results.

Suggested review order: request/result invariants in `_models.py`, exact row-cleared rank replay in `_operations.py`, then the adversarial and boundary tests.

## Testing

- `make test-math TESTS=tests/math/matrices/test_rational_spectrum_claim.py` — 11 passed.
- `uv run --locked pytest -q tests/catalog/test_admission.py tests/catalog/test_catalog.py tests/integration/catalog/test_builtin_examples.py tests/integration/catalog/test_public_operation_contract_conformance.py` — 985 passed.
- `make check` — Ruff and formatting clean, mypy clean, 2,660 tests passed.
- Boundary measurement on CPython 3.12.13 with python-flint 0.9.0: one dense symmetric order-32 request with 64-digit rational entries and 32 claims, including source-bound result replay, completed in 1.042184 seconds and returned an exact `INVALID` result. The fixture was generated deterministically with `Random(2283)`.

## Public contract impact

Adds operation ID `matrix.symmetric.rational_spectrum_claim.check` with new request and source-bound result schemas. Existing operation IDs, schemas, and native APIs are unchanged.

## Public operation admission

- **Concrete gap:** no existing operation checks a complete rational spectrum claim while binding all shifted nullities and multiplicities to the canonical source matrix.
- **Why existing operations or typed values are insufficient:** `matrix.rank.compute` and `matrix.inertia.compute` return independent computations; they do not retain the source/profile relation or own completeness replay.
- **Semantic mathematical domain and postcondition:** decide exactly whether distinct rational eigenvalue/multiplicity claims form the complete spectrum of a symmetric matrix over `QQ`.
- **Canonical public value type:** `RationalMatrix`; claims and ledger eigenvalues use `CanonicalRational`.
- **Source representation:** materialized dense canonical rational matrix.
- **Expansion performed by the kernel and its pre-execution bound:** one shifted matrix and one row-cleared integer matrix per claim; order and claim count are each at most 32.
- **Representation-sensitive complexity:** no succinct or generated matrix representation is admitted; the kernel performs no hidden matrix expansion.
- **Admitted request envelope and controlling quantities:** order ≤ 32, nonzero entries ≤ 1,024, claims ≤ 32, source components ≤ 64 digits, shifted components ≤ 129 digits, per-replay shifted-rank work ≤ 1,048,576 `claim_count * order^3` units, and predicted result ≤ 384 KiB. Construction performs two bounded replay passes: computation and source-bound result validation.
- **Producer/consumer closure:** the request consumes `RationalMatrix` unchanged. The output is a source-bound decision, not a new canonical matrix producer.
- **Degenerate inputs:** the nonempty zero matrix is admitted; its zero eigenvalue has nullity equal to the matrix order. Empty matrices are outside the canonical matrix domain.
- **Parent/ring/field identity:** the matrix and claims remain over exact `QQ`; row denominator clearing is an explicit rank-preserving map to an integer matrix used only by the private backend.
- **Deterministic work bound:** each replay performs at most 32 exact rank calls on order-32 matrices; the public result is deterministic in submitted claim order.
- **Maximum intermediate growth:** row clearing gives integer entries with at most `order * shifted_digits` digits; every relevant minor is bounded by `order^2 * shifted_digits + digits(order!) + 1`, capped at 132,256 decimal digits.
- **Exact result-size bound:** one retained matrix, one retained profile, and at most 32 constant-shape ledger entries, conservatively bounded to 384 KiB.
- **Backend and supported version:** pinned python-flint 0.9.0, using `fmpz_mat.rank()`.
- **Backend input domain:** nonempty square integer matrices produced only after complete symmetric-`QQ` request admission and exact row denominator clearing.
- **Conversion/coercion behavior:** canonical rationals become Python `Fraction` values; each shifted row is multiplied by a positive common denominator. No floating-point conversion or implicit field change occurs.
- **Result type:** `RationalSpectrumClaimResult`, a source-bound exact `VALID`/`INVALID` decision with nullity and completeness ledgers.
- **Reconstruction or defining invariant:** every ledger nullity equals `n - rank(A - λI)`; each match flag equals the nullity/multiplicity equality; completeness requires all matches and total multiplicity `n`. Result validation replays these relations.
- **Typed execution failures:** all admitted ordinary executions return the exact result type. Unexpected backend or resource failures remain operational errors and never establish `INVALID`.
- **Property and boundary tests:** repeated and nonintegral rational spectra, zero matrix, irrational spectra, correct incomplete subsets, wrong multiplicities, invalid request classes, independent result-field mutations, simultaneous row/column permutations, and order/claim/digit/nonzero/result boundaries.
- **Remaining fixed caps and classification:** order 32 follows the existing canonical `RationalMatrix` representation and is the measured honest initial envelope. The issue's order-50 source fixture would require a separate evidence-backed widening of that shared canonical value.
- **Admission decision:** `KEEP`.

## Closure matrix

| --- | --- | --- |
| Complete rational spectrum claim check | delivered | `matrix.symmetric.rational_spectrum_claim.check` |
| Algebraic-number, nonsymmetric/Jordan, approximate, discovery, and eigenvector checks | deferred outside this focused operation | none |

## Checklist

- [x] `make check` passes
- [x] Result semantics distinguish mathematical invalidity from operational failure
- [x] New bounds include boundary, permutation, adversarial, and measured scale evidence
- [x] No new shared abstraction was introduced

### DIFF
--- a/src/jacobian/math/matrices/analysis/_admission.py
+++ b/src/jacobian/math/matrices/analysis/_admission.py
@@ -20,6 +20,11 @@ ADMISSIONS: tuple[OperationAdmission, ...] = (
         AdmissionDecision.KEEP,
         "distinct exact bounded mathematical value or invariant with material computational or reliability leverage",
     ),
+    OperationAdmission(
+        "matrix.symmetric.rational_spectrum_claim.check",
+        AdmissionDecision.KEEP,
+        "one source-bound exact predicate that replays every rational eigenspace multiplicity and completeness from a canonical symmetric QQ matrix",
+    ),
 )
 
 REGISTRATION = OperationRegistration(TOOLS, ADMISSIONS)
--- a/src/jacobian/math/matrices/analysis/_models.py
+++ b/src/jacobian/math/matrices/analysis/_models.py
@@ -2,12 +2,216 @@
 
 from __future__ import annotations
 
-from typing import Self
+from math import factorial
+from typing import Literal, Self
 
-from pydantic import Field, model_validator
+from pydantic import Field, StrictInt, model_validator
 
-from jacobian._exact import CanonicalRational
+from jacobian._exact import CanonicalRational, require_bounded_rational
 from jacobian._models import StrictModel
+from jacobian.math.matrices.values import RationalMatrix, require_matrix_scalar_digits
+
+MAX_RATIONAL_SPECTRUM_ORDER = 32
+MAX_RATIONAL_SPECTRUM_CLAIMS = 32
+MAX_RATIONAL_SPECTRUM_INPUT_DIGITS = 64
+MAX_RATIONAL_SPECTRUM_NONZERO_ENTRIES = 1_024
+MAX_RATIONAL_SPECTRUM_SHIFTED_DIGITS = 129
+MAX_RATIONAL_SPECTRUM_RANK_WORK = 1_048_576
+MAX_RATIONAL_SPECTRUM_MINOR_DIGITS = 132_256
+MAX_RATIONAL_SPECTRUM_RESULT_BYTES = 384 * 1024
+
+
+def _component_digits(value: CanonicalRational) -> int:
+    return max(len(value.num.lstrip("-")), len(value.den))
+
+
+class RationalSpectrumMultiplicityClaim(StrictModel):
+    """One distinct rational eigenvalue and its claimed positive multiplicity."""
+
+    eigenvalue: CanonicalRational
+    multiplicity: StrictInt = Field(ge=1, le=MAX_RATIONAL_SPECTRUM_ORDER)
+
+
+class RationalSpectrumClaimRequest(StrictModel):
+    """A complete rational spectrum claim for one symmetric matrix over QQ.
+
+    ``matrix`` is the canonical materialized rational-matrix value. It must be
+    square, symmetric, and have order at most 32. ``claimed_profile`` contains
+    1..32 pairwise-distinct rational eigenvalues with positive multiplicities.
+    Matrix and claim rational components are limited to 64 decimal digits.
+    """
+
+    matrix: RationalMatrix = Field(
+        description=(
+            "Canonical materialized square symmetric matrix over QQ, of order "
+            "at most 32, with rational components of at most 64 decimal digits."
+        )
+    )
+    claimed_profile: tuple[RationalSpectrumMultiplicityClaim, ...] = Field(
+        min_length=1,
+        max_length=MAX_RATIONAL_SPECTRUM_CLAIMS,
+        description=(
+            "Nonempty list of at most 32 pairwise-distinct rational eigenvalues "
+            "and positive claimed multiplicities."
+        ),
+    )
+
+    @model_validator(mode="after")
+    def require_admitted_complete_claim(self) -> Self:
+        order = len(self.matrix.entries)
+        if order != len(self.matrix.entries[0]):
+            raise ValueError("rational spectrum claims require a square matrix")
+        if order > MAX_RATIONAL_SPECTRUM_ORDER:
+            raise ValueError(
+                "rational spectrum claims support matrix order at most "
+                f"{MAX_RATIONAL_SPECTRUM_ORDER}"
+            )
+        if any(
+            self.matrix.entries[row][column] != self.matrix.entries[column][row]
+            for row in range(order)
+            for column in range(row + 1, order)
+        ):
+            raise ValueError("rational spectrum claims require a symmetric matrix")
+
+        require_matrix_scalar_digits(
+            self.matrix.entries,
+            maximum=MAX_RATIONAL_SPECTRUM_INPUT_DIGITS,
+            label="rational spectrum matrix",
+        )
+        nonzero_entries = sum(
+            entry.num != "0" for row in self.matrix.entries for entry in row
+        )
+        if nonzero_entries > MAX_RATIONAL_SPECTRUM_NONZERO_ENTRIES:
+            raise ValueError(
+                "rational spectrum matrix exceeds the nonzero-entry budget"
+            )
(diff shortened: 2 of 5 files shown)


morluto:
@codex

PR #2330, opened 2026-08-23
feat(math): add complete indexed subset-sum multiplicity profiles

### LINKED ISSUE #2287: [Operation]: Return complete exact subset-sum multiplicity profiles for bounded indexed inputs
## Parent gap / RFC

The Atlas gap-map audit found repeated demand in Erdős problems 1, 131, 301, 470, and 1108. The existing operation from #1682 counts ordered pairs from two supplied sets; it does not represent selection of indexed items from one sequence.

The clearest fixture is [Erdős problem 1](https://github.com/techno-optimist/erdos-frontier-atlas/blob/0394e3d3b249439ffabec7d96a3311aa441651b8/views/graph/P1.md), where all `2^11 = 2048` indexed subset sums must be distinct.

Return the complete exact map from integer sums to the number of index subsets of a bounded integer sequence attaining each sum.

## Proposed operation ID

`additive.subset_sum.profile.compute`

## Input

A finite indexed tuple of integers. Positions are selectable objects: repeated values and zeros remain distinct inputs. The empty subset is included by definition in this complete-profile operation.

## Result

A canonical source-bound profile of records `(sum, multiplicity)`, sorted by sum, with positive exact integer multiplicities and enough retained source information to replay the profile.

Optional collision witnesses may be returned for multiplicities greater than one, but they must be index sets rather than value multisets.

## Completeness semantics

Every accepted request returns the complete profile. There is no truncation, sampling, cutoff-dependent `None`, or implication of global nonexistence. Requests whose predicted work or result exceed admission are rejected before enumeration.

## Bounds and cost class

Preflight must bound independently:

- item count and `2^n`;
- input magnitude and total absolute sum span;
- DP/table cells or meet-in-the-middle rows;
- maximum multiplicity digits;
- output rows and serialized bytes.

The admission rule may choose between exact enumeration, integer DP, or meet-in-the-middle privately, but public completeness cannot depend on an undocumented fallback.

## Maintained implementation

A bounded Jacobian-owned dynamic-programming or meet-in-the-middle kernel is proportionate. The implementation should preserve index multiplicity and exact counts rather than first converting the tuple to a set.

## Defining invariant

For input `a_0, ..., a_(n-1)`, each output multiplicity is

```text
m(s) = #{ I subseteq {0, ..., n-1} : sum(i in I) a_i = s }.
```

The profile must satisfy `sum_s m(s) = 2^n`, include `m(0) >= 1`, and reconstruct exactly under a bounded independent enumeration oracle.

## Evidence plan

- empty input gives `{0: 1}`;
- `[1, 1]` gives `{0: 1, 1: 2, 2: 1}`;
- zeros increase multiplicity without changing support;
- mixed positive and negative inputs;
- permuting the indexed input preserves the numeric profile;
- the Conway–Guy 11-set recorded by Atlas has 2,048 support values, each of multiplicity one;
- plant one repeated subset sum and require the corresponding multiplicity/collision;
- compare with brute-force bitmask enumeration for all small tuples;
- mutate a source value, profile row, or multiplicity and reject.

Repeated binary sumset calls lose the exclusion constraint, index identity, and exact multiplicities needed for subset selection. The current ordered `A + B` profile cannot express choosing each indexed item at most once.

## Deferred operation family

Target-only solving, modular-residue profiles, minimum-cardinality representations, global distinct-subset-sum optimization, and exhaustive lower-bound searches are separate contracts.

morluto:
## Problem

The additive-combinatorics catalog can count ordered sums from two supplied
sets, but it cannot represent selecting each position of one integer sequence
at most once. Repeated values, zeros, and exact indexed-subset multiplicities
therefore cannot be recovered by composing the existing pair-sum profile.

Closes #2287.

## Solution

Add `additive.subset_sum.profile.compute`, which accepts a canonical
`IndexedIntegerSequence` and returns the complete sorted map from each attainable
integer sum to its positive exact multiplicity. The empty subset is included;
equal values and zeros remain distinct selectable positions. The source,
`support_size`, and `total_subsets` are retained, and result validation replays
the bounded computation so mutated sources, rows, multiplicities, and totals do
not revalidate.

The same canonical sequence, profile values, and `subset_sum_profile` kernel are
available through `jacobian.math.additive_combinatorics` for direct typed
composition. Target-only solving, modular profiles, witnesses, and optimization
remain separate contracts.

### Library choice

This uses an exact sparse integer DP rather than adding a dependency. SymPy's
documented combinatorial enumerators do not expose this indexed, at-most-once,
all-sums multiplicity postcondition; adapting them would still require custom
counting and work accounting. The local kernel preserves positional
multiplicity directly and bounds the quantities it expands before execution.

## Testing

- `make check` — Ruff and strict mypy pass; 2,671 Lean-free owner tests pass.
- `make test-math TESTS='tests/math/additive_combinatorics tests/math/public_api/test_namespace.py'` — 67 passed.
- `uv run --locked pytest -q tests/integration/catalog/test_builtin_examples.py tests/integration/catalog/test_public_operation_contract_conformance.py tests/catalog/test_admission.py` — 976 passed.
- The operation-specific tests compare every tuple of length 0 through 4 over `[-2, 2]` with an independent bitmask oracle. They also cover empty input, repeats, zeros, mixed signs, permutation invariance, forged results, admission boundaries, an 8,192-row result over 256-digit source values, 256 zeros, and the Conway–Guy 11-set's 2,048 distinct subset sums.

## Public contract impact

Adds public operation `additive.subset_sum.profile.compute`, its strict request
and source-bound result schema, and the native
`jacobian.math.additive_combinatorics` value/function surface. Existing
operation IDs and request/result semantics are unchanged.

## Public operation admission

- Concrete gap: no operation returns complete exact subset-sum multiplicities for one indexed source.
- Why existing operations or typed values are insufficient: repeated binary sumsets lose the per-position exclusion constraint and indexed multiplicity; `additive.representation_profile.compute` counts pairs from two sets.
- Stable mathematical result: the complete function `m(s) = #{I subseteq {0,...,n-1} : sum(i in I) a_i = s}`, represented as unique sum-sorted rows with `sum_s m(s) = 2^n`.
- Semantic domain and admitted execution envelope: an ordered tuple of at most 256 canonical integers, each at most 256 decimal digits excluding its optional sign. For `S = min(2^n, positive_sum-negative_sum+1, product(m_v+1) over distinct nonzero values v)`, admission requires `S <= 50,000`, `4*n*S <= 4,000,000`, and a conservative serialized result estimate at most 4 MiB.
- Controlling work, intermediate, memory, and output quantities: two sparse-DP passes (construction and validation replay), at most `S` live sums per pass, bounded integer/sum and multiplicity digits, at most 50,000 result rows, and a 4 MiB result envelope.
- Exact representation, algorithm regimes, and maintained backend: canonical decimal integers and a deterministic Jacobian-owned sparse DP over `Z`; no backend coercion, ambient context, approximation, truncation, or fallback regime.
- Remaining fixed caps and their classification: 256 items and 256 digits conservatively bound input/bigint work; 4,000,000 transitions bound execution; 50,000 rows and 4 MiB bound the complete exact output. Result-sensitive admission still accepts large low-support inputs such as 256 zeros.
- Admission decision: `KEEP` — this is a reusable complete profile that cannot be projected from the existing pair-sum operation.

### Suggested review order

Start with the canonical values and source-binding validator, then check the
support/work/output envelope beside the sparse DP. The request, catalog wiring,
and admission row are thin projections of that contract; the focused tests
exercise the defining invariant and each bound.

## Closure matrix

None. This issue requests one operation; target-only, modular, witness, and optimization variants remain explicitly deferred by the issue.

## Checklist

- [x] `make check` passes
- [x] Explicitly relevant catalog and domain validation is listed above
- [x] Harbor task or verifier validation is not applicable; neither changed
- [x] The result is exact and complete; over-envelope requests are rejected before computation
- [x] New bounds include boundary and realistic source-backed scale evidence
- [x] Shared-abstraction replacement is not applicable; the new types are domain-owned canonical mathematical values

### DIFF
--- a/src/jacobian/math/__init__.py
+++ b/src/jacobian/math/__init__.py
@@ -1,6 +1,7 @@
 """Native-value mathematical APIs supported by Jacobian."""
 
 from jacobian.math import (
+    additive_combinatorics,
     algebraic_combinatorics,
     arithmetic,
     arithmetic_dynamics,
@@ -34,6 +35,7 @@ from jacobian.math import (
 )
 
 __all__ = [
+    "additive_combinatorics",
     "algebraic_combinatorics",
     "arithmetic",
     "arithmetic_dynamics",
--- a/src/jacobian/math/additive_combinatorics/__init__.py
+++ b/src/jacobian/math/additive_combinatorics/__init__.py
@@ -1,3 +1,15 @@
-"""Additive-combinatorics operation ownership."""
+"""Supported native additive-combinatorics API."""
 
-__all__: list[str] = []
+from jacobian.math.additive_combinatorics.operations import subset_sum_profile
+from jacobian.math.additive_combinatorics.values import (
+    IndexedIntegerSequence,
+    SubsetSumProfile,
+    SubsetSumProfileEntry,
+)
+
+__all__ = [
+    "IndexedIntegerSequence",
+    "SubsetSumProfile",
+    "SubsetSumProfileEntry",
+    "subset_sum_profile",
+]
--- a/src/jacobian/math/additive_combinatorics/_admission.py
+++ b/src/jacobian/math/additive_combinatorics/_admission.py
@@ -35,6 +35,11 @@ ADMISSIONS: tuple[OperationAdmission, ...] = (
         AdmissionDecision.DROP,
         "cheap deterministic projection of additive.representation_profile.compute",
     ),
+    OperationAdmission(
+        "additive.subset_sum.profile.compute",
+        AdmissionDecision.KEEP,
+        "one complete source-bound profile of sums and multiplicities over indexed at-most-once selections, which pair-sum composition cannot retain",
+    ),
 )
 
 REGISTRATION = OperationRegistration(TOOLS, ADMISSIONS)
--- a/src/jacobian/math/additive_combinatorics/_models.py
+++ b/src/jacobian/math/additive_combinatorics/_models.py
@@ -11,6 +11,15 @@ from pydantic import Field, StringConstraints, model_validator
 from jacobian._exact import CanonicalInteger
 from jacobian._models import StrictModel
 from jacobian.canonical import format_canonical_integer, parse_canonical_integer
+from jacobian.math.additive_combinatorics.operations import (
+    MAX_SUBSET_SUM_DP_TRANSITIONS,
+    MAX_SUBSET_SUM_PROFILE_RESULT_BYTES,
+    _subset_sum_profile_envelope,
+)
+from jacobian.math.additive_combinatorics.values import (
+    MAX_SUBSET_SUM_PROFILE_ENTRIES,
+    IndexedIntegerSequence,
+)
 
 _MAX_SET_SIZE = 256
 _MAX_RESULT_SIZE = _MAX_SET_SIZE * _MAX_SET_SIZE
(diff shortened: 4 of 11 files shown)


morluto:
@codex

PR #2341, opened 2026-08-23
feat(math): add bounded generalized exact cover

### LINKED ISSUE #2256: Add bounded generalized exact cover with primary and secondary items
## Observed mathematical tasks

The trajectory repeatedly needed a conflict-free transversal while assigning witnesses to all four-point constraints. It implemented a dedicated recursive solver with:

- one group of candidate rows for each required constraint;
- exactly one selected row from every required group;
- resource incidences that could be used at most once;
- most-constrained-first branching; and
- complete success/failure search.

A later six-point enumeration used the same semantics explicitly as 15 required groups and 60 secondary resources.

## Missing mathematical result

Expose bounded generalized exact cover with Knuth-style primary and secondary items:

```text
primary item: covered exactly once
secondary item: covered at most once
row: finite subset of primary and secondary items
```

The result should be either one checked selected-row family or exact nonexistence after complete bounded search. A solution must reconstruct coverage multiplicities for every item. If the public envelope cannot guarantee completion, the request should be rejected before execution or the result must distinguish operational incompleteness from `NO_COVER`.

## Bespoke escape observed

The agent wrote and optimized its own exact-cover kernel, resource bitmasks, branching heuristic, and state memoization. This code was independent of the circumradius formulas and would recur in many finite incidence and witness-selection investigations.

## Gap classification

Potential operation gap. SAT can encode the constraints and #1989 proposes a broad finite-CSP owner, but neither currently gives agents this standard typed combinatorial postcondition without a caller-authored encoding and decoding protocol.

## Admission posture

Resolve ownership before public admission. The preferred outcome may be a focused exact-cover value/operation or a concrete projection of the finite-CSP work in #1989. It should not become a generic solver DSL, mutable search session, or theorem-specific witness-assignment workflow.

## Existing discovery queries and neighboring operations

Queries for `exact cover`, `generalized exact cover`, and `primary secondary columns` found no focused current owner. `sat.solve` is an encoding-level alternative, and #1989 proposes bounded CSP solution operations. The Latin-square and finite-model issues use neighboring search techniques but do not expose generalized exact cover as a reusable value.

## Why existing composition is insufficient

Using SAT requires the caller to invent Boolean variables, encode exactly-one and at-most-one clauses, bind model indices back to row identities, replay item multiplicities, and decide what solver failure means. The bespoke trajectory code exists because the mathematical input and output are incidence data, not arbitrary clauses.

## Evidence and downstream uses

- select compatible local witnesses across a family of constraints;
- exact transversal and packing constructions;
- incidence-design and tiling subproblems;
- complete feasibility checks inside bounded symmetry searches; and
- return selected ordinary rows that downstream code can inspect directly.

## Acceptance criteria

- [ ] Primary items, secondary items, and row identities are explicit finite typed values.
- [ ] Every accepted row references only declared items and duplicate incidences have defined normalization or rejection semantics.
- [ ] A positive result covers each primary item exactly once and each secondary item at most once.
- [ ] A negative result is returned only after complete search or independently replayed unsatisfiability evidence.
- [ ] Row/item/incidence, candidate-state, integer, and output bounds are enforced before or during work with honest failure semantics.
- [ ] Row and item relabelling transport solutions without changing feasibility.
- [ ] The operation remains stateless and does not expose branching heuristics, checkpoints, or a solver-expression language.

## Related

- #1989 — finite relational structures and CSPs.
- `sat.solve` — a lower-level Boolean encoding route.

morluto:
## Problem

Agents that need to choose one compatible witness for each required constraint currently have to encode generalized exact cover as SAT, maintain a separate row-to-variable map, decode an arbitrary model, and reconstruct coverage multiplicities. That obscures the primary/exactly-once and secondary/at-most-once semantics, and makes it easy to treat a bounded search stop as nonexistence.

Closes #2256.

## Solution

Add `combinatorics.generalized_exact_cover.find` over a canonical materialized incidence value. Primary items must be covered exactly once, secondary items at most once, and distinct row IDs remain distinct candidates even when their incidences agree.

The operation uses a request-scoped bitset implementation of Algorithm X. It chooses the uncovered primary item with the fewest available rows, breaks ties by canonical item order, and visits rows in canonical row-ID order. The result is therefore deterministic for a canonical input and node allowance:

- `FOUND` carries selected row IDs and a reconstructed multiplicity for every declared item;
- `NO_COVER` is returned only after exhaustive search and is checked by source-bound replay; and
- `UNKNOWN` is an explicit non-conclusion at the node allowance. It is source-bound but intentionally not replayed, so a private kernel improvement cannot invalidate an old non-conclusion.

The public surface stays at exact-one/at-most-one feasibility. It does not add weighted set cover, optimization, generated candidate families, checkpoints, or durable search state.

## Testing

- `make check` — 2,666 passed; Ruff and mypy passed.
- `uv run --locked pytest -q tests/math/combinatorics/test_exact_cover.py tests/math/combinatorics/test_public_api.py tests/catalog/test_catalog.py tests/tooling/test_pytest_collection_ownership.py tests/integration/catalog/test_public_operation_contract_conformance.py` — 515 passed.
- Specialist validation run: catalog public-operation mutation conformance is included in the focused command above.

The tests include Knuth's known-answer example, all 128 subfamilies of the seven nonempty rows over two primary and one secondary item against an independent exhaustive oracle, the pinned Erdős 743 K3 packing control and an overfull mutation, relabelling, source mutations, empty and uncovered-primary degeneracies, exact count/output boundaries, and `UNKNOWN` versus `NO_COVER` behavior. The exhaustive oracle now requires infeasible families to return `NO_COVER`; `UNKNOWN` fails the test.

Indicative single-run measurements used `time.perf_counter` around request construction plus the public function on CPython 3.12.13, Linux x86-64, and an Intel i7-10875H. They establish usefulness, not the work proof:

| Instance | Bound/result | Elapsed |
| --- | --- | ---: |
| 129 primary items, one row containing all of them | 2 nodes / `FOUND` | 0.007405 s |
| 1 primary item, 4,096 candidate rows | 2 nodes / `FOUND` | 0.191453 s |
| 83 pigeons, 48 secondary holes, all 3,984 assignment rows | 100,000 nodes / `UNKNOWN` | 1.545194 s |

The last case exercises the maximum-node fallback with nearly the maximum row-bitset width. A previous 1,000,000-node adversarial pass was too costly to repeat during exact-negative validation, so the public ceiling is 100,000 nodes per pass; `NO_COVER` construction plus replay is bounded to 200,000 visits.

## Public contract impact

Adds operation version 1, the native `find_generalized_exact_cover` function, and these canonical/source-bound types under `jacobian.math.combinatorics`:

- `ExactCoverRow`
- `GeneralizedExactCoverInstance`
- `GeneralizedExactCoverRequest`
- `GeneralizedExactCoverResult`
- `ExactCoverItemMultiplicity`
- `ExactCoverSearchStatus`

## Public operation admission

- Concrete gap: generalized exact cover is a reusable incidence postcondition. The existing `sat.solve` route leaves Boolean encoding, row identity, model decoding, multiplicity reconstruction, and incomplete-search interpretation to every caller.
- Why existing operations or typed values are insufficient: finite-set exact-cover checks validate one supplied family; they do not search primary/secondary incidence rows. SAT accepts clauses rather than this mathematical value and does not return selected ordinary rows.
- Stable mathematical result: one row family covering each primary item exactly once and each secondary item at most once, exact nonexistence after exhaustive bounded search, or an explicit non-conclusion.
- Semantic mathematical domain and postcondition: finite materialized generalized exact-cover instances with explicit item and row identities. Duplicate incidences, undeclared items, overlapping primary/secondary declarations, and noncanonical ordering are rejected.
- Canonical public value type: `GeneralizedExactCoverInstance` owns the ordered primary items, secondary items, and identified `ExactCoverRow` values. Results retain that instance unchanged.
- Source representation: materialized. The kernel does not expand a succinct generator or query an oracle.
- Expansion performed by the kernel and its pre-execution bound: for `P` primary items, `M` total items, `R` rows, and `I` incidences, index construction materializes `I` item indices, `M` row masks, `R` primary masks, and `R` conflict masks using at most `2I` mask updates.
- Representation-sensitive complexity: row IDs distinguish candidates; equal incidence sets are not normalized together. Generated or symmetry-reduced row families remain caller-owned because their completeness proof is outside this operation.
- Semantic domain and admitted execution envelope: `M <= 256`, `R <= 4,096`, `I <= 65,536`, at most 256 selected rows, and at most 100,000 search nodes per pass. Opaque-label and canonical JSON limits also apply. The prior independent 128-primary ceiling was removed; a 129-primary one-row instance and the 256-primary/256-selected-row boundary execute successfully.
- Producer/consumer closure: no existing operation produces this incidence value. Its canonical serialization can be supplied unchanged to the request, and the result retains the full source plus ordinary row IDs and multiplicities for caller-side composition. Empty-primary instances retain their secondary axis and return the empty selected family.
- Degenerate inputs: an empty primary family has the empty cover; required items with no candidate row prove `NO_COVER` at the root; rows containing no primary item are valid source data but never appear in the canonical witness.
- Parent/ring/field identity: not applicable. Opaque item and row IDs define one finite incidence instance; no coercion or ambient algebraic parent exists.
- Deterministic work bound: index construction uses at most `2I` mask updates. Each node performs at most `P` intersections/population counts on `R`-bit masks plus constant-many child-mask operations, with at most `N-1` child edges for `N <= 100,000`. Exact-negative result replay makes at most two passes.
- Maximum intermediate growth: the logical index payload is `M*R + R*P + R^2` bits plus `I` bounded item indices, at most 18,874,368 mask bits under the public envelope. Recursion depth is at most `P <= 256`; frame mask payload is bounded by `(P+1)*(P+R)` bits plus `P` selected-row indices. Python object overhead is additionally bounded by the same item, row, incidence, and depth counts.
- Exact result-size bound: a witness has at most `min(P, R)` selected IDs and exactly `M` multiplicity records. Admission materializes and canonically encodes the worst `FOUND` shape using the longest possible selected row IDs, plus both non-positive shapes, each retaining the source. Search starts only if every shape fits the shared 10 MiB output limit.
- Exact representation, algorithm regimes, and maintained backend: one Jacobian-owned bitset Algorithm X regime over canonical ordinal indices. The installed Z3 backend can encode these Boolean constraints and distinguish `sat`, `unsat`, and `unknown`, but it returns an arbitrary model rather than the canonical row witness and its resource units follow solver internals. Canonicalizing that model would require repeated solver calls or optimization and a second row-variable protocol. No maintained exact-cover dependency already in the stack supplies the complete typed postcondition; adding a larger solver package is disproportionate to this bounded 127-line kernel.
- Backend and supported version: no external runtime backend. The kernel is ordinary supported Python and uses arbitrary-precision integer bitsets.
- Backend input domain: the already-validated canonical instance; no backend-specific domain is discovered during execution.
- Conversion/coercion behavior: item and row labels map bijectively to their canonical tuple positions. No label is parsed or evaluated, no duplicate is normalized away, and output maps selected positions back to the original row IDs.
- Result type: source-bound `GeneralizedExactCoverResult` with `FOUND`, `NO_COVER`, or `UNKNOWN` status.
- Reconstruction or defining invariant: `FOUND` validation recomputes every primary and secondary multiplicity from the retained rows. `NO_COVER` validation repeats the bounded exhaustive search and rejects a contradictory replay. `UNKNOWN` carries no mathematical conclusion and therefore does not turn a private search convention into durable public semantics.
- Typed execution failures: malformed or over-envelope instances fail request validation; reaching the admitted node allowance returns `UNKNOWN`; every accepted request returns the declared result type.
- Property and boundary tests: known answers, independent exhaustive feasible/infeasible classification, primary/secondary conflict cases, relabelling, empty inputs, source and witness mutations, 4,096-row and 65,536-incidence boundaries, 129- and 256-primary execution, exact retained-output preflight, catalog discovery, and public mutation conformance.
- Remaining fixed caps and their classification: the 256-item, 4,096-row, and 65,536-incidence ceilings are conservative materialization/memory fallbacks and can be raised after larger logical-mask and canonical-output measurements. The 100,000-node ceiling is an execution-safety fallback and can be raised if a faster maintained backend or a compact independently checked negative certificate removes replay cost. Opaque-label and 10 MiB JSON limits are shared transport bounds, not mathematical restrictions.
- Admission decision: `KEEP`, recorded in the combinatorics owner.

## Closure matrix

None. This PR admits the single focused operation from #2256; weighted or at-least-one covering remains outside its semantics.

## Suggested review order

1. Public value, result, and bound contract in `exact_cover.py`.
2. Bitset work/intermediate proof and search in `_exact_cover_kernel.py`.
3. Known-answer, exhaustive-oracle, mutation, and boundary tests.
4. Catalog declaration and owner-local admission.

## Checklist

- [x] `make check` passes
- [x] Harbor task or verifier changes ran their dedicated validation (not applicable; no Harbor files changed)
- [x] Result semantics distinguish exact and unknown outcomes
- [x] New bounds include boundary and realistic source-backed scale evidence
- [x] New shared abstractions replace duplication in at least two production paths (not applicable; no shared abstraction was added)

### DIFF
--- a/src/jacobian/math/combinatorics/__init__.py
+++ b/src/jacobian/math/combinatorics/__init__.py
@@ -1,5 +1,14 @@
 """Provider-independent exact combinatorics values and functions."""
 
+from jacobian.math.combinatorics.exact_cover import (
+    ExactCoverItemMultiplicity,
+    ExactCoverRow,
+    ExactCoverSearchStatus,
+    GeneralizedExactCoverInstance,
+    GeneralizedExactCoverRequest,
+    GeneralizedExactCoverResult,
+    find_generalized_exact_cover,
+)
 from jacobian.math.combinatorics.operations import (
     bell_number,
     bernoulli_number,
@@ -22,6 +31,12 @@ from jacobian.math.combinatorics.recurrence_tables import (
 )
 
 __all__ = [
+    "ExactCoverItemMultiplicity",
+    "ExactCoverRow",
+    "ExactCoverSearchStatus",
+    "GeneralizedExactCoverInstance",
+    "GeneralizedExactCoverRequest",
+    "GeneralizedExactCoverResult",
     "IndexedRecurrenceResidual",
     "PolynomialCoefficientRecurrenceTableRequest",
     "PolynomialCoefficientRecurrenceTableResult",
@@ -31,6 +46,7 @@ __all__ = [
     "derangement_number",
     "double_factorial",
     "fibonacci_number",
+    "find_generalized_exact_cover",
     "integer_partitions",
     "lucas_number",
     "motzkin_number",
--- a/src/jacobian/math/combinatorics/_admission.py
+++ b/src/jacobian/math/combinatorics/_admission.py
@@ -128,6 +128,11 @@ ADMISSIONS: tuple[OperationAdmission, ...] = (
         native_symbol="jacobian.math.combinatorics.integer_partitions",
     ),
     OperationAdmission(
+        "combinatorics.generalized_exact_cover.find",
+        AdmissionDecision.KEEP,
+        "one checked primary/secondary exact-cover witness or exact bounded nonexistence over materialized incidence rows",
+    ),
+    OperationAdmission(
         "combinatorics.generating_function.coefficients.compute",
         AdmissionDecision.KEEP,
         "distinct exact bounded mathematical value or invariant with material computational or reliability leverage",
--- a/src/jacobian/math/combinatorics/_exact_cover.py
+++ b/src/jacobian/math/combinatorics/_exact_cover.py
@@ -1,0 +1,62 @@
+"""Public declaration for bounded generalized exact cover."""
+
+from jacobian.catalog._examples import example
+from jacobian.catalog.models import MathTool
+from jacobian.math.combinatorics.exact_cover import (
+    GeneralizedExactCoverRequest,
+    GeneralizedExactCoverResult,
+    find_generalized_exact_cover,
+)
+
+GENERALIZED_EXACT_COVER_OPERATION = MathTool(
+    operation_id="combinatorics.generalized_exact_cover.find",
+    version="1",
+    title="Find a generalized exact cover",
+    description=(
+        "Find one row family that covers every primary item exactly once and "
+        "every secondary item at most once. Return FOUND with a checked "
+        "selected-row family, NO_COVER only after complete bounded search, or "
+        "UNKNOWN when the deterministic search-node limit is reached."
+    ),
+    request_type=GeneralizedExactCoverRequest,
+    result_type=GeneralizedExactCoverResult,
+    run=find_generalized_exact_cover,
+    tags=(
+        "combinatorics",
+        "exact-cover",
+        "generalized-exact-cover",
+        "primary-items",
+        "secondary-items",
+        "incidence",
+        "algorithm-x",
+        "bounded-search",
+        "deterministic",
+    ),
+    examples=(
+        example(
+            "two_constraints_one_resource",
+            "Select one row for each of two primary constraints while using "
+            "the optional resource at most once; item and row labels must be "
+            "declared in sorted canonical order.",
+            {
+                "instance": {
+                    "primary_items": ["constraint:a", "constraint:b"],
+                    "secondary_items": ["resource:x"],
+                    "rows": [
+                        {
+                            "row_id": "a-use-x",
+                            "items": ["constraint:a", "resource:x"],
+                        },
+                        {
+                            "row_id": "b-free",
+                            "items": ["constraint:b"],
+                        },
+                    ],
+                },
+                "search_node_limit": 100,
+            },
+        ),
+    ),
+)
+
+__all__ = ["GENERALIZED_EXACT_COVER_OPERATION"]
--- a/src/jacobian/math/combinatorics/_exact_cover_kernel.py
+++ b/src/jacobian/math/combinatorics/_exact_cover_kernel.py
@@ -1,0 +1,127 @@
+"""Deterministic bounded Algorithm X kernel for generalized exact cover."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Literal
+
+from jacobian.math.combinatorics.exact_cover import GeneralizedExactCoverInstance
+
+
+@dataclass(frozen=True, slots=True)
+class ExactCoverKernelResult:
+    status: Literal["FOUND", "NO_COVER", "UNKNOWN"]
+    selected_rows: tuple[int, ...] = ()
+
+
+def search_generalized_exact_cover(
+    instance: GeneralizedExactCoverInstance,
+    search_node_limit: int,
+) -> ExactCoverKernelResult:
+    """Search one materialized primary/secondary incidence matrix.
+
+    A node is one conflict-free partial selected-row family, represented by
+    its uncovered-primary and available-row bitsets. The root and terminal
+    states count as nodes. Branching chooses the uncovered primary item with
+    the fewest available rows, ties by canonical item order, and tries rows in
+    canonical row-ID order. Thus the same canonical input and node limit always
+    produce the same status and witness.
+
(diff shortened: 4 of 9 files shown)


morluto:
@codex

Issue #974, opened 2026-08-09
[Capability]: Add exact polygon-kernel and bound measure certificates


yuelgrace1810-ops:
## Capability gap

Jacobian has useful polygon primitives (signed area, convex hull, segment and line intersections), but no composed exact operation that reconstructs a simple polygon's visibility kernel from oriented half-planes and binds the resulting areas/perimeters/ratios to one input polygon.

## Held-out reproduction

Nakano's arXiv:2606.05052v2 publishes the counterclockwise pentagon

`(0,4620),(0,-4620),(23100,-385),(22176,0),(23100,385)`.

An independent exact oracle gives:

- orientation crosses, centered at `v1,...,v5`: `213444000,213444000,12806640,-711480,12806640`;
- unique reflex vertex `v4`;
- kernel `(0,-4620),(19800,-990),(22176,0),(19800,990),(0,4620)`;
- areas `|F|=115259760`, `|K|=113430240`, `|C|=115615500`;
- perimeters `Per(F)=58212`, `Per(C)=56980`;
- `G=62/63`, `P=185/189`, and `G-P=1/189>0`.

In a selectively grounded weak-model run, discovery found the four existing geometry primitives, but the only `math.run` was cancelled. Manual fallback then flipped the first half-plane to `x<=0` (the correct interior is `x>=0`), collapsed the kernel to a segment, and reversed the theorem's conclusion. This routing behavior is separately recorded on #845.

## Proposed bounded contract

Consider a checker-backed family such as:

- `geometry.polygon.kernel.compute/verify`: explicit cyclic rational vertices -> oriented half-planes, exact kernel vertices, emptiness/dimension, canonical digest;
- `geometry.polygon.measure_certificate.compute/verify`: one bound polygon plus derived hull/kernel -> signed orientations, exact shoelace areas, exact squared lengths (and radicals/integer lengths), perimeters, and caller-specified rational comparisons.

The result should preserve cyclic ordering and sign convention explicitly. A verifier should independently replay every half-plane inequality and every reported intersection, not merely re-hash producer output.

## Acceptance criteria

1. Reproduce the frozen pentagon and exact kernel above.
2. Identify `v4` as the unique reflex vertex without index ambiguity.
3. Bind hull, kernel, areas, lengths, perimeters, and final ratio gap to the exact input/order.
4. Reject clockwise/mixed-order input or normalize it with explicit provenance.
5. Distinguish empty, lower-dimensional, and positive-area kernels.
[...]

PR #2361, opened 2026-08-24
Add exact visibility kernels for rational polygons

morluto:
## Problem

Part of #974. Jacobian could compute individual rational-polygon predicates, but it had no source-bound operation for the visibility kernel of one simple counter-clockwise polygon. Callers therefore had to assemble oriented half-planes, enumerate the lower-dimensional cases, and recompute areas themselves.

## Solution

Add `geometry.polygon.visibility_kernel.compute`. It accepts a bounded simple counter-clockwise rational polygon and returns its oriented-edge half-planes, turns/reflex vertices, exact kernel as `EMPTY`, `POINT`, `SEGMENT`, or `POLYGON`, active source edges, convex hull, exact areas, and the kernel/polygon and kernel/hull area ratios.

The kernel uses exact `Fraction` arithmetic: pairwise half-plane boundary intersections plus the polygon vertices are the complete candidate set in two dimensions. The result replays source binding, every half-plane inequality, active-edge membership, canonical cyclic order, and area/ratio identities. No generic verifier or durable certificate surface is introduced.

Perimeters and the `G-P` comparison remain out of scope because the current canonical values have no exact finite radical-sum carrier.

## Testing

- `make check` — 2,670 passed.
- Polygon-kernel and namespace tests — 28 passed.
- Catalog examples and public-operation conformance — 2 passed.
- Independent exact clipping oracle — 850 generated simple polygons matched, covering 456 empty, 1 point, and 393 polygon kernels; 250 additional cases matched area and ratio calculations.

## Public contract impact

Adds `geometry.polygon.visibility_kernel.compute` with its bounded request/result schema. Geometry remains catalog-owned: no MCP request or result model is exposed as a native-value API.

## Public operation admission

- Concrete gap: compute the complete visibility kernel and bound measures of one typed rational polygon.
- Why existing operations or typed values are insufficient: the existing primitives do not return the composed half-plane intersection, its complete dimension classification, source-edge binding, and exact measures.
- Stable mathematical result: the exact kernel of a simple counter-clockwise polygon, with its defining half-plane representation and exact area projections.
- Semantic domain and admitted execution envelope: simple counter-clockwise rational polygons with at most 64 vertices and operation-owned coordinate bounds.
- Controlling work, intermediate, memory, and output quantities: at most 64 source vertices and 4,096 boundary pairs; every candidate, active edge, serialized result, and exact coordinate is bounded before construction.
- Exact representation, algorithm regimes, and maintained backend: domain-owned exact `Fraction` arithmetic and existing canonical geometry values; this finite two-dimensional enumeration has no dependency requiring a new backend.
- Remaining fixed caps and their classification: the 64-vertex and coordinate/result limits are conservative execution-envelope caps, not a restriction on the mathematical definition.
- Admission decision: keep.

## Closure matrix

| --- | --- | --- |
| Exact polygon visibility kernel and area measures | delivered | `geometry.polygon.visibility_kernel.compute` |
| Exact perimeters and `G-P` comparison | deferred: no finite radical-sum carrier | #974 |

## Checklist

- [x] `make check` passes
- [x] Result semantics distinguish exact kernel dimensions without incomplete or unavailable states
- [x] New bounds include boundary and independent scale evidence

### DIFF
--- a/src/jacobian/math/geometry/polygon_kernel/__init__.py
+++ b/src/jacobian/math/geometry/polygon_kernel/__init__.py
@@ -1,23 +1,3 @@
-"""Exact visibility kernels of simple rational polygons."""
+"""Polygon visibility-kernel operation ownership."""
 
-from jacobian.math.geometry.polygon_kernel._models import (
-    KernelBoundaryIntersection,
-    KernelPolygon,
-    OrientedEdgeHalfPlane,
-    PolygonKernelRequest,
-    PolygonKernelResult,
-    PolygonVertexTurn,
-)
-from jacobian.math.geometry.polygon_kernel._operations import (
-    compute_visibility_kernel,
-)
-
-__all__ = [
-    "KernelBoundaryIntersection",
-    "KernelPolygon",
-    "OrientedEdgeHalfPlane",
-    "PolygonKernelRequest",
-    "PolygonKernelResult",
-    "PolygonVertexTurn",
-    "compute_visibility_kernel",
-]
+__all__: list[str] = []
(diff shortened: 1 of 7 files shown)


morluto:
@codex

PR #2362, opened 2026-08-24
Add rational-box expression enclosure

### LINKED ISSUE #2303: [Operation]: Enclose bounded elementary expressions over rational boxes
## Parent gap / RFC

This is a focused extension of the elementary-expression interval work delivered by #1652. It is distinct from #2280, whose proposed box contract is intentionally polynomial-only.

The Erdős 1038 certificate repeatedly needs rigorous enclosures of coupled elementary expressions over rational boxes, not merely evaluation at one rational point. Its [proved expression evaluator](https://github.com/ShouqiaoW/erdos/blob/d28713ac8245ca86a686b8c67370a8d19d81b242/1038/lean/Erdos1038/OneCutIntervalExpr.lean#L18-L80) supports variables, rational constants, arithmetic, inversion, and logarithms, and proves the returned interval contains every real value in the source box. The numerical certificate extends the same move to `exp`, `sin`, and `cos` while checking derivative signs over coupled parameter boxes.

Given a bounded, non-evaluating elementary expression and an ordered rational box for its named variables, return a rigorous rational or dyadic interval containing every defined real value of the expression on that box.

## Proposed operation ID

`interval.expression.box_enclosure.compute`

## Input

- A canonical expression tree over an explicit ordered variable axis.
- A rational closed interval for every variable.
- A precision or enclosure-work budget.
- A small admitted grammar: rational constants, variables, `+`, unary `-`, `*`, integer powers, reciprocal, `exp`, `log`, `sqrt`, `sin`, and `cos`.

Public syntax must use the existing non-evaluating expression grammar; it must not reach `sympify`, `eval`, or `lambdify`.

## Result

A source-bound result containing the expression, variable axis, source box, enclosure interval, and precision/work metadata. Domain failure should identify the first subexpression whose entire source box cannot be shown to remain in its real domain.

## Completeness semantics

Every accepted request returns a sound enclosure. The operation need not return the exact range or decide a sign when the enclosure contains zero. A pole, nonpositive logarithm argument, or negative square-root argument that cannot be excluded on the whole box is a typed domain rejection, not a fabricated interval.

Automatic subdivision, global optimization, root isolation, and theorem-level sign conclusions are outside this postcondition.

## Bounds and cost class

Bound separately:

- variable count and expression-node count;
- source coefficient and endpoint digits;
- transcendental precision;
- integer exponents;
- backend intermediate precision and result digits.

The operation should reject above the admitted envelope before recursive evaluation. It should not silently subdivide a box or increase precision without a public bound.

## Maintained implementation

A thin in-process adapter to the maintained Arb/python-flint interval backend is the natural kernel. Jacobian should own the expression grammar, domain checks, work admission, source binding, and canonical interval result.

The Lean evaluator in the source repository is useful independent evidence, not a runtime dependency.

## Defining invariant

For every real point `x` in the supplied box at which the expression is defined, the returned interval `I` satisfies `eval(expression, x) in I`. The result must retain enough source data for an independent checker to replay that claim.

## Evidence plan

- Degenerate point boxes agree with `interval.compute.enclosure`.
- Monotone examples: `exp(x)` and `log(x)` on positive boxes.
- Trigonometric boxes that contain an interior extremum, so endpoint-only evaluation is caught.
- A dependency example such as `x - x`, whose enclosure may be conservative but must remain sound.
- Coupled multivariate expressions modeled on the Erdős 1038 `q,s,t,z` boxes.
- Reciprocal boxes separated from zero and boxes crossing zero.
- Invalid `log` and `sqrt` domains.
- Variable-axis permutation and source-expression mutation tests.
- Differential checks against high-precision sampling only as a secondary oracle; sampling never establishes soundness.

The current point operation establishes one value enclosure. Evaluating corners does not bound a general elementary expression over a box, and polynomial box evaluation does not cover transcendental nodes or their domain constraints. The missing postcondition is the uniform enclosure over the complete source box.

## Deferred operation family

- Automatic adaptive subdivision or branch-and-bound.
- Exact range computation.
- Symbolic directional derivatives and gradient/Hessian enclosures.
- Interval Newton/Krawczyk root certificates.
- Concluding positivity, uniqueness, or feasibility when an enclosure alone does not establish it.

morluto:
## Summary

Adds `interval.expression.box_enclosure.compute`, which returns a sound Arb enclosure for a restricted rational expression over an ordered rational interval box. The input retains its source expression, box, precision, and optional domain path; the result replays that source binding before it is returned.

The operation deliberately establishes only interval enclosure. It does not subdivide, differentiate, search for roots, or claim a sharp range. That keeps it useful as a composable primitive for later analyses without embedding a proof strategy.

The implementation uses python-flint's Arb real-ball arithmetic as the private natural-interval-extension backend. Jacobian owns the canonical rational syntax, ordered-box semantics, explicit status values, and conservative work/intermediate bounds. A deterministic domain path reports the first undefined operation instead of treating a backend error as a mathematical conclusion.

## Validation

- `make check` (2,687 passed)
- `make test-integration` (1,026 passed)
- Focused enclosure, strict-JSON composition, source-replay, boundary, and adversarial temporary-growth tests

An independent exact-diff review also verified that raw rational/container limits apply before nested parsing and that all Fraction candidates are bounded before comparison.

Closes #2303


morluto:
@codex

PR #2371, opened 2026-08-24
Measure strict polynomial sublevels exactly

### LINKED ISSUE #2308: [Operation]: Compute exact strict polynomial sublevel measures on intervals
## Parent gap / RFC

#1872 supplies much of the real-root and sign-chart machinery, but no current operation returns the exact measure of a strict polynomial sublevel set. The Erdős 1038 proof makes this value central: boundaries lie among roots of `f^2-t^2`, components are open intervals, and their lengths must be summed exactly.

See [`SublevelBoundary.lean`](https://github.com/ShouqiaoW/erdos/blob/d28713ac8245ca86a686b8c67370a8d19d81b242/1038/lean/Erdos1038/SublevelBoundary.lean#L18-L64) and [`SublevelComponents.lean`](https://github.com/ShouqiaoW/erdos/blob/d28713ac8245ca86a686b8c67370a8d19d81b242/1038/lean/Erdos1038/SublevelComponents.lean#L64-L228).

Given `f in QQ[x]`, a nonnegative rational threshold `t`, and a bounded rational interval, return the exact one-dimensional Lebesgue measure of `{x : |f(x)| < t}` inside that interval.

## Proposed operation ID

`polynomial.real.strict_sublevel_measure.compute`

## Input

- A canonical univariate rational polynomial.
- A rational threshold `t >= 0`.
- A rational closed scope interval `[a,b]`.

The strict inequality and endpoint convention must be part of the request semantics.

## Result

A source-bound exact real-algebraic measure, together with a canonical finite decomposition into contributing open intervals whose endpoints are rational scope endpoints or certified real algebraic roots of `f-t` and `f+t`. The result should retain enough data to reconstruct the exact sum.

## Completeness semantics

Every accepted request returns the complete sublevel decomposition and exact measure within the supplied finite scope. Strict and non-strict sublevel sets have the same measure for nonzero boundary polynomials but not the same component membership, so the result must preserve strict semantics rather than silently changing the predicate.

Constant and zero polynomials need explicit exact conventions. Requests whose algebraic-number output cannot fit the admitted representation are rejected, not approximated.

## Bounds and cost class

Bound polynomial degree, coefficient digits, threshold/scope digits, isolation work, number and degree/height of algebraic endpoints, algebraic arithmetic work, and serialized result size. Preflight the degrees of `f-t` and `f+t` and the worst-case boundary count.

## Maintained implementation

Use the maintained exact real-root isolation/sign backend behind the real-algebra operations, plus Jacobian's canonical real-algebraic-number arithmetic. Jacobian owns interval decomposition, strict-boundary conventions, source binding, and exact length reconstruction.

## Defining invariant

- Every finite interior boundary is a root of `f-t` or `f+t`.
- On each open cell determined by the sorted scope endpoints and roots, the sign of `f^2-t^2` is constant.
- The returned contributing cells are exactly those with negative sign.
- The measure is the exact sum of their endpoint differences.

## Evidence plan

- Zero and constant polynomials above, below, and exactly at the threshold.
- `f=x`, `f=x^2`, and polynomials with several alternating components.
- Repeated/tangent roots of `f-t` or `f+t` that do not create a sign change.
- Roots exactly on scope endpoints.
- Strict versus closed membership at equality.
- An irrational algebraic length such as a quadratic boundary, proving the result is not forced to rational or decimal output.
- Reconstruction from the returned interval decomposition.
- Differential comparison with exhaustive sign charts for bounded small integer polynomials.
- Independent mutations of source polynomial, threshold, scope, endpoint, and measure.

Root isolation and sign charts locate the cells, but callers currently cannot add general real-algebraic endpoint differences through one canonical source-bound value at the relevant scale. The exact measure and its reconstructible component decomposition are the missing postcondition; asking every caller to rebuild the complete root/sign/algebraic-sum pipeline is not ordinary lightweight composition.

## Deferred operation family

- Multivariate semialgebraic volume.
- Unbounded scopes without a separate finiteness certificate.
- Approximate numerical measure.
- Optimizing the threshold or polynomial.
- The Erdős 1038 extremal theorem itself.

morluto:
## Summary

Adds `polynomial.real.strict_sublevel_measure.compute`: the complete connected-component decomposition and exact measure of

```text
{ x in [lower, upper] : |f(x)| < threshold }
```

for a bounded canonical univariate polynomial over `QQ`.

SymPy's exact `Poly.real_roots(..., radicals=False)` is the private root-isolation backend. Jacobian combines the roots of `f - threshold` and `f + threshold` into the deterministic strict sign chart, retains every endpoint by equation/index/multiplicity, and replays the source on externally supplied results.

The measure is represented as a rational part plus a canonical signed incidence ledger of retained irrational level roots. This stays exact and composable without forcing a potentially expensive primitive-element expansion. Rational roots are folded into the rational part.

The operation intentionally does not solve a more general semialgebraic measure problem: it is the narrow reusable strict univariate sublevel primitive. Admission bounds degree, input/cleared-level height, root-isolation work, result size, and raw JSON structures before nested normalization.

## Validation

- `make check` (2,669 passed)
- `make test-integration` (1,026 passed)
- Exhaustive degree-one/two differential comparison against SymPy inequalities
- Exact source-replay, strict-JSON transport, rational/irrational boundary, repeated-root, singleton/constant, and adversarial raw-input tests

An independent exact-diff review found and verified the raw exponent-vector admission guard added before JSON normalization.

Closes #2308

### DIFF
--- a/src/jacobian/math/polynomials/real_algebra/_admission.py
+++ b/src/jacobian/math/polynomials/real_algebra/_admission.py
@@ -11,6 +11,11 @@ from jacobian.math.polynomials.real_algebra._tools import TOOLS
 
 ADMISSIONS: tuple[OperationAdmission, ...] = (
     OperationAdmission(
+        "polynomial.real.strict_sublevel_measure.compute",
+        AdmissionDecision.KEEP,
+        "complete exact measure and source-bound component decomposition of a strict polynomial sublevel set",
+    ),
+    OperationAdmission(
         "polynomial.root_count.compute",
         AdmissionDecision.KEEP,
         "distinct exact or explicitly bounded search outcome with material computational leverage",
--- a/src/jacobian/math/polynomials/real_algebra/_operations.py
+++ b/src/jacobian/math/polynomials/real_algebra/_operations.py
@@ -14,6 +14,13 @@ from jacobian.math.polynomials.real_algebra._models import (
     SturmChainResult,
     UnivariatePolynomial,
 )
+from jacobian.math.polynomials.real_algebra._strict_sublevel import (
+    compute_strict_sublevel_payload,
+)
+from jacobian.math.polynomials.real_algebra._strict_sublevel_models import (
+    StrictSublevelMeasureRequest,
+    StrictSublevelMeasureResult,
+)
 
 
 def _poly_to_terms(poly: UnivariatePolynomial) -> list[tuple[Fraction, int]]:
@@ -53,4 +60,22 @@ def compute_root_count(request: RootCountRequest) -> RootCountResult:
         root_count=count,
         lower=request.lower,
         upper=request.upper,
+    )
+
+
+def compute_strict_sublevel_measure(
+    request: StrictSublevelMeasureRequest,
+) -> StrictSublevelMeasureResult:
+    payload = compute_strict_sublevel_payload(request)
+    # The payload is built entirely from the already validated request and the
+    # exact kernel. External result validation still replays the source once;
+    # the trusted producer must not repeat root isolation merely to accept its
+    # own derived fields.
+    return StrictSublevelMeasureResult.model_construct(
+        source_polynomial=request.polynomial,
+        threshold=request.threshold,
+        lower=request.lower,
+        upper=request.upper,
+        components=payload.components,
+        measure=payload.measure,
     )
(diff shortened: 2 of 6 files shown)


morluto:
@codex

Issue #2277, opened 2026-08-22
feat(optimization): accept rational inequalities, bounds, and free variables in exact LPs


morluto:
## Area

Exact rational linear optimization under `jacobian.math.optimization`.

## Problem

`optimization.linear.rational_optimum.compute` currently accepts only standard form:

\[
\min c^Tx\quad\text{subject to}\quad Ax=b,\ x\ge0.
\]

Ordinary exact LP workflows are usually authored with a mixture of equality and inequality rows, finite lower/upper bounds, and free variables. They are mathematically reducible to standard form, but requiring every caller or agent to perform that reduction by hand is error-prone:

- inequality direction can be reversed accidentally;
- surplus/slack signs can be wrong;
- free-variable splitting changes variable identity and dimensions;
- lower-bound shifts introduce objective constants;
- finite upper bounds create additional rows; and
- primal/dual candidates must be translated back before they are intelligible.

This is contract and interoperability friction around the existing exact operation, not a need for another solver or a generic modeling language.

## Proposed public contract

Extend the existing exact rational LP family with a canonical general-form source value. A possible shape is:

```text
variables:
  - name
    lower_bound: rational | null
    upper_bound: rational | null

objective:
  sense: MINIMIZE | MAXIMIZE
  coefficients: exact rational vector

constraints:
  - label
    coefficients: exact rational vector
    relation: LE | EQ | GE
    rhs: exact rational
```

Both bound fields should be explicit:

- `null, null` means a free variable;
- `0, null` recovers the current nonnegative variable;
- finite endpoints give one- or two-sided bounds; and
- equal finite endpoints give a fixed variable.

Reject a finite lower bound greater than the finite upper bound before invoking a backend. Row and variable labels remain canonical source identity, not display-only metadata.

The operation may retain the current ID if this is judged a compatible pre-stable contract expansion. A separately named `general_form` operation is acceptable only if the two source types genuinely establish different stable postconditions; do not expose a second operation merely because the private backend wants standard form.

## Private normalization

The implementation should convert admitted general form to the maintained backend's standard form privately:

- `LE` rows receive nonnegative slack variables;
- `GE` rows are normalized consistently or receive surplus variables;
- a free variable becomes the difference of two nonnegative variables;
- a finite lower bound shifts a variable and updates RHS values and the objective constant;
- finite upper bounds become exact normalized constraints after any shift; and
- maximization may be normalized by exact objective negation.

The converter must retain a request-scoped normalization ledger sufficient to map:

[...]

PR #2382, opened 2026-08-24
Bind rational LP outcomes to their source

### LINKED ISSUE #2276: fix(optimization): bind exact LP outcomes to their source and replay certificates
## Summary

`optimization.linear.rational_optimum.compute` validates candidates produced during its own backend call, but the public `RationalLinearProgramResult` is not bound to the source program. Its model validator checks only which fields accompany each status. It accepts authored candidates, objectives, residuals, and slacks without replaying any of them against \(A,b,c\).

`INFEASIBLE` and `UNBOUNDED` are currently bare statuses with no replayable mathematical certificate.

This leaves the serialized exact result weaker than the operation's claimed postcondition: a result can validate independently of any LP, and a negative conclusion cannot be checked without trusting the backend branch that produced it.

This is a correctness/trust-boundary follow-up to #969 and #1657, not a request for a second LP solver.

## Current behavior

The source program is:

```text
minimize c^T x
subject to A x = b
           x >= 0
```

The operation does replay a SymPy-produced primal and dual before returning. However, `_models.py` currently allows this independently authored result:

```python
RationalLinearProgramResult(
    status="OPTIMAL",
    primal_candidate=(999,),
    primal_objective=-123,
    primal_residuals=(77,),
    dual_candidate=(456,),
    dual_objective=-123,
    dual_slacks=(-88,),
)
```

It validates even though:

- no source program is present;
- the claimed primal residual is nonzero;
- the claimed dual slack is negative; and
- neither objective is derived from a source objective vector.

The same model accepts:

```python
RationalLinearProgramResult(status="INFEASIBLE")
RationalLinearProgramResult(status="UNBOUNDED")
```

with no evidence.

This was reproduced locally with `uv run python` against the current model: all three values serialized successfully. This does not show that the current SymPy adapter returns a wrong answer; it shows that the public exact value does not carry or enforce the evidence needed to justify its own status.

## Required mathematical contracts

Every LP outcome should retain the exact source program and derive its diagnostics from that source. Authored residuals, slacks, and objectives must never be independent authority.

### Optimal

For a minimization problem, return source-bound \(x\) and \(y\) satisfying:

\[
x\ge0,\qquad Ax=b,
\]

\[
A^Ty\le c,
\]

and

\[
c^Tx=b^Ty.
\]

The validator recomputes dimensions, primal residuals, dual slacks, both objective values, and strong-duality equality from the retained source. If residual/slack fields remain public, they are derived values and must equal the recomputation exactly.

### Primal feasible but not certified optimal

A `PRIMAL_FEASIBLE` result retains the source and a point \(x\) with \(x\ge0\) and \(Ax=b\). It establishes feasibility and its exact objective value only. Backend difficulty constructing a dual certificate must not be interpreted as optimality.

### Infeasible

For standard form \(Ax=b, x\ge0\), return a Farkas witness \(y\) under one documented sign convention, for example:

\[
A^Ty\ge0,\qquad b^Ty<0.
\]

The result validator replays both inequalities exactly. A backend exception alone is not the public certificate.

### Unbounded

Return:

- one feasible point \(x_0\ge0\) with \(Ax_0=b\); and
- a recession direction \(d\ge0\) with \(Ad=0\) and \(c^Td<0\).

These establish that \(x_0+td\) remains feasible and the minimization objective tends to \(-\infty\). Replay all conditions exactly.

If the selected backend cannot produce a required negative-status certificate within the admitted domain, narrow the public result or return an explicitly non-mathematical unavailable/incomplete execution outcome. Do not convert missing evidence into `INFEASIBLE` or `UNBOUNDED`.

## Source binding

The preferred result shape retains the canonical `StandardFormRationalLinearProgram`, not only a digest. A digest may additionally protect transport, but it cannot replace the mathematical source needed for downstream replay.

At minimum, the validator must reject:

- the same candidate attached to a different objective;
- a row, column, RHS, or variable-order mutation;
- candidates with the right dimensions but wrong residuals;
- authored objective values inconsistent with source evaluation;
- negative dual slacks for minimization;
- equal authored primal/dual objectives whose candidates are infeasible;
- an infeasibility witness with the wrong Farkas sign; and
- an unbounded ray that is not a recession direction or does not improve the objective.

## Implementation direction

Keep SymPy as a private producer if it can supply the required data. Jacobian owns:

1. the canonical source LP;
2. the source-bound result/certificate types;
3. exact replay of every mathematical conclusion; and
4. the distinction between a mathematical status and backend failure.

It may be useful to factor one domain-owned replay function used by both the operation and model/result construction so the producer path and authored-value path cannot drift. The replay function should compute diagnostics rather than trust submitted residual/slack arrays.

The related general-form input improvement should be handled separately. This issue can first repair the existing standard-form contract.

## Required tests

- A known optimal LP with primal/dual equality replays successfully.
- Mutating one source coefficient, RHS entry, objective coefficient, or variable order rejects a previously valid result.
- Mutating a primal or dual candidate rejects even if submitted objectives still match.
- Nonzero submitted primal residuals and negative submitted dual slacks cannot validate as `OPTIMAL`.
- A feasible point without a valid dual remains only `PRIMAL_FEASIBLE`.
- Valid and corrupted Farkas certificates exercise `INFEASIBLE`.
- Valid and corrupted feasible-point/recession-ray pairs exercise `UNBOUNDED`.
- Dimension mismatches are rejected against the retained source.
- Zero-row, redundant-row, degenerate-optimum, and exact rational boundary cases are covered within the admitted domain.
- Serialization round trips preserve source binding and replay.

## Acceptance criteria

- [ ] Every mathematical LP status is bound to the exact source program.
- [ ] Primal/dual objectives, residuals, and slacks are recomputed exactly from source and candidates.
- [ ] `INFEASIBLE` carries a replayable exact certificate.
- [ ] `UNBOUNDED` carries a replayable feasible point and improving recession ray.
- [ ] Missing backend evidence cannot become a mathematical negative conclusion.
- [ ] Source/candidate mutation tests fail closed.
- [ ] The fix preserves the existing operation ID where its public claim remains supportable.

## Related

- #969 — original exact rational LP certificate request, closed after the current operation was added.
- #1657 — exact optimality verification and source/objective binding.

morluto:
## Summary

Repairs the public contract of `optimization.linear.rational_optimum.compute` for its existing standard-form domain:

```text
minimize c^T x  subject to A x = b, x >= 0
```

Every mathematical result now retains the exact source program and replays its claim:

- `OPTIMAL` carries exact primal and dual points with derived residuals, slacks, objectives, and strong-duality equality.
- `PRIMAL_FEASIBLE` establishes only a replayed feasible point when no dual proof is available.
- `INFEASIBLE` carries a Farkas witness using `A^T y >= 0, b^T y < 0`.
- `UNBOUNDED` carries a replayed feasible point and nonnegative recession direction with `Ad = 0, c^T d < 0`.
- `UNKNOWN` carries no mathematical conclusion.

SymPy remains a private exact simplex producer. Jacobian owns canonical source binding, certificate types, replay, status semantics, and explicit work/result admission. Redundant `0 = 0` source rows are pruned from auxiliary dual/Farkas tableaux but restored as zero coordinates in public certificates, so the stated tableau envelope matches actual backend work.

Raw source input is limited to 128-digit rationals before generic parsing. Raw result transport uses the generic canonical pre-cap, then the result’s source-derived rational-height bound is replayed; this preserves legitimate large exact coordinates without admitting unbounded caller input.

This deliberately does not add general inequalities, bounds, or free variables; that representation extension remains #2277.

## Validation

- `make check` (2,649 passed)
- `make test-integration` (1,040 passed)
- Source/candidate/objective/matrix/order mutations, primal-only fallback, Farkas and recession negative controls, zero/redundant rows, strict JSON round trips, pre-parse limits, and large source-derived-coordinate regressions

An independent exact-diff review found the zero-row work-accounting and raw-bound gaps; both fixes were followed by focused, full, and integration validation.

Closes #2276

### DIFF
--- a/src/jacobian/math/optimization/_admission.py
+++ b/src/jacobian/math/optimization/_admission.py
@@ -13,7 +13,7 @@ ADMISSIONS: tuple[OperationAdmission, ...] = (
     OperationAdmission(
         "optimization.linear.rational_optimum.compute",
         AdmissionDecision.KEEP,
-        "distinct exact or explicitly bounded search outcome with material computational leverage",
+        "distinct exact source-bound standard-form LP outcome with replayable optimality, Farkas, or recession witnesses",
     ),
 )
 
--- a/src/jacobian/math/optimization/_models.py
+++ b/src/jacobian/math/optimization/_models.py
@@ -2,32 +2,388 @@
 
 from __future__ import annotations
 
+from collections.abc import Mapping
+from fractions import Fraction
+from math import comb, factorial
 from typing import Literal, Self
 
 from pydantic import Field, model_validator
 
-from jacobian._exact import CanonicalRational, require_bounded_rational
+from jacobian._exact import (
+    MAX_CANONICAL_RATIONAL_DIGITS,
+    CanonicalRational,
+    require_bounded_rational,
+)
 from jacobian._models import StrictModel
 
 MAX_RATIONAL_DIGITS = 128
+MAX_LINEAR_PROGRAM_RESULT_BYTES = 10 * 1024 * 1024
+MAX_LINEAR_PROGRAM_SIMPLEX_BASES = 1_000_000
+MAX_LINEAR_PROGRAM_SIMPLEX_SCALAR_UPDATES = 50_000_000
 type RationalLinearProgramStatus = Literal[
     "OPTIMAL",
     "PRIMAL_FEASIBLE",
     "INFEASIBLE",
     "UNBOUNDED",
+    "UNKNOWN",
 ]
 
 
+def _bound_raw_rational(
+    value: object,
+    *,
+    maximum_digits: int = MAX_RATIONAL_DIGITS,
+    label: str,
+) -> None:
+    """Apply the LP scalar limit before generic rational construction."""
+
+    if isinstance(value, CanonicalRational):
+        components: tuple[object, object] = (value.num, value.den)
+    elif isinstance(value, Mapping):
+        components = (value.get("num"), value.get("den"))
+    else:
+        return
+    for component in components:
+        if (
+            isinstance(component, str)
+            and len(component) - component.startswith("-") > maximum_digits
+        ):
+            raise ValueError(f"{label} exceeds the {maximum_digits}-digit bound")
+
+
+def _bound_raw_rational_vector(
+    value: object,
+    *,
+    maximum_length: int,
+    maximum_digits: int = MAX_RATIONAL_DIGITS,
+    label: str,
+) -> None:
+    """Bound one raw LP vector before recursive model construction."""
+
+    if not isinstance(value, (list, tuple)):
+        return
+    if len(value) > maximum_length:
+        raise ValueError(f"{label} exceeds the {maximum_length}-entry bound")
+    for item in value:
+        _bound_raw_rational(item, maximum_digits=maximum_digits, label=label)
+
+
+def _prepare_raw_rational_vector(
+    value: object,
+    *,
+    maximum_length: int,
+    maximum_digits: int = MAX_RATIONAL_DIGITS,
+    label: str,
+) -> object:
+    """Bound and normalize one raw LP vector before nested parsing."""
+
+    _bound_raw_rational_vector(
+        value,
+        maximum_length=maximum_length,
+        maximum_digits=maximum_digits,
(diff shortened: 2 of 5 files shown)


morluto:
@codex

PR #2653, opened 2026-08-25
fix(mcp): harden cancellation, statelessness, and releases

### LINKED ISSUE #2646: [Bug]: Packaged MCP execution is not exercised by release checks, allowing deployed find/run timeouts
**Priority**

P1 (high): a deployed MCP endpoint can pass the current source, wheel, container, and discovery-oriented checks while ordinary agent calls are unusable.

**Area**

ci, mcp, deployment

**Component**

Built-wheel/container validation, `deploy/smoke_remote.py`, and release/deployment smoke coverage.

**Observed behavior**

On 2026-08-25, the installed Jacobian connector available to an agent returned only `TimeoutError` for all four of these minimal calls:

```json
{"request":{"op":"search","query":"greatest common divisor of two integers","domain":"integer","limit":5}}
```

```json
{"request":{"op":"browse","domain":"integer","limit":1}}
```

```json
{"request":{"op":"inspect","operation_id":"integer.compute.extended_gcd"}}
```

```json
{"operation_id":"integer.compute.extended_gcd","payload":{"left":"84","right":"30"}}
```

This is not evidence of an expensive mathematical backend: search, browse, inspect, and a trivial exact GCD all fail at the deployed tool boundary.

The current checks leave this deployment class uncovered:

- `tests/mcp/test_mcp_invocation_journey.py` exercises `Client(create_server())` directly from the source checkout, including a successful inspect -> extended-GCD journey.
- The wheel job in `.github/workflows/ci.yml` installs the wheel and runs the direct CLI `jacobian run integer.compute.extended_gcd`, but never starts the installed `jacobian-mcp`/`jacobian-remote-mcp` executable and invokes it through MCP.
- `.github/workflows/container-image.yml` likewise checks direct CLI execution inside the image, not the MCP server path.
- `deploy/smoke_remote.py` checks handshake, tool listing, catalog, and one `math.find` search, but never calls `math.run`.
- Tests for the deployment smoke helper cover error classification and exit codes, not an end-to-end remote find/run journey.

Consequently, packaging, startup, transport, connector, or deployed-runtime regressions can escape while direct kernel/CLI checks remain green.

**Expected behavior**

Release and deployment gates should start the exact built artifact through its production MCP entry point and prove a bounded read-only journey such as:

1. handshake and list tools;
2. `math.find` search;
3. `math.find` inspect for `integer.compute.extended_gcd`;
4. `math.run` for `84,30`, requiring gcd `6`;
5. a second small call to verify the server remains responsive.

Each phase should have a short, independent deadline and report which phase exceeded it. The remote deployment smoke should include the same pure `math.run` canary, not only discovery. A test double where handshake/discovery works but `math.run` hangs should prove the smoke fails.

The check should run against:

- the built wheel's installed executable, not imports from the checkout; and
- the built container's configured server entry point, not only `jacobian run`.

**Failure stage and owner**

- [ ] request parsing
- [ ] backend or kernel execution
- [ ] result construction
- [x] transport projection
- Owner: packaged MCP/release validation and deployment smoke

**Typed outcome**

The release/deployment check should fail with a phase-specific operational timeout such as `MCP_SMOKE_TIMEOUT` / `MCP_RUN_TIMEOUT`, including the failed operation and elapsed phase, without treating the timeout as a mathematical result.

**Steps to reproduce**

1. Start the currently deployed/installed Jacobian MCP connector.
2. Send any of the four calls above.
3. Observe `TimeoutError`, including for catalog-only discovery and the trivial extended GCD.
4. Compare with current release checks: they can validate source-level MCP or direct CLI paths without invoking the packaged MCP server's `math.run` path.

**Environment**

- Source package version at that revision: `0.13.0`
- Deployed connector version: unavailable because all discovery/inspection calls timed out
- Date reproduced: 2026-08-25
- Python/OS of deployed connector: not exposed

**Additional context**

This is a concrete release-gate and smoke-coverage defect, not a request to redesign every operation's execution budget. The broader deadline architecture issue #1145 was closed as not planned; this issue asks for a small end-to-end packaged canary that would catch the currently observed outage class.

### LINKED ISSUE #2647: [Bug]: jacobian-remote-mcp defaults to stateful HTTP despite its stateless contract and deployment guide
**Priority**

P2 (medium): the production entry point starts a different Streamable HTTP lifecycle from the one its module, CLI, and deployment documentation claim.

**Area**

mcp, docs, deployment

**Component**

`src/jacobian/mcp/remote_cli.py` and `docs/how-to/deploy-remote-mcp.md`

**Observed behavior**

The remote implementation is repeatedly presented as stateless:

- `src/jacobian/mcp/remote.py` describes “stateless serving” and `create_remote_server()` as a “stateless remote host”;
- `src/jacobian/mcp/remote_cli.py` describes the command as “Run the stateless remote Jacobian MCP host”;
- the deployment guide presents the service as one stateless immutable service artifact.

However, the CLI defines:

```python
parser.add_argument(
    "--stateless-http",
    action="store_true",
    help="use stateless Streamable HTTP sessions",
)
```

so the default value is `False`. The Streamable HTTP launch then passes:

```python
stateless_http=args.stateless_http
```

The documented production command omits `--stateless-http`, therefore it launches the SDK with `stateless_http=False` despite the declared stateless contract.

Repository search finds no test that asserts the intended default or verifies the value passed from the CLI to `server.run()`.

**Expected behavior**

Choose and enforce one truthful lifecycle:

- If the remote service is intended to be stateless, make stateless Streamable HTTP the default and provide an explicit, clearly named opt-in for stateful sessions only if they are supported.
- If stateful sessions are intended by default, remove the stateless claims and document their lifecycle, affinity, restart, and load-balancing requirements.

Given the current module/API contract, the first option appears consistent: the documented command should pass `stateless_http=True` without requiring an extra flag.

Add a CLI test that captures `server.run()` arguments for the default command and for any explicit override. The deployment smoke/report should also expose the selected session mode so operators can verify it.

**Failure stage and owner**

- [x] request parsing
- [ ] backend or kernel execution
- [ ] result construction
- [x] transport projection
- Owner: remote MCP CLI/configuration contract

**Typed outcome**

This is deployment configuration rather than a mathematical outcome. The server should start in the declared session mode, or fail startup with an explicit configuration error if a requested mode is unsupported.

**Steps to reproduce**

1. Run the documented command from `docs/how-to/deploy-remote-mcp.md` without `--stateless-http`.
2. Inspect `argparse` output or intercept the call to `server.run()`.
3. Observe `stateless_http=False`.
4. Compare this with the remote module, CLI description, and deployment documentation, all of which declare a stateless host.

**Environment**

- Package version: `0.13.0`
- Date inspected: 2026-08-25

**Additional context**

This mismatch was found while investigating deployed MCP timeouts. It is not asserted here to be the sole cause of those timeouts; it is an independent transport-lifecycle defect that can affect connector behavior, routing, restart semantics, and reproducibility.

### LINKED ISSUE #2648: [Bug]: MCP request cancellation is not connected to bounded subprocess termination
**Priority**

P1 (high): cancelled/timed-out MCP requests can leave backend work running until an operation-specific timeout, consuming workers and process resources after the caller has abandoned the result.

**Area**

mcp, process

**Component**

`src/jacobian/mcp/tools.py`, `src/jacobian/dispatch.py`, `src/jacobian/mcp/runtime.py`, and `src/jacobian/process.py`

**Observed behavior**

The current MCP execution path is synchronous and has no request-cancellation bridge:

```text
math_run
  -> invoke_operation
     -> operation.run(parsed)
```

`math_run()` directly calls `invoke_operation()`. `AppState` contains the catalog and authorization callback but no request-scoped deadline or cancellation event. `invoke_operation()` directly calls the selected operation's `run()` method.

Jacobian does have process cancellation machinery:

- `bounded_process_cancellation(event)` installs a context-local cancellation event;
- `bounded_process_cancelled()` exposes it to operation code;
- `run_bounded_process(..., cancellation_event=...)` polls the event and kills/reaps the owned process tree.

However, repository search at the audit baseline finds no production call site for `bounded_process_cancellation`, no production caller supplying `cancellation_event=`, and no operation polling `bounded_process_cancelled()`. Those names occur only in the process implementation and tests/tooling.

Real mathematical backends do call `run_bounded_process`, including the logic solver, Singular adapters, multivariate factorization, and polynomial-map replay paths. Because the MCP request lifecycle never installs or passes an event, their process monitor cannot learn that the MCP client cancelled or timed out. At best they stop at their separate operation/process timeout; client cancellation itself is not propagated.

The previous architecture issue #1077 explicitly described a request `threading.Event`, cancellation drain, and late-worker ownership. It was closed as not planned on 2026-08-15. The current simplified code no longer contains the bridge described there, making this a concrete present-code regression rather than a request for the full architecture proposed in #1077.

**Expected behavior**

Bind MCP request cancellation to the existing kill-safe subprocess mechanism:

1. create one request-scoped cancellation event before dispatch;
2. install it through `bounded_process_cancellation(...)` and/or pass it explicitly to every bounded subprocess owner;
3. set the event when the MCP request is cancelled or its transport deadline expires;
4. wait for a bounded drain interval;
5. verify the owned process tree is terminated and reaped;
6. return/record an operational cancellation outcome only—never a mathematical negative conclusion.

For in-process work, either add tested cooperative checkpoints or explicitly declare it non-interruptible and bound admission accordingly. Do not imply that cancelling the request coroutine stops Python/native work when it does not.

Add an end-to-end MCP regression test with a synthetic long-running bounded child process:

- start the operation through the real MCP transport;
- cancel the client call;
- assert the child/process group exits within the declared bound;
- assert no result is delivered to another request;
- immediately invoke `integer.compute.extended_gcd` and require a prompt successful response;
- repeat cancellations to prove workers/processes do not accumulate.

**Failure stage and owner**

- [ ] request parsing
- [x] backend or kernel execution
- [ ] result construction
- [x] transport projection
- Owner: MCP execution lifecycle and bounded-process cancellation bridge

**Typed outcome**

Cancellation must remain an operational non-conclusion, for example a protocol cancellation or a bounded `CANCELLED`/`DEADLINE_EXCEEDED` diagnostic. It must not produce `false`, `no witness`, `infeasible`, or any other mathematical claim.

**Steps to reproduce**

1. Register or select an operation that launches a long-running command through `run_bounded_process`.
2. Invoke it through the MCP server.
3. Cancel the MCP client request before the child finishes.
4. Observe that no Jacobian production path sets a cancellation event for `run_bounded_process`; the child can continue until its own timeout/completion.
5. Inspect `math_run`, `AppState`, and `invoke_operation` to confirm no request-scoped cancellation object is created or propagated.

**Environment**

- Package version: `0.13.0`
- Date inspected: 2026-08-25

**Additional context**

The installed connector also timed out on search, browse, inspect, and trivial GCD calls during this investigation; that deployment/gating symptom is tracked separately in #2646. This issue is specifically about the statically demonstrable missing cancellation link for work that has already begun.

kaoru0822-kitauji:
## Summary

- connect the MCP SDK's per-request cancellation signal to Jacobian's bounded subprocess ownership, with a bounded graceful process-tree termination phase followed by forced termination and reaping
- make `jacobian-remote-mcp` Streamable HTTP stateless by default while retaining stateful sessions only behind the explicit `--stateful-http` opt-in
- install and exercise the actual wheel, sdist, npm tarball, and container entry points over the real MCP protocol in CI/release checks, with named startup/request/shutdown deadlines and diagnostic stderr tails

These changes remain one narrow production-hardening PR because all three issues meet at the MCP delivery boundary: request lifetime, remote transport lifetime, and the release evidence for those entry points. Mathematical operation semantics are unchanged.

## Reproduction before the fix

- `remote_cli._parser().parse_args([])` produced `transport=streamable-http` with `stateless_http=False`, despite the deployment contract describing a stateless service.
- A real stdio MCP request to a synthetic process-backed operation was cancelled by the client after about 0.4 seconds and returned MCP error `-32001`, but the spawned child PID was still alive. A subsequent `math.run` remained usable, showing that request cancellation and worker ownership—not the server loop—were disconnected.
- The focused baseline suites passed (`8` MCP tests; `16` process tests plus one platform skip) because there was no regression coverage for request cancellation before/during spawn, TERM escalation, descendant cleanup, or installed-artifact protocol execution.
- CI installed only the wheel and exercised help/direct CLI paths; the sdist was not installed, the npm carrier only had mocked forwarding tests, container checks bypassed its configured MCP entry point, and the release workflow uploaded exact artifacts without starting them over MCP.

### Request cancellation and bounded process cleanup

`math.run` is now async at the MCP boundary. It relays the pinned MCP 2.0 request cancellation signal into a request-owned thread event while the synchronous typed operation runs in a non-abandoned worker thread. Process-backed operations consume that event through the existing cancellation context.

The bounded process supervisor now:

- refuses to spawn when cancellation is already set
- distinguishes `cancelled`, `timed_out`, and ordinary nonzero execution failure
- sends graceful termination to the complete owned POSIX process group, waits at most 0.5 seconds, escalates to `SIGKILL`, and bounds the forced-reap phase to another 0.5 seconds
- uses bounded `taskkill /T` then `/T /F` phases when the trusted Windows tool is available, with a root-process fallback
- joins the cancellation relay and worker cleanup rather than abandoning a task or swallowing cancellation

The POSIX process-tree contract is exercised with parent and child PIDs, including a TERM-ignoring tree and three repeated real MCP cancellations followed by a successful `math.run`.

### Stateless remote default

Streamable HTTP now starts with `stateless_http=True` by default. `--stateless-http` remains an accepted explicit spelling, while `--stateful-http` is the only stateful opt-in. The flags are mutually exclusive, startup logs identify the selected session mode, and systemd/deployment examples exercise the default rather than redundantly forcing it. Authentication CLI/environment precedence is unchanged; there is intentionally no environment or config-file session-mode layer.

The remote smoke records whether any response supplied `mcp-session-id` and fails if that observed mode differs from `--expect-session-mode` (stateless by default).

### Distributable MCP evidence

`deploy/smoke_stdio.py` imports no Jacobian source modules, removes `PYTHONPATH`, sets an isolated working directory, captures server stderr, and drives initialization, exact tool discovery, `math.find` search and inspection, and two bounded `math.run` requests over stdio. Every phase has an explicit timeout and names itself in diagnostics.

CI now covers:

- wheel and sdist installs on CPython 3.12 and 3.13
- the packed npm carrier on Node 18 and 24, resolving its Python server from the exact built wheel
- the built PR container through its configured entry point
- the digest-addressed published container before publication evidence is recorded

The release workflow repeats the exact wheel, sdist, and npm tarball protocol smoke before uploading distributions. The remote/container smoke now performs the same discovery/inspection/execution journey and verifies statelessness.

## Validation and packaged-artifact evidence

Focused final-tree checks:

- `make test-mcp TESTS='tests/mcp/test_mcp_cancellation.py tests/mcp/test_mcp_sdk_2_conformance.py tests/mcp/test_remote_mcp_auth.py'` — 11 passed
- `make test-process TESTS='tests/process/test_bounded_process.py tests/process/tooling/test_deploy_helpers.py'` — 24 passed, 1 skipped locally because macOS has no `prlimit`
- `make test-tooling TESTS='tests/tooling/test_product_ci_contract.py tests/tooling/test_release_workflow.py tests/tooling/test_ci_test_plan.py'` — 32 passed
- a live default `jacobian-remote-mcp` plus `python -m deploy.smoke_remote` — stateless mode observed; exact two-tool surface, catalog, search, inspect, gcd 6, and second gcd 7 all passed

Required/static checks:

- `make check-static` — passed (lint-full, dependency/dead-code checks, mypy, import contracts, architecture, build, wheel contents)
- `make docs-linkcheck` — passed
- workflow YAML parse and `git diff --check` — passed
- `make check` — lint and mypy passed; the test phase reported 6166 passed and 60 skipped. Seventeen unrelated multivariate-factor tests fail on macOS because the worker correctly refuses to run without the hard address-space limit supplied by Linux `prlimit`. One additional namespace failure came from a stale ignored `__pycache__` left by an older checkout; after removing that cache, the exact namespace test passed. No mathematical or resource-bound behavior was weakened to mask either local condition.

From freshly rebuilt final-tree artifacts in isolated CPython 3.12 environments, the wheel, sdist, and installed npm tarball each passed the complete MCP journey and reported:

```text
wheel  -> tools [math.find, math.run], gcd runs [6, 7]
sdist  -> tools [math.find, math.run], gcd runs [6, 7]
npm    -> tools [math.find, math.run], gcd runs [6, 7]
```

The same wheel/sdist journey was also characterized locally on CPython 3.13 before the final upstream rebase; the final-tree 3.12/3.13 and Node 18/24 matrices are mandatory CI jobs. Docker is installed locally but its daemon is unavailable, so the container route could not be executed on this host; the PR and publish container workflows now make that protocol smoke mandatory on Linux.

Fixes #2646
Fixes #2647
Fixes #2648

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,11 @@ jobs:
       - run: make docs-linkcheck
       - run: make npm-test
       - run: make build
+      - name: Pack npm distribution
+        working-directory: npm
+        run: |
+          mkdir -p ../dist/npm
+          npm pack --pack-destination ../dist/npm
       - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: dist
@@ -246,7 +251,7 @@ jobs:
       - run: make test-singular
 
   wheel:
-    name: wheel (Python ${{ matrix.python-version }})
+    name: ${{ matrix.artifact }} (Python ${{ matrix.python-version }})
     needs: [static]
     runs-on: ubuntu-latest
     timeout-minutes: 10
@@ -254,6 +259,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.12", "3.13"]
+        artifact: [wheel, sdist]
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
@@ -268,26 +274,94 @@ jobs:
         with:
           name: dist
           path: dist
-      - name: Install and start the built wheel
+      - name: Install and smoke the built Python artifact over MCP
         run: |
-          wheel="$(find dist -maxdepth 1 -name '*.whl' -print -quit)"
-          test -n "$wheel"
-          environment="$RUNNER_TEMP/jacobian-wheel-${{ matrix.python-version }}"
+          case "${{ matrix.artifact }}" in
+            wheel)
+              package="$(find dist -maxdepth 1 -name '*.whl' -print -quit)"
+              install_policy=(--only-binary :all:)
+              ;;
+            sdist)
+              package="$(find dist -maxdepth 1 -name '*.tar.gz' -print -quit)"
+              install_policy=()
+              ;;
+          esac
+          test -n "$package"
+          package="$(realpath "$package")"
+          environment="$RUNNER_TEMP/jacobian-${{ matrix.artifact }}-${{ matrix.python-version }}"
           uv venv --python "${{ matrix.python-version }}" "$environment"
           uv pip install --python "$environment/bin/python" \
-            --only-binary :all: "$wheel"
-          "$environment/bin/jacobian" --help
-          "$environment/bin/jacobian-mcp" --help
-          "$environment/bin/jacobian" run integer.compute.extended_gcd \
-            --json '{"left":"84","right":"30"}'
-      - if: matrix.python-version == '3.13'
+            "${install_policy[@]}" "$package"
+          version="$("$environment/bin/python" -c \
+            'import importlib.metadata; print(importlib.metadata.version("jacobian"))')"
+          PYTHONPATH= "$environment/bin/python" "$GITHUB_WORKSPACE/deploy/smoke_stdio.py" \
+            --expect-version "$version" \
+            --startup-timeout-seconds 60 \
+            --request-timeout-seconds 20 \
+            --shutdown-timeout-seconds 10 \
+            -- "$environment/bin/jacobian-mcp"
+      - if: matrix.python-version == '3.13' && matrix.artifact == 'wheel'
         uses: ./.github/actions/setup-python-tests
         with:
           python-version: "3.13"
-      - if: matrix.python-version == '3.13'
+      - if: matrix.python-version == '3.13' && matrix.artifact == 'wheel'
         run: make test-compatibility
         env:
           PYTEST_ARGS: --junitxml=pytest.xml
+
+  npm-package:
+    name: npm tarball (Node ${{ matrix.node-version }})
+    needs: [static]
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    strategy:
+      fail-fast: false
+      matrix:
+        node-version: [18, 24]
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
+        with:
+          python-version: "3.12"
+      - uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
+        with:
+          version: "0.11.28"
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: ${{ matrix.node-version }}
+          package-manager-cache: false
+      - uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: dist
+          path: dist
+      - name: Install and smoke the packed npm carrier over MCP
+        run: |
+          wheel="$(find dist -maxdepth 1 -name '*.whl' -print -quit)"
+          tarball="$(find dist/npm -maxdepth 1 -name '*.tgz' -print -quit)"
+          test -n "$wheel"
+          test -n "$tarball"
+          wheel="$(realpath "$wheel")"
+          tarball="$(realpath "$tarball")"
+          client_environment="$RUNNER_TEMP/jacobian-npm-client"
+          project="$RUNNER_TEMP/jacobian-npm-project"
+          uv venv --python 3.12 "$client_environment"
+          uv pip install --python "$client_environment/bin/python" \
+            --only-binary :all: "$wheel"
+          mkdir -p "$project"
+          npm install --prefix "$project" "$tarball"
+          version="$("$client_environment/bin/python" -c \
+            'import importlib.metadata; print(importlib.metadata.version("jacobian"))')"
+          JACOBIAN_PACKAGE="$wheel" \
+          JACOBIAN_UV_BIN="$(command -v uvx)" \
+          PYTHONPATH= \
+            "$client_environment/bin/python" "$GITHUB_WORKSPACE/deploy/smoke_stdio.py" \
+              --expect-version "$version" \
+              --startup-timeout-seconds 60 \
+              --request-timeout-seconds 20 \
+              --shutdown-timeout-seconds 10 \
+              -- "$project/node_modules/.bin/jacobian" mcp
 
   lean:
     name: Lean Runtime
@@ -362,11 +436,11 @@ jobs:
   required:
     name: required
     if: ${{ always() }}
-    needs: [plan, static, math, catalog, catalog_examples, python, boundaries, singular, wheel, coverage, lean]
+    needs: [plan, static, math, catalog, catalog_examples, python, boundaries, singular, wheel, npm-package, coverage, lean]
     runs-on: ubuntu-latest
     timeout-minutes: 5
     steps:
-      - name: Require selected evidence and final-tree specialist lanes
+      - name: Require selected evidence packages and final-tree specialist lanes
         env:
           # True for pull_request runs and for autofix dispatches that select
           # the pull_request lane set; Lean then counts skipped as passing.
@@ -383,6 +457,7 @@ jobs:
           BOUNDARIES_RESULT: ${{ needs.boundaries.result }}
           SINGULAR_RESULT: ${{ needs.singular.result }}
           WHEEL_RESULT: ${{ needs.wheel.result }}
+          NPM_PACKAGE_RESULT: ${{ needs.npm-package.result }}
           COVERAGE_RESULT: ${{ needs.coverage.result }}
           LEAN_RESULT: ${{ needs.lean.result }}
         run: |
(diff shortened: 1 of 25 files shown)


morluto:
Closing this conflicted, MCP-2.0-based combined PR as superseded. Its useful outcomes will be rebuilt from current main as focused changes: (1) stateless remote HTTP default, (2) packaged-artifact MCP protocol smoke coverage, and (3) MCP 2.1 cancellation propagation into the shared bounded-process runner. The branch and tests remain available as reference; no history rewrite is planned.

PR #2656, opened 2026-08-25
fix(mcp): default remote HTTP to stateless

### LINKED ISSUE #2647: [Bug]: jacobian-remote-mcp defaults to stateful HTTP despite its stateless contract and deployment guide

morluto:
Fixes #2647.

### Problem

`jacobian-remote-mcp` described its Streamable HTTP host as stateless, but the documented command started it with `stateless_http=False` unless callers supplied an extra flag.

### Change

Streamable HTTP is now stateless by default. `--stateful-http` is the explicit opt-in; the existing `--stateless-http` spelling remains accepted. The service examples now exercise the default mode, and the deployment guide documents the lifecycle choice.

### Regression coverage

The MCP CLI test observes the argument passed to the server for both the default and explicit stateful modes.

### Validation

- `make test-mcp TESTS=tests/mcp/test_remote_mcp_auth.py` — 4 passed
- `make docs-linkcheck` — passed
- `uv run --locked ruff format --check src/jacobian/mcp/remote_cli.py tests/mcp/test_remote_mcp_auth.py` — passed

### DIFF
--- a/.github/workflows/heldout-benchmarks.yml
+++ b/.github/workflows/heldout-benchmarks.yml
@@ -87,7 +87,7 @@ jobs:
           docker run --rm -d --name jacobian-heldout-probe \
             -p 127.0.0.1:8000:8000 "$image" \
             --transport streamable-http --host 0.0.0.0 --port 8000 \
-            --allow-anonymous --stateless-http
+            --allow-anonymous
           echo "url=http://127.0.0.1:8000/mcp" >> "$GITHUB_OUTPUT"
       - name: Execute and resume at complete pair boundaries
         env:
--- a/deploy/systemd/jacobian-mcp-anonymous.conf
+++ b/deploy/systemd/jacobian-mcp-anonymous.conf
@@ -11,6 +11,5 @@ ExecStart=/usr/bin/env /opt/jacobian/current/.venv/bin/jacobian-remote-mcp \
   --host 127.0.0.1 \
   --port 8765 \
   --path /mcp \
-  --stateless-http \
   --allow-anonymous \
   --anonymous-tenant-id replace-with-unique-test-id
--- a/deploy/systemd/jacobian-mcp.service
+++ b/deploy/systemd/jacobian-mcp.service
@@ -15,7 +15,6 @@ ExecStart=/usr/bin/env /opt/jacobian/current/.venv/bin/jacobian-remote-mcp \
   --host 127.0.0.1 \
   --port 8765 \
   --path /mcp \
-  --stateless-http \
   --auth-tokens-file %d/tokens.json \
   --public-base-url https://math-tools.example.org
 Restart=on-failure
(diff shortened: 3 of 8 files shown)


morluto:
Credit to @kaoru0822-kitauji for the original combined implementation and investigation in #2653. This focused PR carries forward its stateless-HTTP outcome.

PR #2661, opened 2026-08-25
fix(mcp): cancel bounded subprocess trees

### LINKED ISSUE #2648: [Bug]: MCP request cancellation is not connected to bounded subprocess termination

morluto:
Fixes #2648.

## Problem

When an MCP client abandons `math.run`, the SDK records the request cancellation, but Jacobian did not bind that signal to its bounded subprocess runner. An external backend could therefore continue after its client had stopped waiting.

## Solution

Keep `math.run` synchronous. MCP dispatches synchronous tools in a worker thread, where the request-scoped SDK cancellation event is bound once to the existing common process runner. The runner polls that signal and kills and reaps the owned process tree. In-process operations remain synchronous and unchanged.

MCP 2.1 does not expose cancellation on its high-level tool `Context`; the bridge is contained in one helper that reads the current request dispatcher event through the SDK's available session seam. It can be replaced when the SDK exposes that event publicly.

## Testing

- `make test-mcp TESTS='tests/mcp/test_mcp_cancellation.py tests/mcp/test_mcp_sdk_2_conformance.py'` — 3 passed. The new stdio regression cancels three process trees, verifies both parent and child exit, then runs `integer.compute.extended_gcd` on the same server.
- `uv run --locked ruff check src/jacobian/mcp/tools.py src/jacobian/process.py tests/mcp/_cancellation_server.py tests/mcp/test_mcp_cancellation.py` — passed.
- `uv run --locked ruff format --check src/jacobian/mcp/tools.py src/jacobian/process.py tests/mcp/_cancellation_server.py tests/mcp/test_mcp_cancellation.py` — passed.
- `make check` completed its lint and typecheck stages; the full math/catalog test command was started but its terminal completion status was not captured, so it is intentionally not claimed as passing here.

## Public contract impact

None. `math.run` remains synchronous and operation request/result contracts are unchanged.

## Operation boundary ownership

- Changed stage: transport cancellation into external-process execution control.
- Request-bounds owner and controlling quantities: unchanged.
- Backend or kernel path: unchanged; only existing subprocess-backed operations observe the request signal.
- Result construction: unchanged.
- Independent-result verifier: none.
- Native/MCP parity: native calls are unchanged; MCP adds transport-only cancellation for request-owned external child processes.
- Serialized-result and round-trip evidence: not applicable.

## Checklist

- [ ] `make check` passes
- [x] New shared abstraction replaces duplicate cancellation bridging across external backends

### DIFF
--- a/src/jacobian/mcp/tools.py
+++ b/src/jacobian/mcp/tools.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 from collections.abc import Mapping, Sequence
 from typing import Any
 
+import anyio
 from mcp.server.mcpserver import Context
 from mcp.server.mcpserver.exceptions import ToolError
 from mcp.shared.exceptions import MCPError
@@ -30,6 +31,7 @@ from jacobian.mcp.runtime import (
     _authorize,
     _catalog,
 )
+from jacobian.process import bounded_process_cancellation
 
 _MAX_VALIDATION_ERRORS = 64
 _MAX_VALIDATION_LOCATION_COMPONENTS = 32
@@ -93,11 +95,11 @@ def math_run(
     if catalog.operation(operation_id) is None:
         raise ToolError(f"unknown operation: {operation_id}")
     try:
-        return invoke_operation(
-            operation_id,
-            payload,
-            catalog,
-        )
+        # MCP runs synchronous tools in a worker thread.  Its request event
+        # is polled by the shared external-process runner, which kills and
+        # reaps only an operation's owned child tree.
+        with bounded_process_cancellation(_request_cancellation(ctx)):
+            return invoke_operation(operation_id, payload, catalog)
     except OperationRequestValidationError as exc:
         errors = _bounded_validation_issues(exc.errors())
         data = OperationInvalidRequestData(
@@ -109,6 +111,12 @@ def math_run(
             message="operation payload failed validation",
             data=data.model_dump(mode="json"),
         ) from exc
+
+
+def _request_cancellation(ctx: Context[AppState, Any]) -> anyio.Event:
+    """Return MCP 2.1's request signal through its only available SDK seam."""
+
+    return ctx.request_context.session._request_outbound.cancel_requested
 
 
 def _bounded_validation_issues(
--- a/src/jacobian/mcp/tools.py
+++ b/src/jacobian/mcp/tools.py
@@ -3,9 +3,8 @@
 from __future__ import annotations
 
 from collections.abc import Mapping, Sequence
-from typing import Any
+from typing import Any, Protocol
 
-import anyio
 from mcp.server.mcpserver import Context
 from mcp.server.mcpserver.exceptions import ToolError
 from mcp.shared.exceptions import MCPError
@@ -37,6 +36,12 @@ _MAX_VALIDATION_ERRORS = 64
 _MAX_VALIDATION_LOCATION_COMPONENTS = 32
 _MAX_VALIDATION_LOCATION_LENGTH = 128
 _MAX_VALIDATION_ISSUES_BYTES = 48 * 1_024
+
+
+class _CancellationSignal(Protocol):
+    """Cooperative request cancellation observed by external process work."""
+
+    def is_set(self) -> bool: ...
 
 
 def math_find(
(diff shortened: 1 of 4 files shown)


morluto:
Credit to @kaoru0822-kitauji for the original cancellation investigation and the combined #2653 implementation that identified the process-lifecycle gap. This focused follow-up keeps that goal while using one synchronous MCP-edge bridge into the existing shared process runner.

Issue #2778, opened 2026-08-26
[Discovery] Make incidence containment profiles discoverable as t-codegree computations


morluto:
## Problem

`incidence.containment_profiles.compute` returns the map from every `t`-subset of points to the number of blocks containing it. In standard hypergraph and design vocabulary, this is the full `t`-codegree profile. Its current title, description, and tags use only “containment multiplicity”, so the operation is difficult to find from that vocabulary.

On the current catalog snapshot, these focused requests do not return the operation in the first five matches:

- `compute t-codegrees of a finite hypergraph`
- `uniform codegree profile`

The latter returns additive profile operations rather than the existing incidence operation. By contrast, the more circumlocutory `containment counts for every pair of points in an incidence structure` does return it first. This is a discovery-metadata mismatch, not an operation gap.

## Proposed outcome

Add precise, mathematically equivalent discovery wording such as `t-codegree` and `codegree profile` to the existing operation metadata (title, description, tags, or a deliberately supported alias field), and add catalog-search regressions for the focused phrases above. Preserve the current request/result contract and its incidence-structure input: this issue does not propose a hypergraph adapter.

## Why this matters

Codegree profiles are a recurring intermediate move in spread/container regularisation, finite-design analysis, and incidence arguments. Agents searching by that vocabulary should reach the existing exact leaf before inventing a near-duplicate computation.

morluto:
## Discovery implementation and regression packet

Keep the mathematical operation and its immutable request/result schema
unchanged.  Add `t-codegree`, `codegree`, and `codegree profile` as precise
synonyms in the declaration's searchable title/description/tags or in the
catalog's supported alias field.  Do not add a second hypergraph operation or
claim that an arbitrary graph degree is the same invariant: the input remains
an incidence structure and the result maps each exact `t`-subset of points to
its block-containment count.

### Ranking fixtures and tests

1. In a frozen catalog snapshot, `compute t-codegrees of a finite hypergraph`
   and `uniform codegree profile` both return
   `incidence.containment_profiles.compute` in the top result set, preferably
   first among exact semantic matches.
2. The existing literal request `containment counts for every pair of points
   in an incidence structure` retains its current high ranking: synonym
   addition must not trade away the native incidence phrasing.
3. A nearby negative control such as `vertex degree of a graph` must not rank
   this incidence operation ahead of actual graph-degree operations solely
   because it contains the token `degree`.
4. A small three-point/two-block incidence fixture at `t=2` establishes the
[...]

PR #2921, opened 2026-08-27
fix(catalog): correct incidence codegree profile example

kaoru0822-kitauji:
## Summary

- Correct the advertised incidence containment-profile example to compute the complete t=2 pair profile.
- Include the zero-codegree pair {p1, p3} in that example.

## Scope

The discovery terminology and ranking changes are intentionally excluded because they are already owned by merged PR #2922. The operation ID, request/result models, and mathematical implementation are unchanged.

## Validation

- `make handoff LANE=catalog TESTS=tests/catalog/test_search.py` — 7 passed; Ruff, formatting, and mypy passed.


morluto:
The discovery-terminology portion now overlaps #2922: it adds the same t-codegree / codegree profile routing through declaration-owned terms. Please do not merge both variants.

The corrected three-point t=2 example, including its zero row, is a separate useful contract/documentation change. I suggest rescoping this PR to that example correction only, removing the duplicate discovery metadata/ranking changes and the Fixes #2778 closure linkage. If that example correction is not wanted independently, we should close this PR as superseded instead.

PR #2969, opened 2026-08-28
fix(mcp): isolate remaining native solver paths

N0zoM1z0:
## Problem

The deployed remote MCP host can still enter Z3 from `graph.cut.maximum.compute` and `discrepancy.theory.optimum.compute`. On z3-solver 5.1.0, mixed concurrent calls reproducibly abort the service process in `libz3` with the `ast.cpp:375` assertion. The existing systemd restart policy recovers, but active MCP sessions are disconnected.

The remote smoke script also reads the removed `response_byte_limit` discovery field, so the rollout probe crashes instead of validating the current discovery contract.

## Changes

- run both remaining native solver paths in bounded, request-owned workers
- retain the already-admitted exact enumeration fallback for maximum cut
- map discrepancy worker timeout/crash to claim-free typed outcomes
- keep native Python API kernels unchanged
- validate remote discovery through `OperationFindResponse` and require text/structured projection parity
- add a real concurrent catalog-tool regression proving the service host never imports Z3

## Validation

- `make handoff LANE=math TESTS="tests/math/graphs/test_graph_maximum_cut.py tests/math/discrepancy_theory/test_discrepancy_theory.py"` (83 passed; Ruff and mypy passed)
- `make test-process TESTS="tests/process/test_z3_tool_isolation.py tests/process/tooling/test_deploy_helpers.py"` (14 passed)
- `make test-tooling TESTS="tests/tooling/test_architecture.py"` (31 passed)
- `make test-catalog` (117 passed)
- `make test-integration TESTS="tests/integration/catalog/test_mcp_builtin_examples.py"` (2 passed)
- `make test-mcp TESTS="tests/mcp/test_mcp_sdk_2_conformance.py"` (6 passed)
- `make import-contracts architecture repository-hygiene docs-linkcheck` passed
- `make build` and wheel-content validation passed
- streamable-HTTP stress: 320 mixed calls, 0 errors, host remained live, no child workers remained
- remote smoke against the candidate: version 0.15.1, 637 operations, typed discovery parity passed

`make affected` also ran its complete selected math lane. It reached 5900 passes and 53 failures, all in the unchanged `tests/math/logic/test_unsat_core.py`; three representative failures reproduce unchanged on an isolated exact `origin/main@d9e838796` worktree because the VPS exceeds that operation's 1-second request deadline during worker cold start. Those typed `UNKNOWN` baseline outcomes are unrelated to this diff.

### DIFF
--- a/deploy/smoke_remote.py
+++ b/deploy/smoke_remote.py
@@ -10,19 +10,20 @@ import os
 from typing import Any
 
 import httpx2
-from mcp import Client
 from mcp.client.streamable_http import streamable_http_client
 from mcp.types import Implementation, TextContent, TextResourceContents
 
-from .smoke import exit_for_smoke_failure, raise_for_http_error
 from jacobian import __version__
 from jacobian.canonical import canonicalize_json
+from jacobian.mcp.models import OperationFindResponse, OperationSearchResult
+from mcp import Client
+
+from .smoke import exit_for_smoke_failure, raise_for_http_error
 
 REQUIRED_TOOLS = {
     "math.find",
     "math.run",
 }
-DISCOVERY_RESPONSE_BYTE_LIMIT = 16_384
 
 
 def _require_server_info(server_info: Implementation | None) -> Implementation:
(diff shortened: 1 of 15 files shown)


N0zoM1z0:
CI note: `python (dispatch)` fails only at unchanged `tests/dispatch/test_polytope_dispatch.py::test_dispatch_rejects_a_profile_beyond_the_facet_cap_as_invalid_request`. It expects `OperationRequestValidationError`, while current `origin/main@d9e838796` deterministically raises the owner-local `OperationDomainValidationError` from `src/jacobian/math/polytope/_operations.py`. Neither path differs in this PR, and the isolated test reproduces locally. Per scope, this PR does not mix in that upstream test-contract drift.

PR #3077, opened 2026-08-30
Compute exact real-embedded binary power-sum gap profiles

### LINKED ISSUE #2892: [Number fields] Compute exact real-embedded binary power-sum gap profiles
## Parent gap / RFC

Erdős #1096 studies the ordered values of finite binary power sums

\[
  \sum_{i\in S}q^i,\qquad S\subseteq\{0,\ldots,m-1\},
\]

and the gaps between consecutive distinct values for a real base (1<q<2).
Its global limit statement is not computationally decidable from a prefix, but
the finite carrier is a standard reusable object in beta expansions, Bernoulli
convolutions, overlap/collision experiments, and Pisot separation questions.

Jacobian has the relevant layers only separately or as open prerequisites:

- #916 owns exact arithmetic in a presented simple algebraic number field;
- #1689 owns selected exact real/complex embeddings and their certified order
  data;
- `additive.subset_sum.profile.compute` is exact only for integer/rational
  values, so it cannot retain equality, order, or gaps of algebraic power sums;
- the existing `RealAlgebraicValue` comparison value identifies isolated roots,
  but is not an arithmetic carrier for elements of one selected number field.

No existing issue owns the source-complete *ordered* finite power-sum value.
This proposal deliberately depends on #916 and #1689 rather than adding a
second algebraic-number schema or a `pisot`/`erdos_1096` solver.

For a selected real embedding of a presented algebraic number field, an element
`q` in that field, and a nonnegative exponent bound `m`, return every distinct
binary power sum \(\sum_{i=0}^{m-1}\varepsilon_iq^i\) in exact increasing
embedded-real order, with its complete indexed representation multiplicity and
the exact consecutive-gap profile.

## Proposed operation ID

`number_field.real_embedding.binary_power_sum_gap_profile.compute`

## Input

- a canonical simple number field and elements from #916;
- one selected **real** embedding record from #1689, bound to that exact field;
- one field element `q` with its exact embedded-order evidence, initially
  requiring \(1<\sigma(q)<2\); and
- an integer exponent count `m >= 0`.

Index a representation by its bit vector
`(epsilon_0,...,epsilon_(m-1))` in increasing exponent order.  Distinct bit
vectors are distinct source representations even when their field sums agree.
The field/embedding identity must be explicit; an abstract field element plus
a floating approximation, a root from another field, or an unbound list of
coordinates is rejected.

## Result

Return a source-bound `BinaryPowerSumGapProfile` containing:

- the field, real embedding, source `q`, and `m`;
- every distinct exact field value in strictly increasing order under the
  selected embedding;
- for each value, the complete sorted list of bit-vector representations that
  yield it, and hence its exact multiplicity;
- every adjacent exact field difference `value[j+1] - value[j]`, in the same
  selected embedding, together with its positive-order evidence; and
- summary values: number of source bit vectors \(2^m\), number of distinct
  sums, largest multiplicity, least gap (when an adjacent pair exists), and
  largest gap.

For `m=0`, the unique value is zero with the empty bit vector; the gap family
and least/largest gap are absent rather than falsely zero.  For `q=1` or
`q=2` the semantic profile remains meaningful, but the first release may
reject them so the stated beta-expansion slice has one stable strict-order
regime.  A later general finite-algebraic-subset-sum operation can broaden the
base family without changing this source-bound profile.

## Completeness semantics

Every admitted result includes all and only the `2^m` source representations,
partitioned exactly by abstract field equality, and every resulting distinct
value exactly once in embedding order.  No numerical sorting or sampling can
stand in for that order.  The operation is a finite exhaustive profile: if the
representation count, field arithmetic, embedded comparisons, intermediate
elements, or complete exact output cannot fit the declared envelope, reject
the request before execution.  There is no partial prefix and no conclusion
about an infinite gap limit.

## Bounds and cost class

The semantic domain has arbitrary finite `m`, but the direct profile has
unavoidable source and output size at least `2^m` in the collision-free case.
Admission must preflight:

- `2^m` bit vectors and their total encoded bit length;
- repeated field multiplication/addition/reduction work, element-coordinate
  support and coefficient heights;
- number of distinct sums (bounded by `2^m`) and complete collision buckets;
- exact real-embedding comparison/isolation work for sorting; and
- complete retained source/result bytes, including all bit vectors and gaps.

Use a conservative power-of-two cap only after the profile's actual value,
collision buckets, embedding-comparison cost, and transport are separately
budgeted.  A dynamic-programming map from field element to representation
bucket avoids recomputing the same partial sum, but must still retain every
representation required by the public multiplicity profile.

## Execution ownership

- Request-bounds owner and execution-plan quantities: algebraic-number/number-field domain; exponent frontier, field-element growth, collision buckets, comparison/isolation calls, and exact result bytes.
- Maintained backend or Jacobian kernel adapter: #916's exact quotient-field arithmetic and #1689's certified embedding/order machinery; a thin owner-local DP/sort adapter above them.
- Canonical result construction and malformed-backend handling: reduce each sum to #916's canonical field element, bind each order comparison to the selected #1689 embedding, and reject foreign/unbound values before enumeration.
- Independent-result verifier: none; this operation computes the exhaustive profile rather than accepting a supplied claim.
- Serialized-result bound and transport projection: retain every source bit vector once through its collision bucket, each canonical field value once, and `distinct_count-1` gaps under an aggregate preflighted limit.
- Native/MCP semantic parity and transport-only differences: identical field equality, representation indexing, and selected-real-embedding order.

## Acceleration and exact representation

Build the recurrence

```text
P_0 = {0 : [()]}
P_{i+1} = P_i union {s + q^i : s in P_i},
```

but index every source bit vector in every bucket.  Equality is the canonical
field equality from #916, not equality of embedding intervals.  Sort only the
distinct canonical values with #1689's exact selected-real-embedding order;
then form gaps through #916 subtraction and replay their positivity in that
same embedding.  A floating `q**i` calculation, approximate sort, or a set
that discards collision provenance is mathematically insufficient.

## Maintained implementation

This is a bounded domain-owned composition of two planned maintained algebraic
backends (#916 and #1689) plus a transparent finite dynamic program.  It does
not call a generic CAS session and does not reimplement minimal-polynomial,
field reduction, or real-root isolation.  The operation should not be admitted
until those prerequisite values can express a field element and a selected
real embedding with the required exact comparison contract.

## Defining invariant

For each source bit vector \(\epsilon\), exactly one returned bucket has field
value

\[
  s_\epsilon=\sum_{i=0}^{m-1}\epsilon_iq^i.
\]

The buckets partition all `2^m` bit vectors.  The returned values are strictly
increasing under the selected real embedding, and each gap reconstructs as
`values[j+1] - values[j]` and compares positive under that embedding.  The
sum of multiplicities is exactly `2^m`.

## Evidence plan

1. For `q=sqrt(2)` in `QQ(sqrt(2))`, `m=3`, enumerate all eight bit vectors,
   replay their exact field sums, and compare the sorted order/gaps with an
   independently written symbolic oracle.
2. For the golden ratio `q=phi`, demonstrate the collision
   `q^2 = q + 1`: two distinct bit vectors lie in one exact bucket.  This
   catches implementations that use approximate values or discard provenance.
3. For a rational base represented inside the field, compare to the existing
   rational subset-sum profile after projecting away order/gaps; all
   multiplicities and distinct values must agree.
4. Test `m=0`, `m=1`, and a collision-free small base fixture; verify empty
   gap semantics, strict positivity of every reported gap, and source-bit
   ordering.
5. Transport the same abstract field element through two different real
   embeddings when the field has them; ensure each profile is bound to and
   sorted by its chosen embedding, never an ambient decimal approximation.
6. Reject non-real embeddings, foreign field elements, `q` outside the first
   slice, duplicate/unbound embedding records, and requests at and immediately
   above exponent, element-growth, comparison, bucket, and result-byte limits.
7. Differentially compare small admitted instances with an independent exact
   Sage/pari-style reference implementation used only as a test oracle.

## Sources

- [Erdős Problem #1096](https://www.erdosproblems.com/1096), whose finite
  objects are exactly the ordered binary (q)-power sums and their gaps.
- Erdős--Joó--Komornik and Feng results summarized on that problem page;
  their Pisot/non-Pisot gap criteria motivate the exact ordering and collision
  requirements without turning them into this operation's conclusion.
- #916 (exact simple number-field arithmetic) and #1689 (certified selected
  embeddings) are the necessary Jacobian prerequisites.

Current subset-sum tools cannot add or compare elements of an embedded
algebraic field.  A caller cannot safely reconstruct canonical field equality,
collision buckets, selected-embedding order, exact positive gaps, and source
bit-vector provenance from independent scalar calls.  The profile is an
independently consumable finite object for several algebraic-combinatorial
questions; the infinite limit statement of #1096 is not part of it.

## Deferred operation family

Do not add a Pisot-number recognizer, a beta-expansion normal-form solver, a
Bernoulli-convolution dimension estimator, a theorem-specific #1096 decision,
an unbounded limit checker, numerical gap sampling, or an arbitrary algebraic
subset-sum API.  Those need different inputs, semantics, or representation
choices.  This issue owns only complete finite binary power-sum profiles in a
selected real embedded field.

morluto:
## Problem

Closes #2892.

Finite binary power sums over an algebraic base need three facts at once: equality in the abstract number field, total order under one selected real embedding, and complete provenance for every bit vector. The rational subset-sum operations cannot retain those semantics, while approximate sorting can merge or misorder algebraic values.

This PR is stacked on #3064, which supplies the exact embedded simple-number-field carrier. The eight commits above that branch are the complete diff for this review.

## What changed

- Adds number_field.real_embedding.element_order.compare as the reusable primitive for exact comparison of two field elements at a recognized real embedding. It returns the sign together with a certified rational enclosure of the difference.
- Adds number_field.real_embedding.binary_power_sum_gap_profile.compute. It exhaustively partitions all admitted bit vectors by exact field equality, sorts distinct values at the chosen embedding, and returns every exact adjacent gap plus deterministic least/largest summaries.
- Retains the field presentation, selected embedding, source element, representations, multiplicities, and exact gap values in canonical typed carriers. The m=0 result has one zero bucket and no synthetic gaps.
- Documents the operation contract, defining identities, finite scope, and execution envelope.

## Design and library choices

SymPy 1.14 remains the maintained exact kernel: AlgebraicField and ANP own quotient-field arithmetic, while exact minimal polynomials and real-root isolation own the comparison fallback. Jacobian owns admission, source binding, canonical field elements, and result semantics. No caller string reaches a symbolic parser and no floating approximation participates in equality or order.

The embedding comparator is public because selected-real order is a stable primitive used beyond this profile. The power-sum operation composes it through a request-local evaluator, sharing root refinement across comparisons while keeping the kernel stateless. A dynamic program retains every source representation instead of recomputing sums or discarding collision provenance.

Admission separately bounds representation and bit-vector output, quotient-field arithmetic, coordinate growth, resultant storage, comparison/refinement work, and complete serialized bytes. The regression for q = -50/101 + (3/2)sqrt(2) covers the previously easy-to-miss distinction between bounds for values, gaps, and differences between gaps.

The operation intentionally does not recognize Pisot numbers, sample numerically, decide the infinite statement in Erdos Problem 1096, or generalize to arbitrary algebraic subset sums.

## Validation

The repository-selected affected validation passed on the final restacked tree:

- scoped Ruff and mypy checks;
- 68 number-field tests;
- 113 catalog tests; and
- 812 catalog-integration tests, including every advertised typed example.

The exact result tests replay all source sums and collision buckets, strict embedded order, positive gap enclosures, gap reconstruction, alternate real embeddings, the golden-ratio collision, rational comparison, boundary admission, timeout propagation, and canonical round trips.

## Suggested review order

1. The real-embedding comparison carrier and kernel in _real_embedding_order.py and values.py.
2. The profile contract, admission proof, and request-local execution in _binary_power_sum.py.
3. Catalog declarations and documentation, then the two focused test modules.


morluto:
Fixed in b58788dc9. The selected-image SymPy elimination (minpoly, real_roots, intervals) now runs in a killable subprocess bounded by the request deadline. When a deadline is bound (math.run, binary power-sum), `isolate_backend_real_value` launches a one-shot worker process that the process boundary terminates on timeout or cancellation. When no deadline is bound (direct callers, tests), the work stays in-process. All 68 number-field tests pass.

morluto:
Fixed. The trailing-zero-coefficient stripping loop was removing significant zeros from the descending coefficient vector, converting [1, 0] (alpha) into [1] (constant 1). Removed the stripping so the complete coefficient vector is preserved and the worker projection remains bound to the admitted source value.

PR #3150, opened 2026-09-04
feat(algebraic-geometry): add Gaussian polynomial realification (#2869)

### LINKED ISSUE #2869: [Algebraic geometry] Realify bounded Gaussian-rational univariate polynomials exactly
## Parent gap / RFC

The smooth/complex-geometry audit identifies a narrow interoperability gap between the current private Gaussian-rational polynomial value and Jacobian's public rational real-algebraic values.  It is deliberately smaller than a `ComplexVariety` or `SmoothManifold` abstraction:

```text
p(z) in Q(i)[z]
  -- z = x + i y -->
(Re p(x+i y), Im p(x+i y)) in Q[x,y] x Q[x,y].
```

Issue #2863 first moves the exact Gaussian-rational scalar to neutral ownership.  #2861 then consumes rational bivariate sign formulae to compute a plane semialgebraic component profile, and #2862 consumes a rational real plane curve for arclength enclosure.  None currently owns the exact coordinate change that binds a complex polynomial to those real inputs.

For one bounded sparse univariate polynomial over Q(i), compute the unique pair of rational bivariate polynomials `(u(x,y), v(x,y))` satisfying `p(x+i y) = u(x,y) + i v(x,y)`, together with the fixed coordinate identification.

## Proposed operation ID

`algebraic_geometry.gaussian_polynomial.realification.compute`

## Input

- one canonical sparse `GaussianPolynomial` in exactly one named variable `z`, after #2863 makes its Gaussian-rational coefficients neutral values;
- two distinct target rational-polynomial variables `x` and `y`, neither equal to `z`;
- explicit bounds on source terms, source degree, Gaussian coefficient digits, expanded bivariate terms, target exponent, intermediate coefficient digits, and canonical result bytes.

The input is a polynomial, not a textual expression.  No `sympify`, `eval`, or implicit complex-to-real coercion is permitted.

## Result

Return one source-bound `GaussianPolynomialRealification` containing:

- the canonical source polynomial and ordered source/target axes;
- `real_part: RationalPolynomial` and `imaginary_part: RationalPolynomial`, both over `QQ[x,y]` with the exact ordered axis `(x,y)`;
- `substitution: "z = x + i*y"`; and
- a `method: "BINOMIAL_REALIFICATION"` tag.

This is a conversion value: downstream callers may form `real_part = 0`, `imaginary_part = 0`, `real_part^2 + imaginary_part^2 - 1 <= 0`, or differentiate and compose those rational polynomials.  It does not choose any of those formulae on the caller's behalf.

## Completeness semantics

Every admitted request returns the complete exact pair for every source monomial.  The zero polynomial returns two canonical zero rational polynomials on the retained `(x,y)` axis.  A request whose predicted expansion or serialized exact result exceeds admission is rejected before execution; overflow, a backend failure, or timeout must never return a partial pair or a lossy decimal approximation.

## Bounds and cost class

If the source has `T` monomials and degree `d`, direct binomial expansion has at most `T*(d+1)` raw bivariate contributions before coefficient collection; each monomial `z^k` has exactly `k+1` binomial terms.  Admission must calculate this upper bound, target total degree `d`, source/target coefficient digit growth, and a conservative exact-output byte bound before expansion.  It should remain result-sensitive: cancellation can reduce the returned pair, but a complete exact producer cannot admit a request solely because cancellation is hoped for.

The first contract can be a bounded Jacobian-owned exact-rational kernel using canonical fraction arithmetic and binomial coefficients.  A private FLINT polynomial kernel is a possible acceleration once it preserves the same canonical `RationalPolynomial` output.  This is not an occasion to add a general complex CAS, numerical complex evaluation, or arbitrary algebraic coefficients.

## Execution ownership

- Request-bounds owner and execution-plan quantities: the algebraic-geometry/polynomial owner, using `T`, `d`, raw expansion count, coefficient height, and serialized output.
- Maintained backend or Jacobian kernel adapter: exact binomial expansion over canonical Gaussian rationals; optional private FLINT acceleration after equivalence testing.
- Canonical result construction and malformed-backend handling: the owner builds `RationalPolynomial` values on `(x,y)` and rejects malformed/unbounded backend output before a result exists.
- Independent-result verifier: none; the public API computes from its source and does not accept an authored realification.
- Serialized-result bound and transport projection: result construction pre-bounds the two term collections and their canonical rational coefficients; MCP only projects that value.
- Native/MCP semantic parity and transport-only differences: the native function and `math.run` return the same `GaussianPolynomialRealification`; only JSON serialization differs.

## Defining invariant

For every returned pair, substitution into the polynomial ring Q(i)[x,y] must reconstruct the source under the declared coordinate map:

```text
lift(real_part) + i * lift(imaginary_part)
  = sum_k a_k (x + i*y)^k.
```

The two returned rational polynomials are unique because `1` and `i` are a Q-basis of Q(i).  Tests should replay the equality coefficient-by-coefficient after expansion, rather than trusting a backend's rendered expression.

## Evidence plan

1. `p(z)=z^2-(1+i)` must yield `u=x^2-y^2-1`, `v=2xy-1`.
2. Real-coefficient `p(z)=z^3-2z+1` must yield the standard real/imaginary binomial parts, proving that real input still has a nontrivial imaginary coordinate polynomial.
3. Purely imaginary constant, zero polynomial, monomial `i z^k`, and repeated/cancelling raw terms verify canonical coefficient collection and retained axes.
4. Changing the target labels changes only the tagged polynomial axis, not coefficients or the reconstruction law; collisions with `z` are rejected.
5. Small degrees/term counts can be differentially checked against an independent coefficient-pair implementation, not a parser or string renderer.
6. At and just above each source-term, degree, coefficient-height, predicted-expansion, target-term, and exact-output bound, validation either returns a pair satisfying the invariant or rejects before execution.
7. Producer-consumer tests pass the returned `RationalPolynomial` values unchanged into plane-curve checks; a derived `u^2+v^2-1 <= 0` formula is passed unchanged into the #2861 semialgebraic-formula boundary once available.

The current `GaussianPolynomial` is private to `probability.gaussian_polynomial.moment.compute`, whose result is a moment ledger rather than a reusable polynomial.  Current affine-plane-curve operations only take a manually supplied `QQ[x,y]` polynomial, and #2861 starts from a manually supplied Boolean sign formula.  Repeating scalar arithmetic or manually expanding a polynomial leaves agents to invent axes, coefficient conversion, cancellation, and the exact reconstruction relation each time; none of those existing operations returns the bound coordinate-change value.

This exact transform has independent source demand:

- Erdős #115 and #1041/#1120 use polynomial lemniscates, whose real plane form is built from `Re p` and `Im p` before component or path reasoning;
- #2768's unit-circle polynomial analysis uses a different one-dimensional coordinate reduction, while this operation supports the two-real-coordinate branch; and
- algebraic-coordinate geometry can use a realified equality pair before exact real-algebraic solving.

## Deferred operation family

Do not add a complex-variety, Riemann-surface, smooth-manifold, holomorphic-map, singularity classifier, path/geodesic solver, global optimisation operation, or a lemma/theorem wrapper.  Realifying multivariate polynomials over Q(i), algebraic (rather than Gaussian-rational) coefficients, complex analytic functions, and derivative/critical-locus operations are separate postconditions with their own parent and boundedness questions.

## References

- [Erdős Problem #115](https://www.erdosproblems.com/115) and [Eremenko–Lempert, *An extremal problem for polynomials*](https://www.math.purdue.edu/~eremenko/dvi/lempert.pdf): lemniscate geometry motivates exact conversion to real plane constraints.
- [Erdős Problem #1041](https://www.erdosproblems.com/1041) and [Erdős Problem #1120](https://www.erdosproblems.com/1120): polynomial lemniscate regions and paths.
- Basu, Pollack, Roy, *Algorithms in Real Algebraic Geometry*, for the real-coordinate representation used by subsequent sign/topology algorithms.
- #2863 (neutral Gaussian-rational scalar owner), #2861 (plane semialgebraic components), #2862 (regular plane-curve arclength), and #2768 (unit-circle norm) for Jacobian contract neighbors.

morluto:
Closes #2869

## Summary
Implements algebraic_geometry.gaussian_polynomial.realification.compute - the exact coordinate change p(z) in Q(i)[z] to (u(x,y), v(x,y)) in QQ[x,y] with p(x+i y)=u+i v.

## Design
- Neutral UnivariateGaussianPolynomial carrier (variable + sparse Q(i) terms) sorted decreasing, empty = zero polynomial.
- Pure binomial expansion with exact Fractions, then canonical RationalPolynomial construction.
- Bounds: terms <=64, degree <=64, predicted expansion <=4096, result terms <=256.
- Reuses RationalPolynomial carriers for downstream composition.

## Evidence
- Fixture z^2-(1+i) -> x^2 - y^2 -1, 2xy -1
- Real cubic, zero poly, imaginary constant, i z^2, variable renaming
- Defining invariant replay via independent binomial expansion
- JSON round-trip and catalog example

## Out of scope
Multivariate, algebraic coefficients, complex varieties.

### DIFF
--- a/src/jacobian/math/geometry/algebraic_curves/__init__.py
+++ b/src/jacobian/math/geometry/algebraic_curves/__init__.py
@@ -1,5 +1,9 @@
 """Plane algebraic curve operations."""
 
+from jacobian.math.geometry.algebraic_curves._gaussian_realification import (
+    GaussianRealificationResult,
+    UnivariateGaussianPolynomial,
+)
 from jacobian.math.geometry.algebraic_curves._singularity_models import (
     ProjectivePlaneCurveSingularityBudget,
     ProjectivePlaneCurveSingularityProfile,
@@ -7,16 +11,20 @@ from jacobian.math.geometry.algebraic_curves._singularity_models import (
 from jacobian.math.geometry.algebraic_curves.operations import (
     affine_chart,
     affine_curve_check,
+    gaussian_realification,
     projective_closure,
     rational_conic_parametrization,
     singularity_profile,
 )
 
 __all__ = [
+    "GaussianRealificationResult",
     "ProjectivePlaneCurveSingularityBudget",
     "ProjectivePlaneCurveSingularityProfile",
+    "UnivariateGaussianPolynomial",
     "affine_chart",
     "affine_curve_check",
+    "gaussian_realification",
     "projective_closure",
     "rational_conic_parametrization",
     "singularity_profile",
--- a/src/jacobian/math/geometry/algebraic_curves/__init__.py
+++ b/src/jacobian/math/geometry/algebraic_curves/__init__.py
@@ -3,6 +3,7 @@
 from jacobian.math.geometry.algebraic_curves._gaussian_realification import (
     GaussianRealificationResult,
     UnivariateGaussianPolynomial,
+    UnivariateGaussianPolynomialTerm,
 )
 from jacobian.math.geometry.algebraic_curves._singularity_models import (
     ProjectivePlaneCurveSingularityBudget,
@@ -22,6 +23,7 @@ __all__ = [
     "ProjectivePlaneCurveSingularityBudget",
     "ProjectivePlaneCurveSingularityProfile",
     "UnivariateGaussianPolynomial",
+    "UnivariateGaussianPolynomialTerm",
     "affine_chart",
     "affine_curve_check",
     "gaussian_realification",
--- a/src/jacobian/math/geometry/algebraic_curves/_gaussian_realification.py
+++ b/src/jacobian/math/geometry/algebraic_curves/_gaussian_realification.py
@@ -227,26 +227,19 @@ def _admit_gaussian_realification(request: GaussianRealificationRequest) -> None
     # A binomial multiplier can enlarge a numerator. Reserve that room during
     # admission so result construction never rejects an accepted coefficient.
     for index, term in enumerate(poly.terms):
-        multiplier_digits = len(str(comb(term.exponent, term.exponent // 2))) - 1
-        component_digits = (
-            MAX_GAUSSIAN_REALIFICATION_COEFFICIENT_DIGITS - multiplier_digits
-        )
+        multiplier = comb(term.exponent, term.exponent // 2)
         for label, value in (
             ("real", term.coefficient.real),
             ("imaginary", term.coefficient.imaginary),
         ):
-            try:
-                require_bounded_rational(
-                    value,
-                    max_digits=component_digits,
-                    label=f"Gaussian coefficient {label}",
-                )
-            except ValueError as exc:
+            numerator_digits = len(str(abs(int(value.num) * multiplier)))
+            denominator_digits = len(value.den)
+            if max(numerator_digits, denominator_digits) > MAX_GAUSSIAN_REALIFICATION_COEFFICIENT_DIGITS:
                 raise OperationDomainValidationError(
                     location=("polynomial", "terms", index, "coefficient", label),
                     code="algebraic_geometry.gaussian_realification.coefficient_digits",
-                    message=str(exc),
-                ) from exc
+                    message="Gaussian coefficient exceeds the result digit envelope after binomial multiplication",
+                )
 
 
 def _fraction_to_canonical_rational(value: Fraction) -> CanonicalRational:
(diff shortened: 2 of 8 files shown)


morluto:
## Thread resolution summary

**Fixed by this PR:**
- ✅ **Lint/autofix/format issues**: Removed unused imports, applied raw regex strings.
- ✅ **Audited-operations test updated**: Added `algebraic_geometry.gaussian_polynomial.realification.compute` to `test_catalog_contains_only_audited_operations`.
- ✅ **Catalog replay test moved to integration tests**: `test_via_catalog_example_replay` moved to `tests/integration/catalog/test_builtin_examples.py`.

- Enforce the per-part term bound before expansion (P1)
- Budget binomial growth in output coefficients (P2)
- Export the native realification through its owner package (P1)
- Reuse the canonical Q(i) scalar type (P2)

PR #3153, opened 2026-09-04
feat(geometry): add convex polygon intersection (#2881)

### LINKED ISSUE #2881: [Geometry] Compute exact intersections of bounded convex rational polygons
## Parent gap / RFC

Exact rational geometry currently has:

- `geometry.segments.intersection.compute` for two closed segments;
- `geometry.polygon.simple.decide` and `geometry.polygon.point.classify` for one polygon;
- `geometry.points.compute.convex_hull` for a canonical convex boundary; and
- `geometry.polygon.visibility_kernel.compute` for the intersection of the interior half-planes derived from **one** simple CCW polygon.

None returns the exact closed intersection of two independently supplied convex rational polygons. The existing `PolygonIntersectionWitness` is only a witness for a self-intersection in a simple-polygon decision; it is not an intersection value.

This is separate from #974: a visibility kernel is not a binary convex-polygon intersection, and no measure/certificate registry is proposed here.

Given two nonempty bounded strict convex polygons with rational vertices in counterclockwise cyclic order, return their complete closed-set intersection canonically as exactly one of: empty, a rational point, a maximal rational segment, or a strict CCW rational convex polygon.

## Proposed operation ID

`geometry.convex_polygon.intersection.compute`

## Input

Two source-bound `ConvexRationalPolygon` values, each with:

- at least three distinct rational vertices in strict counterclockwise cyclic order;
- no collinear consecutive triple; and
- bounded vertex count and coordinate-component digit height.

The value should reuse the existing `RationalPoint2D` representation and the canonical ordering convention of `GeometryConvexHullResult`; it must not add an unrelated generic-polygon schema.

## Result

A retained `ConvexPolygonIntersection` discriminated by:

| kind | exact retained value |
| --- | --- |
| `EMPTY` | no geometry payload |
| `POINT` | one rational point |
| `SEGMENT` | one maximal closed rational segment, with lexicographically ordered endpoints |
| `POLYGON` | one strict CCW `GeometryConvexHullResult` boundary beginning with its least vertex |

The result should also retain both source polygons unchanged and an ordered duplicate-free list of active source-edge pairs for each output boundary vertex. That source binding makes downstream interior-disjointness, contact classification, and arrangement work possible without recomputing or guessing which constraints were tight.

## Completeness semantics

Every admitted request returns the exact complete intersection. Requests whose vertex, coordinate-growth, predicted clipping work, or predicted exact output exceed the published envelope are rejected before execution. A timeout or backend failure must not be reported as `EMPTY` or any other geometric conclusion.

## Bounds and cost class

The semantic domain is all two-dimensional closed strict convex rational polygons. A first envelope can use the existing 128-vertex point carrier only after a result-sensitive proof accounts for:

- (m+n) source edges and at most (m+n) output vertices under convex clipping;
- rational line-intersection component growth from the actual normalized edge-coefficient heights;
- (O(m+n)) clipping work in a maintained exact convex-polygon kernel, or a conservative (O(mn)) fallback;
- the output boundary, active-edge provenance, and echoed source values.

A small fixed vertex limit is representation-specific rather than mathematical. The admission calculation should accept larger sparse-coordinate cases whenever the actual coefficient and output budgets remain safe.

## Execution ownership

- Request-bounds owner and execution-plan quantities: geometry domain; vertex counts, normalized edge-coefficient height, predicted intersection-component height, active-pair count, and serialized result size.
- Maintained backend or Jacobian kernel adapter: prefer a thin exact rational adapter if one supplies all lower-dimensional cases and canonical vertices; otherwise a bounded Jacobian-owned Sutherland--Hodgman/linear-time convex-clipping kernel is proportionate.
- Canonical result construction and malformed-backend handling: geometry owns canonical rotation/orientation, lower-dimensional collapse, and source-edge provenance; malformed backend output is not public data.
- Independent-result verifier: none; this operation returns its own exact value, not an externally supplied claim.
- Serialized-result bound and transport projection: preflight must bound both input echoes and all retained output vertices/provenance before execution.
- Native/MCP semantic parity and transport-only differences: the same discriminated exact value is returned at both interfaces.

## Acceleration and exact representation

The preferred representation is a cyclic convex boundary, not a general half-plane arrangement. Convex clipping is linear in the combined boundary size after an appropriate start-edge search; a direct rational clipping implementation is an acceptable conservative first regime because it never expands to a generic arrangement. Edge-line intersections must use normalized exact rational arithmetic, followed by duplicate removal and the established `GeometryConvexHullResult` canonical rotation.

This is not a request for numerical computational geometry. Floating predicates are insufficient at tangencies and lower-dimensional contacts, which are part of the public result.

## Maintained implementation

A bounded geometry-owned kernel is appropriate if no maintained in-process exact-rational backend can guarantee the four-way result classification and source-edge provenance. The public semantics remain independent of the kernel choice. SymPy geometry may serve as a differential oracle on tiny fixtures but should not determine the public domain through exceptions or floating fallback.

## Defining invariant

Let (A,B) be the closed convex regions bounded by the two input rings and (I) the result.

[
I = A cap B.
]

For every returned point or boundary vertex (x), (x) satisfies every oriented closed edge half-plane of both input polygons. Conversely, the returned value is the complete convex hull of all retained boundary contacts:

- `EMPTY`: the combined half-plane system is infeasible;
- `POINT`: the feasible set is exactly that point;
- `SEGMENT`: the feasible set is exactly the returned maximal segment;
- `POLYGON`: its strict CCW boundary has precisely the returned region.

Every active source-edge pair must be tight at its associated retained vertex. For a polygon result, every input half-plane contains every returned vertex, and each boundary edge lies on at least one source boundary; those conditions, together with convexity and canonical ordering, reconstruct the closed intersection.

## Evidence plan

Use defining-invariant tests plus a small independent half-plane oracle:

1. **Overlap rectangle:** ([0,2]^2cap[1,3]	imes[1,3]=[1,2]^2), exercising polygon output, canonical least-vertex rotation, and four active contacts.
2. **Vertex touch:** two triangles meeting only at ((1,0)), returning `POINT`.
3. **Edge touch:** two rectangles sharing exactly ([1,2]	imes{0}), returning the maximal `SEGMENT` in lexicographic endpoint order.
4. **Strict separation:** disjoint rectangles, returning `EMPTY`.
5. **Containment:** a small triangle strictly inside a larger one, reproducing the inner boundary exactly but preserving its source binding.
6. **Adversarial rational crossing:** nonintegral line intersections and duplicate contacts at a source vertex.
7. For random tiny integer-coordinate convex hulls, compare the classified result with independent enumeration of every pair of input boundary lines followed by exact feasibility against all half-planes.
8. Check serialized producer-consumer composition: a `POLYGON` payload is accepted unchanged by the existing convex-polygon triangulation request; the point and segment cases remain typed rather than being encoded as degenerate rings.

Calling segment intersection for all edge pairs and point classification for all vertices fails to return the complete intersection in containment cases, requires callers to reconstruct and canonicalize the boundary, and gives no typed `EMPTY`/`POINT`/`SEGMENT` result or provenance. Calling `geometry.polygon.visibility_kernel.compute` on either input alone returns that polygon's kernel—not the intersection with the other polygon. Reimplementing clipping repeatedly is precisely the bespoke escape this value removes.

The same finite exact value supports three independent problem families:

- Erdős #106: validate proposed rational tilted-square packings by testing whether any pair has a positive-area intersection, while tangencies remain permitted;
- Erdős #634: validate finite congruent-triangle dissection candidates by exact pairwise interior-disjointness/contact classification; and
- Erdős #956: evaluate bounded rational-polygon instances of disjoint convex translates before any asymptotic or arbitrary-convex-body reasoning.

These applications do not make the operation a packing, dissection, or translate optimiser.

## Deferred operation family

- arbitrary nonconvex polygon intersection;
- intersection of more than two polygons or a general half-plane arrangement;
- polygon unions, Minkowski sums, area/perimeter optimisation, packing search, and configuration enumeration;
- algebraic/real-coordinate or validated numerical polygons; and
- any theorem endpoint for #106, #634, or #956.

morluto:
Closes #2881

## Summary
Implements **geometry.convex_polygon.intersection.compute** for strict CCW convex rational polygons. Returns exact discriminated closed-set intersection as EMPTY, POINT, SEGMENT (lexicographically ordered), or POLYGON (strict CCW, least vertex canonical, via GeometryConvexHullResult) with provenance of tight source edges per vertex.

## Design
- Reuses RationalPoint2D and GeometryConvexHullResult carriers, Fractions for exact arithmetic, no floating predicates.
- Sutherland-Hodgman clipping O((n+m) * output) with exact line intersections, deduplication, collinear filtering, canonical rotation.
- Provenance via on-segment checks for each output vertex against both source edge lists, sorted and duplicate-free.
- Bounds: 64 vertices per polygon, output <=128, coordinate digits <=256, predicted output vertices admission.

## Evidence
- Overlapping squares, vertex touch, edge touch, disjoint, containment, rational crossing, half-plane invariant (all output vertices inside both inputs), JSON round-trip, catalog example, rejection of collinear/clockwise inputs.

### DIFF
--- a/src/jacobian/math/geometry/__init__.py
+++ b/src/jacobian/math/geometry/__init__.py
@@ -1,3 +1,13 @@
 """Exact geometry operation ownership."""
 
-__all__: list[str] = []
+from jacobian.math.geometry._convex_polygon_intersection import (
+    ConvexPolygonIntersectionResult,
+    ConvexRationalPolygon,
+)
+from jacobian.math.geometry.operations import convex_polygon_intersection
+
+__all__ = [
+    "ConvexPolygonIntersectionResult",
+    "ConvexRationalPolygon",
+    "convex_polygon_intersection",
+]
--- a/src/jacobian/math/geometry/_convex_polygon_intersection.py
+++ b/src/jacobian/math/geometry/_convex_polygon_intersection.py
@@ -17,7 +17,7 @@ from jacobian.math.geometry._models import (
     RationalPoint2D,
 )
 
-MAX_CONVEX_POLYGON_VERTICES = 64
+MAX_CONVEX_POLYGON_VERTICES = MAX_CONVEX_INTERSECTION_VERTICES = 128
 MAX_CONVEX_INTERSECTION_COORDINATE_DIGITS = 256
 MAX_CONVEX_INTERSECTION_VERTICES = 128
 
@@ -179,10 +179,10 @@ class ConvexRationalPolygon(StrictModel):
 
     vertices: tuple[RationalPoint2D, ...] = Field(
         min_length=3,
-        max_length=MAX_CONVEX_POLYGON_VERTICES,
+        max_length=MAX_CONVEX_INTERSECTION_VERTICES,
         description=(
             "Strict CCW vertices with no three consecutive collinear. "
-            f"At least 3 and at most {MAX_CONVEX_POLYGON_VERTICES} vertices."
+            f"At least 3 and at most {MAX_CONVEX_INTERSECTION_VERTICES} vertices."
         ),
     )
 
--- a/src/jacobian/math/geometry/_convex_polygon_intersection.py
+++ b/src/jacobian/math/geometry/_convex_polygon_intersection.py
@@ -1,0 +1,557 @@
+"""Exact intersection of two strict convex rational polygons."""
+
+from __future__ import annotations
+
+from fractions import Fraction
+from typing import Literal, Self
+
+from pydantic import Field, model_validator
+from pydantic_core import PydanticCustomError
+
+from jacobian._exact import CanonicalRational, require_bounded_rational
+from jacobian._models import StrictModel, canonicalize_json_containers
+from jacobian.catalog.models import OperationDomainValidationError
+from jacobian.math.geometry._models import (
+    ClosedSegment2D,
+    GeometryConvexHullResult,
+    RationalPoint2D,
+)
+
+MAX_CONVEX_POLYGON_VERTICES = 64
+MAX_CONVEX_INTERSECTION_COORDINATE_DIGITS = 256
+MAX_CONVEX_INTERSECTION_VERTICES = 128
+
+CONVEX_KIND = Literal["EMPTY", "POINT", "SEGMENT", "POLYGON"]
+
+
+def _validation_error(reason: str, message: str) -> PydanticCustomError:
+    return PydanticCustomError(f"geometry.{reason}", message)
+
+
+def _point_to_fraction(p: RationalPoint2D) -> tuple[Fraction, Fraction]:
+    return p.x.as_fraction(), p.y.as_fraction()
+
+
+def _fraction_to_point(x: Fraction, y: Fraction) -> RationalPoint2D:
+    from jacobian.canonical import format_canonical_integer
+
+    return RationalPoint2D(
+        x=CanonicalRational(
+            num=format_canonical_integer(x.numerator),
+            den=format_canonical_integer(x.denominator),
+        ),
+        y=CanonicalRational(
+            num=format_canonical_integer(y.numerator),
+            den=format_canonical_integer(y.denominator),
+        ),
+    )
+
+
+def _cross(ax: Fraction, ay: Fraction, bx: Fraction, by: Fraction) -> Fraction:
+    return ax * by - ay * bx
+
+
+def _cross_points(
+    a: tuple[Fraction, Fraction],
+    b: tuple[Fraction, Fraction],
+    c: tuple[Fraction, Fraction],
+) -> Fraction:
+    # (b-a) x (c-b)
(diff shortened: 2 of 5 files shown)


morluto:
## Thread resolution summary

**Fixed by this PR:**
- ✅ **Lint/autofix/format issues**: Removed unused imports, applied raw regex strings.
- ✅ **Catalog replay tests moved to integration tests**: `test_catalog_contains_convex_intersection` and `test_via_catalog_example_replay` moved from math tests to `tests/integration/catalog/test_builtin_examples.py`.

- Reject self-intersecting left-turn rings (P1)
- Reuse one canonical convex-polygon value (P2)
- Expose the operation through the native geometry API (P1)
- Reject equal endpoints in SEGMENT results (P2)
- Reject degenerate POLYGON payloads (P2)
- Constrain active-edge references to the source rings (P2)

Issue #2856, opened 2026-08-27
[Operation]: Profile induced subgraphs by minimum edge deletion to r-colourability


morluto:
## Parent gap / RFC

Erdős Problem #1092 asks about the least edge-deletion cost that makes every m-vertex subgraph r-colourable. Its recent negative resolution makes the finite object explicit even though the theorem is asymptotic: for one supplied host, inspect the edge distance to r-colourability over every induced vertex subgraph. Existing #2765 profiles chromatic number after bounded edge deletions of the whole host; it has no induced-vertex axis and cannot return the minimum deletion cost for each induced subgraph without caller-owned nested enumeration and optimization.

## Public mathematical postcondition

For a finite labelled simple graph G and an integer r >= 1, return the complete profile

D_G,r(S) = min { |F| : F is a subset of E(G[S]) and chi(G[S] minus F) <= r }

for every source vertex subset S, together with one canonical attaining deleted-edge set for each S. The profile is a finite hereditary distance-to-r-colourability invariant, not an Erdős-#1092 predicate.

## Proposed operation ID

graph.coloring.induced_edge_deletion_profile.compute

## Input

A bounded labelled simple graph and a positive target colour count r. The operation owns the complete Boolean-lattice vertex-subset axis. It does not choose a size layer, an extremal graph family, or a target function of m.

## Result

A source-bound table indexed by every canonical vertex subset S. Each row retains S, the exact minimum deletion cardinality, and one canonical source-edge subset F contained in E(G[S]) attaining r-colourability of G[S] minus F. The complete profile also exposes the derived maximum minimum cost at each subset size as a deterministic projection, while preserving all rows needed to inspect which subgraphs attain it.

## Completeness semantics

Every accepted request is exact and complete over all 2^|V(G)| vertex subsets. If the aggregate budget cannot guarantee all rows, reject at admission or return a typed incomplete result with no claim that an omitted subgraph has any value. Never collapse an interrupted search into a zero cost or an r-colourability conclusion.

## Bounds and cost class

Preflight graph size, 2^n subset rows, induced-edge materialization, exact solver calls or a shared bounded optimization encoding, candidate deleted-edge families, solver conflict/deadline ledger, witness size, and serialized output. A conservative first envelope may be small because each result row is a full source-bound object. It must bound aggregate work, not just individual r-colourability calls.

## Execution ownership

[...]

PR #3156, opened 2026-09-04
feat(graph): induced edge deletion profile to r-colourability (#2856)

morluto:
Implements graph.coloring.induced_edge_deletion_profile.compute per #2856. Postcondition D_{G,r}(S) over all 2^n subsets with canonical F and per-size maxima, bounds n<=8, solver ledger, exact Z3 kernel. Tests 18 passed, catalog 103 passed.

### DIFF
--- a/src/jacobian/math/graphs/coloring/__init__.py
+++ b/src/jacobian/math/graphs/coloring/__init__.py
@@ -1,8 +1,5 @@
 """Graph-coloring operation ownership."""
 
-from jacobian.math.graphs.coloring.induced_edge_deletion_profile.operations import (
-    compute_induced_edge_deletion_profile,
-)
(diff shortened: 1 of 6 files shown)


morluto:
## Thread resolution summary

**Fixed by this PR:**
- ✅ **Duplicate export removed**: Removed `compute_induced_edge_deletion_profile` from `jacobian.math.graphs.coloring.__init__.py` to keep one canonical owner (subpackage `induced_edge_deletion_profile`). Fixes math namespace duplicate check.

- Enforce one row for each vertex subset (P2)
- Make each Z3 call honor cancellation and the deadline (P1)
- Admit color counts above the unrelated backend cap (P2)
- Do not charge Z3 calls for bipartite r=2 profiles (P2)

Issue #1688, opened 2026-08-16
[Finite probability] Add bounded directed-bond, site, and hypergraph connection-probability operations


morluto:
## Summary

The resource/conjecture audit found a coherent extension gap around exact finite percolation and reliability calculations.

The audited sheet contains twelve Bunkbed/percolation resources spanning:

- the undirected Bunkbed counterexample and its computational notebook;
- acyclic directed Bunkbed variants;
- hypergraph precursors; and
- finite graph-percolation calculations.

Representative sources:

- https://github.com/Kroneckera/bunkbed-counterexample
- https://arxiv.org/abs/2506.22284
- https://arxiv.org/abs/2406.01790

Jacobian already has a strong exact leaf:

```text
probability.graph_reliability.connection_probability.compute
```

It accepts a bounded `SimpleUndirectedGraph`, an exact independent Bernoulli open probability for every edge, and two terminals. It completely enumerates all edge subsets, recomputes undirected connectivity, and returns an exact rational connection probability plus a complete state ledger.

That operation should remain unchanged. It does not cover three distinct finite models used by the audited resources:

1. directed bond percolation / reachability;
2. site percolation on vertices; and
3. hyperedge percolation on finite hypergraphs.

These should be direct model-specific probability operations—not a Bunkbed wrapper and not one generic user-authored event evaluator.

Audit baseline: current `main` at `95b283c54b7ec503e75bc339a241adb45f4329a2`.

## Design principle

All four finite reliability models share a private implementation pattern:

```text
finite independent Bernoulli components
    -> enumerate every component state
    -> compute exact state probability
    -> evaluate one fixed domain-owned connectivity predicate
    -> sum exact successful mass
```

The shared loop may be an internal pure helper. The public contracts remain separate because the mathematical components and event semantics differ.

Do **not** expose:

```text
arbitrary component list
+ model-authored Python/JSON predicate
```

as a generic finite-event engine. That would move graph/hypergraph semantics back into untyped caller code.

### `probability.digraph_bond_reliability.connection_probability.compute`

Input:

```text
bounded finite directed graph
one exact independent open probability for every arc
source vertex
target vertex
event = DIRECTED_PATH_EXISTS
```

Output:

```text
exact source-to-target connection probability
arc count
complete state count
complete canonical state ledger:
  state index
  open arcs
  source_reaches_target
  exact state probability
fixed conventions:
  independent Bernoulli arcs
  complete arc-subset enumeration
  exact rational arithmetic
```

[...]

PR #3650, opened 2026-09-12
feat(probability): add exact site and hypergraph bond reliability

morluto:
## Problem

Existing graph bond reliability does not model independently open vertices or independently open hyperedges. Related to #1688.

## Outcome

Adds two atomic exact operations:

- `probability.graph_site_reliability.connection_probability.compute`
- `probability.hypergraph_bond_reliability.connection_probability.compute`

Each returns an exact connection probability and a complete source-bound powerset state ledger. Site connectivity requires both terminals open; hypergraph connectivity uses chains of intersecting open hyperedges. The authoritative graph/hypergraph axes and state bitmasks survive JSON round-trip.

## Contract and bounds

At most 12 random vertices or hyperedges (4,096 states) and 24 graph edges or hypergraph vertices are admitted. Hyperedges are nonempty distinct vertex sets. Source-label codepoints, memberships, powerset work, state records, and exact-rational digit growth are admitted semantically before enumeration; transport adapters separately own encoded byte ceilings. Operation-specific limits are owned by typed admission rather than schema parsing.

## Evidence

Final local validation at `7efb317897983e7780eb2360d622a167a1642609`:

- `make affected AFFECTED_BASE=origin/main`
- 313 probability tests
- 160 catalog tests
- 968 catalog integration tests
- 32 focused reliability tests
- Ruff and mypy
- Independent exact-diff audit completed and repaired admission ownership and resource-bound findings.

This PR implements the bounded site/hypergraph reliability slice of #1688; broader reliability-polynomial and reduction operations remain open.

### DIFF
--- a/src/jacobian/math/probability/_hypergraph_bond_reliability.py
+++ b/src/jacobian/math/probability/_hypergraph_bond_reliability.py
@@ -16,6 +16,7 @@ from pydantic_core import PydanticCustomError
 
 from jacobian._exact import CanonicalRational, require_bounded_rational
 from jacobian._models import StrictModel
+from jacobian.canonical import encode_strict_json
 from jacobian.catalog.models import (
     MathTool,
     OperationDomainValidationError,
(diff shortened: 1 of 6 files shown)


morluto:
I have reproduced the current review findings and am addressing the complete set in this PR. I will reply to each thread with the validated change.

Issue #1192, opened 2026-08-11
Add verified bounded Ehrhart and weighted lattice-point enumeration


yuelgrace1810-ops:
## Area

Exact polytope capabilities, lattice-point enumeration, polynomial interpolation, independent verification.

## Observable gap

Current main exposes `polytope.separate`, which decides membership of one rational point in a finite convex hull and can preserve membership/separation evidence. It does not count lattice points in bounded dilates, compute an Ehrhart or weighted Ehrhart polynomial, interpolate exact coefficients from a degree bound, or independently verify such a counting certificate.

`math.find` searches for `Ehrhart`, `lattice point enumeration`, `integer point polytope`, and `polytope counting` return no matching operation. Searches across open and closed issues and open PRs found no owner for this capability.

This is not safely representable as repeated convex-hull membership: the proof obligation is a complete count over every lattice point in each declared dilation, followed by exact coefficient recovery. A manual/local enumeration cannot be promoted to Jacobian verification.

### Ordinary Ehrhart polynomial

Li–St. Dizier, arXiv:2608.00377v2 (2026), disprove Schubitope Ehrhart positivity with

- `E1={8,9,10,11,12,13,14}` and `E2={12,13,14}`;
- `S_E = Δ_{7,14}+Δ_{3,14} = Perm(2,2,2,1,1,1,1,0,0,0,0,0,0,0)`;
- an exact degree-13 lattice-count polynomial whose quadratic coefficient is `-687323/79200`.

A stdlib-only reconstruction independently reproduced all fourteen published values `L(0),...,L(13)` and the complete rational interpolation.

### Weighted Ehrhart tensor polynomial

Bajo et al., arXiv:2303.09614 (2024), disprove Berg–Jochemko–Silverstein Conjecture 6.1 with a displayed 20-dimensional lattice simplex and exact quadratic weight `w=ℓ²`, where `ℓ` takes values `+1` and `-1` on the displayed vertices. The proof-critical weighted coefficient is `-40/9` at `t^11`.

These families are independent: one is ordinary lattice-point counting for a generalized permutahedron/Minkowski sum; the other is polynomially weighted counting for a simplex. Both fail at the same missing reusable boundary: exact bounded enumeration plus coefficient recovery and verification.

## Required contract behavior

A solution should provide a bounded public producer and an independently implemented verifier that:

1. accept an exact, explicitly bounded lattice-polytope representation and dilation range;
[...]

PR #3656, opened 2026-09-12
feat(polytopes): interpolate bounded Ehrhart polynomials

morluto:
## Problem

Related to #1192.

Existing bounded lattice-point enumeration handled one polytope at one scale but did not return a source-bound table of dilations or interpolate an exact Ehrhart polynomial.

## Outcome

Adds `polytope.ehrhart.compute` for a full-dimensional integral polytope. It counts every lattice point in the admitted dilations, returns the complete exact count table, and interpolates a canonical `QQ[t]` polynomial from sufficient values. The source vertices, variable axis, degree bound, maximum dilation, and counts survive strict JSON serialization.

## Contract and bounds

Ambient dimension is at most four and positive dilation count at most 32. Before constructing scaled geometries, the operation admits full-dimensionality, scaled coordinate height, maximum per-axis span, every dilation box, aggregate scan work, facet work, polynomial/result height, and serialized output. It reuses the existing exact lattice-point scanner and rational-polynomial owner. Result decoding checks structural axes without replaying rank, counting, or interpolation.

## Evidence

Final local validation at `d52b782c38cad1e62741f5be3cbf7ff247a51b97`:

- `make affected AFFECTED_BASE=origin/main`
- 76 affected math tests
- 160 catalog tests
- 967 catalog integration tests
- Focused Ehrhart regressions, Ruff, formatting, and mypy
- Independent exact-diff audit completed; aggregate-scan, full-dimensionality ownership, and scaled-axis-span findings were repaired.

## Residual and superseded scope

This is the ordinary integral-polytope interpolation slice of #1192. Weighted Ehrhart counting remains open. The issue's older request for a paired count-replaying verifier is intentionally not implemented: under the current product contract, an exact producer result remains an ordinary value, and a downstream consumer checks an authored relation only when its own conclusion relies on it. No generic producer/verifier wrapper or replay during decoding is introduced.

### DIFF
--- a/src/jacobian/math/geometry/polytopes/lattice/_models.py
+++ b/src/jacobian/math/geometry/polytopes/lattice/_models.py
@@ -8,9 +8,11 @@ from pydantic import Field, model_validator
 from pydantic_core import PydanticCustomError
 
 from jacobian._exact import (
+    MAX_CANONICAL_INTEGER_DIGITS,
     ExactInteger,
     require_bounded_rational,
 )
+from jacobian.canonical import format_canonical_integer
 from jacobian._models import StrictModel
 from jacobian.math.geometry.polytopes.values import Halfspace as RationalHalfspace
 from jacobian.math.geometry.polytopes.values import Vertex as RationalVertex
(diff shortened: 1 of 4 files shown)


morluto:
I have reproduced the review findings and am addressing the complete set in this PR. I will reply to each thread with the validated change.

PR #3739, opened 2026-09-15
fix(polynomials): fail closed on root-critical isolation across all siblings

morluto:
## Summary

Follow-up to #3738. The real-root fix still left two same-class holes: non-real siblings were never pruned when publishing a real rectangle, and `exact_splitting_field` built its rectangles with an **empty** sibling family (`_isolating_rectangle(root, None, ())`). This makes isolation fail closed across the complete root family.

## Changes

**Isolation correctness**
- Thread the complete root family and the squarefree support through rectangle construction.
- Real rectangles are refused with the typed `polynomial.root_critical.sibling_separation` resource error unless `strict_root_count(support, lo, hi) == 1`.
- Non-real roots are refined against every sibling box (real, rational, and complex) with carrier fitting; a real rectangle that cannot be separated from a non-real sibling is refused.
- Roots whose `is_real` is undecidable are rejected with `polynomial.root_critical.root_reality_undecidable` instead of being silently dropped from the obstacle set (the original partial-obstacle-set failure mode).
- `_splitting.rectangle_for_root(root, polynomial, roots)` now refines splitting-field root and embedding rectangles against the whole support.

**Tests (PyTorch-OpInfo-style registry, SymPy-style properties)**
- `_isolation_invariants.py`: independent oracle asserting one support root per rectangle (`strict_root_count`/`count_roots`) and pairwise disjointness, plus axis completeness.
- `test_isolation_registry.py`: registry of every root-critical/splitting rectangle producer, audited uniformly.
- `test_hypothesis_isolation.py`: strategies that **construct** near-degenerate structure (`sqrt(a)` vs `sqrt(a + 1/q)`) and dense low-degree supports.
- `test_profile.py`: end-to-end near-collision regression through the public operation.

**Docs**
- Operation doc states the complete-family isolation invariant and its typed refusal.
- `testing-strategy.md` gets an "Isolation, refinement, and separation evidence" subsection (independent oracle, deliberately degenerate generators).
- `public-operation-admission.md` requires the invariant to quantify over the complete family, not an exactly-representable subset.

## Deviation from the earlier plan

The plan proposed reusing `Poly.intervals(all=True)` regions directly. Empirically those regions share boundaries and contain multiple roots under a closed-box reading (e.g. `z^4-2`, `z*(z^2-Az+1)`), so publishing them as closed rectangles is unsound. The implementation keeps the corrected per-root refinement and adds the fail-closed exact guard instead.

## Validation

- `uv run pytest tests/math/polynomials/root_critical -q` -> 71 passed
- `uv run pytest tests/catalog -q` -> 171 passed
- `tests/integration/catalog/test_builtin_examples.py` + `test_mcp_builtin_examples.py` -> 1011 passed (one unrelated macOS address-space worker failure in the multivariate factor example)
- `ruff check` / `ruff format --check` clean; `mypy` clean on the package

Closes the residual isolation defect from #3701 / #3738.


morluto:
Follow-up pushed: the residual complex-oracle limitation is now addressed. The invariant helper retries an axis-degenerate box by growing it outward by a quarter of a certified separation bound (`algebraic_root_separation_denominator_bound` on the primitive integer support); that expansion cannot include a distinct root, so the exact `count_roots` result remains the oracle instead of a disjointness fallback. Also added a documented 'Domain evidence patterns' table to `testing-strategy.md` (result family -> invariant / independent oracle / constructed degeneracy). `tests/math/polynomials/root_critical` remains 71 passed.

PR #3747, opened 2026-09-15
docs: clarify PARI unit identity and class/unit verifier status

morluto:
Surfaced by the class/unit-group audit: fundamental units are defined only up to sign/torsion associates with no published canonical normalization, and class/unit results have no standalone claim verifier. Docs-only, 1 file.


morluto:
Duplicate: this content already merged as #3746 (75561a694). Closing.

PR #3787, opened 2026-09-17
feat(polynomials): bind exact unit-circle sup-norm maxima (#2768)

### LINKED ISSUE #2768: [Polynomial analysis] Compute bounded exact unit-circle sup norms
## Need

Erdős #1150 asks whether every sufficiently long Littlewood polynomial has unit-circle maximum modulus bounded away from the Parseval scale by a fixed constant. A finite candidate or counterexample search needs a *global*, certified value of

```
||P||_∞,T = max_{|z|=1} |P(z)|,
```

not a sampled FFT grid. A grid can provide a lower bound only; it cannot certify the upper bound required to establish an ultraflat candidate.

Jacobian has exact rational-polynomial, root-isolation, real-algebraic comparison, and strict real-polynomial sublevel infrastructure, but no operation exposes this compact one-variable complex extremum. `RealAlgebraicValue` currently accepts minimal-polynomial degree at most 8, also too small to be a general result carrier for the algebraic extrema produced by degree-`n` input.

## Proposed public postcondition

Add an operation such as `polynomial.unit_circle.sup_norm.compute`:

> For one bounded polynomial `P(z)` with Gaussian-rational coefficients, return the exact real-algebraic value of `max_{|z|=1}|P(z)|^2`, a complete finite maximizing-point certificate, and a certified rational enclosure for its square root.

Returning the squared norm keeps the main value algebraic without introducing unnecessary square-root normalization. A Littlewood polynomial is simply a client of this general operation.

## Exact reduction

Use the rational parametrisation of the unit circle away from `z=-1`,

```
z = (1 + i t) / (1 - i t),  t in R,
```

and clear the positive denominator in `|P(z)|^2` to obtain a rational function `Q(t)/(1+t^2)^d` with rational coefficients. Every finite maximizer is a real root of the exact derivative numerator; the omitted point `z=-1` is a separately retained endpoint-at-infinity candidate. Exact root isolation and exact comparison of all critical and endpoint values then establish the maximum.

The result must retain enough source-bound data to replay:

- the transformed numerator and denominator exponent;
- completeness of the real critical-point family;
- the value comparison ledger;
- the `z=-1` candidate; and
- the correspondence from every returned maximizing parameter to a point on the unit circle.

## Existing-tool improvement required

This is intentionally an extension of the real-algebra vocabulary, not a floating optimiser. It should reuse the exact polynomial/root infrastructure delivered around #1872 and `polynomial.real.strict_sublevel_measure.compute`, but its result contract needs an algebraic-extremum carrier beyond the current degree-8 `RealAlgebraicValue` envelope. The admission review should decide whether to widen that shared value safely or introduce a source-bound extremum value whose defining polynomial degree and isolation work are preflighted together.

## Bounds and semantics

V1 should use explicit limits on input degree, Gaussian-rational coefficient height, transformed/derivative degree and height, number of real critical points, algebraic-value degree, comparison work, and retained certificate size. Every admitted call returns an exact complete maximum; an out-of-envelope request is rejected before root isolation. No sampled maximum, floating tolerance, or failed solver is an upper-bound certificate.

## Why this is a stable operation

The same norm is central to Littlewood, Newman, and flat-polynomial experiments, spectral/analytic polynomial bounds, and certificate checking for proposed extremisers. It is not an Erdős-specific search, and it does not choose coefficient signs or make an asymptotic claim.

## Sources

- [Erdős Problem #1150](https://www.erdosproblems.com/1150)
- [Formal conjecture statement and Parseval baseline](https://github.com/google-deepmind/formal-conjectures/blob/main/FormalConjectures/ErdosProblems/1150.lean)
- #1872 and #2308

## Non-goals

- searching the `2^n` Littlewood coefficient space;
- proving the asymptotic #1150 conjecture;
- numerical-only FFT grids;
- arbitrary complex-analytic optimization or general CAD;
- a generic workflow combining candidate generation, pruning, and literature conclusions.

morluto:
## Problem
#2768 asks for the exact real-algebraic value of `max_{|z|=1}|P(z)|^2`, not just a certified enclosure. `polynomial.unit_circle.sup_norm_squared.compute` returned the enclosure plus the complete critical-point ledger but no exact value: the maximum's minimal polynomial can exceed the shared degree-16 `RealAlgebraicValue` carrier, so the owner deliberately did not normalize it.

## Outcome
The result gains `sup_norm_squared_exact: RealAlgebraicValue | None`:
- rational maxima (FULL_CIRCLE constant, ENDPOINT, CRITICAL_AND_ENDPOINT) bind the degree-one value directly;
- isolated critical maxima bind the indexed root of the unique irreducible resultant factor strictly sign-changing inside the maximal isolating interval, with the Sturm root index below the interval;
- the field is `None` (fail-closed) when the carrying factor leaves the shared carrier (degree > 16 or coefficient digits > 1000); the enclosure remains the complete certificate in that case.
No shared carrier was widened; the new field defaults to `None`, so previously serialized results still parse.

## Validation
- `tests/math/polynomials/test_unit_circle_sup_norm.py`: 29 passed, including new rational fixtures (`z^2+z+1 -> 9`, `(1+i)z -> 2`, `z-1 -> 4`, `z^3-2z^2-1 -> 16`), irrational fixtures with sympy-free bracket replays (Littlewood `1+z-z^2+z^3` -> quadratic `(27,-216,176)` index 1; quartic -> cubic `(4608,-318684,2623484,8833)` index 2), forgery rejection, None-weakening acceptance, and serialization round-trip.
- `make lint-scoped`, `make typecheck-scoped`: clean (also fixed a pre-existing missing return annotation on the test helper).
- `make test-catalog`: 171 passed. `tests/math/polynomials` (minus multivariate factor) and `tests/integration/catalog`: all green except the known pre-existing sandbox failures in `polynomial.multivariate.factor` paths (`no hard address-space limit could be applied`), which fail identically on clean `origin/main`.

## Contract impact
Additive optional result field on `polynomial.unit_circle.sup_norm_squared.compute`; tool description updated. The verifier replays a present exact value by rerunning the admitted kernel and requiring equality.

## Review closure
Closes #2768.


morluto:
## Limitation analysis (follow-up commit 7c9d91801)

The original PR flagged a `None` exact-value fallback. I ran that down with bounded probes and it splits into two independent claims:

**1. Degree — eliminated (theorem, not a measurement).**
The critical resultant has `y`-degree equal to the coprime derivative-numerator degree, which is **at most `2d`**: writing `Q = A**2 + B**2` with `deg W = m <= d`, the `t**(2m+1)` leading coefficients of `Q'(1+t**2)` (`2m|w_m|**2`) and `2*d*t*Q` (`2d|w_m|**2`) cancel when `m = d`, and `deg D = 2m+1 <= 2d-1` otherwise. So every minimal factor has degree `<= 2d <= 16` and the shared degree-16 carrier always suffices. Docstrings and two degree-8 regression tests now pin this (`test_top_admitted_degree_carries_an_exact_value`, `test_exact_value_respects_the_resultant_degree_carrier`).

**2. Height — bounded, fail-closed, effectively unreachable.**
Measured derived digit growth is *linear*, not quadratic: `numer/deriv digits ~ 2.2*h + 2` (h=8→20/21, h=16→37/38, h=24→53/54), so the existing (4h+8n+32)/512 bound is sound and generous. A worst-case height-overflow factor (>1000 digits) is therefore the only `None` path; it is defensive and does not occur in the tested envelope.

**Remaining limitation (reported, not repaired): admitted-but-slow corner.**
A random degree-8 input with 32-digit coefficients is *admitted* by `MAX_SUP_NORM_ISOLATION_WORK` yet the resultant + `factor_list` step exceeded **150 s** and never produced output, so it dies on the request deadline rather than returning. This is operationally safe (timeout is non-completion, not a wrong answer) but the isolation-work proxy does not model the resultant/factorization cost. Calibrating that admission term is a separate scale change and was deliberately not attempted here (avoiding an uncalibrated over-rejection). Flagging for a dedicated scale pass.

Validation: 33 focused tests pass (3.1 s), `lint-scoped`/`typecheck-scoped` clean, `test-catalog` 171 passed.

morluto:
Split the remaining limitation out to #3788 to keep this PR scoped to the sound degree/height contract.

